PR #4, opened 2020-03-12
[final] invalidate purchaser info cache

aboedo:
added calls to invalidate purchaser info cache on android and iOS

PR #5, opened 2020-03-17
[final] Feature/subscriber attributes

aboedo:
Adds calls to methods relating to subscriber attributes
### Requirements: 
- [x] ~#4~


android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
```diff
@@ -8,6 +8,7 @@ import com.revenuecat.purchases.Offerings
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.PurchaserInfo
 import com.revenuecat.purchases.util.Iso8601Utils
+import org.json.JSONObject
```
aboedo: forgot about it, thanks for catching!

PR #12, opened 2020-03-30
Checks for NSNull when setting attributes

vegaro:
When passing `nil` in Flutter when calling one of the special attributes setter, the value is converted to `NSNull`. I added a check so that this is fixed for all platforms potentially doing the same.

### DIFF
--- a/ios/RCCommonFunctionality.m
+++ b/ios/RCCommonFunctionality.m
@@ -314,22 +314,30 @@ signedDiscountTimestamp:(nullable NSString *)discountTimestamp
 
 + (void)setEmail:(nullable NSString *)email {
     NSAssert(RCPurchases.sharedPurchases, @"You must call setup first.");
-    [RCPurchases.sharedPurchases setEmail:email];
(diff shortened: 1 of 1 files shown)


ios/RCCommonFunctionality.m
```diff
@@ -314,22 +314,30 @@ + (void)setAttributes:(NSDictionary<NSString *, NSString *> *)attributes {
...
 
 + (void)setPushToken:(nullable NSString *)pushToken {
     NSAssert(RCPurchases.sharedPurchases, @"You must call setup first.");
-    [RCPurchases.sharedPurchases _setPushTokenString:pushToken];
+    NSString *nonNSNullAttribute = [self nonNSNullAttribute:pushToken];
+    [RCPurchases.sharedPurchases _setPushTokenString:nonNSNullAttribute];
+}
+
++ (NSString * _Nullable)nonNSNullAttribute:(NSString * _Nullable)attribute {
+    return ([attribute isEqual:NSNull.null]) ? @"" : attribute;
```
aboedo, line 340: FFTI: this'd make for a great `NSString *` extension

PR #15, opened 2020-04-08
[FINAL] Unity compatibility

vegaro:
Dates in Unity were using millis instead of Strings. I basically copied over all the dates from the Unity files and changed the name to end with "Millis".

The cons of these changes will also be present in the Javascript objects. RN and cordova converts anything we send to the Javascript layer automatically to Javascript objects. I don't think it's a big deal since we have well defined Typescript types.

### DIFF
--- a/android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
+++ b/android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
@@ -8,8 +8,11 @@ import com.revenuecat.purchases.Offerings
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.PurchaserInfo
 import com.revenuecat.purchases.util.Iso8601Utils
+import org.json.JSONArray
+import org.json.JSONObject
 import java.text.NumberFormat
 import java.util.Currency
+import java.util.Date
 
 fun EntitlementInfo.map(): Map<String, Any?> =
     mapOf(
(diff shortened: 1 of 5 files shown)


android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
```diff
@@ -17,19 +17,28 @@ fun EntitlementInfo.map(): Map<String, Any?> =
         "isActive" to this.isActive,
         "willRenew" to this.willRenew,
         "periodType" to this.periodType.name,
+        "latestPurchaseDateMillis" to this.latestPurchaseDate.time / 1000.0,
```
aboedo, line 20: maybe a Date extension would make cleaner by killing duplication as in `toMillis()`

vegaro: yeah let's do it

android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
```diff
@@ -222,3 +231,35 @@ private fun String?.mapPeriod(): Map<String, Any?> {
         }
     }
 }
+
+fun Map<String, *>.convertToJson(): JSONObject {
```
aboedo, line 235: 😍

ios/NSDate+HybridAdditions.m
```diff
@@ -0,0 +1,35 @@
+//
+// Created by Andrés Boedo on 3/4/20.
```
aboedo, line 2: LOL I don't think I created this one

PR #17, opened 2020-04-15
Use period serialization extension

vegaro:
Added a `convertToMap` function that can be reused

### DIFF
--- a/android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
+++ b/android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
@@ -254,6 +254,15 @@ fun List<*>.convertToJsonArray(): JSONArray {
     return writableArray
 }
 
+fun JSONObject.convertToMap(): Map<String, String?> =
+    this.keys().asSequence<String>().associate { key ->
+        if (this.isNull(key)) {
+            key to null
+        } else {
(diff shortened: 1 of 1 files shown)


aboedo:
@vegaro what's missing from this one?

vegaro:
Nothing, I totally forgot about it

android/common/src/main/java/com/revenuecat/purchases/common/mappers.kt
```diff
@@ -258,6 +254,15 @@ fun List<*>.convertToJsonArray(): JSONArray {
     return writableArray
 }
 
+fun JSONObject.convertToMap(): Map<String, String?> =
+    this.keys().asSequence<String>().associate { key ->
+        if (this.isNull(key)) {
```
aboedo, line 259: will this map null to null? would we want to skip instead?

vegaro: It maps `JSON.Null` to `null`, and we want the nulls in the map

vegaro:
We can batch this one with the other PR for release

aboedo:
might wanna do it separately, since the other one changes source files locations and requires a bit more work when integrating

vegaro:
ok, I will do it now then

PR #18, opened 2020-04-23
add xcproject and tests target

aboedo:
Does a few changes: 

- Adds a new xcode project, which gives us the ability to compile the files here to make sure that they work correctly. 
- The xcode project compiles a framework, which should be the preferred way to use the hybrid common stuff in the future, but updating isn't strictly necessary yet. 
- [BREAKING CHANGE] Moves the source files from ios/<file> to ios/PurchasesHybridCommon/PurchasesHybridCommon/<file>. References will have to be updated. 
- Adds test target, configured with Quick tests. 
- Adds a sample test
- Sets up `purchases-ios` as a git submodule of purchases-hybrid-common. 
This doesn't make a difference for SDKs that just pull the files from common, they should still work as usual. However, if you use the `PurchasesHybridCommon.framework`, you need to make sure that the `Purchases.framework` can be located by hybrid common in framework search paths. We'll automate this a bit more in the future. 

For development purposes, you'll need to `git submodule init` to make sure that `purchases-ios` gets checked out so that you can compile the xcode project.


ios/PurchasesHybridCommon/PurchasesHybridCommon.h
```diff
@@ -0,0 +1,31 @@
...
+//  Copyright © 2020 RevenueCat. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for PurchasesHybridCommonFramework.
+FOUNDATION_EXPORT double PurchasesHybridCommonFrameworkVersionNumber;
+
+//! Project version string for PurchasesHybridCommonFramework.
+FOUNDATION_EXPORT const unsigned char PurchasesHybridCommonFrameworkVersionString[];
```
vegaro, line 15: oh this is a good idea

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/SKProductDiscountHybridAdditionsTests.swift
```diff
@@ -0,0 +1,21 @@
...
+import Quick
+import Nimble
+
+class SkuProductDiscountHybridAdditionsTests: QuickSpec {
+    override func spec() {
+        describe("the 'Documentation' directory") {
+            it("has everything you need to get started") {
+                expect(true) == true
+            }
+        }
```
vegaro, line 19: fancy

PR #22, opened 2020-04-24
Android CI tests

aboedo:
added android CI configuration for purchases-hybrid-common


vegaro:
we can improve the builds with gradle cache, but for now it's good

aboedo:
agreed, but I figured we can always improve it later on, it's nice to have something working

PR #26, opened 2020-05-15
Feature: podspec

aboedo:
Adds a podspec, does minor cleanup in framework headers


PurchasesHybridCommon.podspec
```diff
@@ -0,0 +1,36 @@
...
+    'ios/PurchasesHybridCommon/PurchasesHybridCommon/**/*.h',
+    'ios/PurchasesHybridCommon/PurchasesHybridCommon/**/*.m'
+  ]
+
+
+  s.public_header_files = [
+    'ios/PurchasesHybridCommon/PurchasesHybridCommon/Public/*.h'
+  ]
+
+end
```
vegaro, line 36: Next step would be to remove the submodule and add the dependency here right?

aboedo: yep, will do it as a follow-up PR

PR #31, opened 2020-05-19
Creates Maven library

vegaro:
Adds a job to CircleCI to automatically release a library to Maven when pushing a task


.circleci/config.yml
```diff
@@ -25,8 +25,6 @@ jobs:
     executor: ios-executor
     steps:
       - checkout
-      - run: git submodule sync
-      - run: git submodule update --init
```
aboedo, line 29: thanks for killing these, I'd forgotten about them

.circleci/config.yml
```diff
@@ -64,11 +70,48 @@ jobs:
...
+              echo "Skipping snapshot deployment: wrong repository. Expected '$REPO' but was '$CIRCLE_PROJECT_REPONAME'."
+            elif [ "$CIRCLE_PROJECT_USERNAME" != "$USERNAME" ]; then
+              echo "Skipping snapshot deployment: wrong owner. Expected '$USERNAME' but was '$CIRCLE_PROJECT_USERNAME'."
+            elif [ "$CIRCLE_JDK_VERSION" != "$JDK" ]; then
+              # $CIRCLE_JDK_VERSION must be manually set in circle.yml
+              echo "Skipping snapshot deployment: wrong JDK. Expected '$JDK' but was '$CIRCLE_JDK_VERSION'."
+            else
+              echo "Deploying snapshot..."
+              ./gradlew androidSourcesJar androidJavadocsJar uploadArchives -P signing.keyId=$GPG_SIGNING_KEY_ID -Psigning.password=$GPG_SIGNING_KEY_PW -Psigning.secretKeyRingFile=./secring.gpg \
+                                      -PSONATYPE_NEXUS_USERNAME=$SONATYPE_NEXUS_USERNAME -PSONATYPE_NEXUS_PASSWORD=$SONATYPE_NEXUS_PASSWORD -PRELEASE_SIGNING_ENABLED=true
```
aboedo, line 98: maybe we could break lines in between params?

aboedo: as in 
```sh
./gradlew androidSourcesJar androidJavadocsJar uploadArchives \
    -P signing.keyId=$GPG_SIGNING_KEY_Iassword=$GPG_SIGNING_KEY_PW \
    -Psigning.secretKeyRingFile=./secring.gpg \
    -PSONATYPE_NEXUS_USERNAME=$SONATYPE_NEXUS_USERNAME \
    -PSONATYPE_NEXUS_PASSWORD=$SONATYPE_NEXUS_PASSWORD \
    -PRELEASE_SIGNING_ENABLED=true
```

aboedo: also, is there a typo in the second param? is it supposed to be `-P signing.secretKeyRingFile` instead of `-Psigning.secretKeyRingFile`?

vegaro: That's strange! it's the same in the purchases-android script https://github.com/RevenueCat/purchases-android/blob/develop/.buildscripts/deploy_snapshot.sh

aboedo: just realized that there's another param that has the same weirdness, `signing.password`

vegaro: I doubled checked in other examples and looks like it's `-Psigning` instead of `-P signing` in both cases

vegaro:
I changed the filters to run a workflow only when pushing tags

PR #33, opened 2020-05-21
method to pass in a suite name to revenuecat

aboedo:
this will provide a workaround for issues with NSUserDefaults in hybrid SDKs. 
This would have been difficult to do from hybrid SDKs otherwise because they would have had to create an instance of NSUserDefaults to pass in through javascript or C#.


ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridAdditionsTests.swift
```diff
@@ -26,5 +26,27 @@ class PurchasesHybridAdditionsTests: QuickSpec {
...
+        context("configure with user defaults suite name") {
+            it("initializes without raising exceptions if no suite name is passed") {
+                expect {
+                    Purchases.configure(withAPIKey: "api key",
+                                        appUserID: nil,
+                                        observerMode: false,
+                                        userDefaultsSuiteName: nil,
+                                        platformFlavor: "hybrid-platform",
+                                        platformFlavorVersion: "1.2.3")
+                }.notTo(raiseException())
```
aboedo, line 38: couldn't come up with a better way to test than "make sure it doesn't raise exceptions", since this is a static method. let me know if you have any ideas

ios/PurchasesHybridCommon/PurchasesHybridCommon/RCPurchases+HybridAdditions.m
```diff
@@ -0,0 +1,38 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+
+@implementation RCPurchases (HybridAdditions)
+
++ (instancetype)configureWithAPIKey:(NSString *)APIKey
+                          appUserID:(nullable NSString *)appUserID
+                       observerMode:(BOOL)observerMode
+              userDefaultsSuiteName:(nullable NSString *)userDefaultsSuiteName
```
vegaro, line 18: How are we going to surface this in the hybrid SDKs? What if we always instantiate using a special name instead? Like what if we do:
`[[NSUserDefaults alloc] initWithSuiteName:@"revenuecat"];`

Would that fix the issues?

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridAdditionsTests.swift
vegaro: What I do in Android is mock the static object and verify the methods are called with certain parameters. But I don't think you can do that in Objective-C

I don't think this `notTo(raiseException())` is bad

ios/PurchasesHybridCommon/PurchasesHybridCommon/RCPurchases+HybridAdditions.m
aboedo: that's a good idea. my one concern though is potential issues with migration - if there are unsynced subscriber attributes and such in the old user defaults, they'd get lost

vegaro: ohhh good thing you thought about that. Let's leave it this way then 👍

Issue #34, opened 2020-05-21
discountWithIdentifier always returns null


danmelton:
in the file RCCommonFunctionality.m:

```
+ (nullable SKProductDiscount *)discountWithIdentifier:(NSString *)identifier
                                            forProduct:(SKProduct *)aProduct API_AVAILABLE(ios(12.2), macos(10.14.4), tvos(12.2)) {
    SKProductDiscount *discountToUse = nil;
    NSArray<SKProductDiscount *> *productDiscounts = aProduct.discounts;
    if (identifier == nil && productDiscounts != nil && productDiscounts.count > 0) {
        discountToUse = productDiscounts.firstObject;
    } else {
        for (SKProductDiscount *discount in productDiscounts) {
            if (identifier == discount.identifier) {
                discountToUse = discount;
            }
        }
    }
    return discountToUse;
}
```
identifier == discount.identifier never returns true, resulting in a null discountToUse.

I believe, the if condition needs to check string equality.

```
            if (identifier == discount.identifier) {
                discountToUse = discount;
            }
```

should be

```
            if ([identifier isEqualToString:discount.identifier]) {
                discountToUse = discount;
            }
```

We're using patch-package to get that method to work. Submitting a PR.

aboedo:
great catch! thanks for reporting and posting the PR!

aboedo:
closing since the fix has been merged and shipped. Thanks again @danmelton!!

PR #35, opened 2020-05-21
updating discountWithIdentifier method to use string equality check GH-34

danmelton:
Adds string equality check for discount.identifier.

We're getting 'Couldn't find discount.' even though the discount is present and available in the discounts array on the product object,


aboedo:
thanks again for catching and posting a PR with the fix!

PR #36, opened 2020-05-21
expose originalPurchaseDate in RCPurchaserInfo

aboedo:
added originalPurchaseDate to RCPurchaserInfo dict

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/RCPurchaserInfo+HybridAdditions.m
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/RCPurchaserInfo+HybridAdditions.m
@@ -7,48 +7,64 @@
 #import "RCEntitlementInfos+HybridAdditions.h"
 #import "NSDate+HybridAdditions.h"
 
+
 @implementation RCPurchaserInfo (HybridAdditions)
 
-- (NSDictionary *)dictionary
-{
+- (NSDictionary *)dictionary {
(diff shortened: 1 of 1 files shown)


aboedo:
Adding tests for this is pretty tricky because we have to create a whole bunch of mocks. I'd like to improve our testing game, but I also don't want to delay this any further, since we've had it in the backlog for a while.

ios/PurchasesHybridCommon/PurchasesHybridCommon/RCPurchaserInfo+HybridAdditions.m
```diff
@@ -7,48 +7,64 @@
...
+    }
+}
+
+- (NSObject *)formattedAsISO8601OrNull:(nullable NSDate *)date {
+    if (date) {
+        return date.formattedAsISO8601;
+    } else {
+        return NSNull.null;
+    }
+}
```
vegaro, line 68: Should we move these two to `NSDate+HybridAdditions.h`?

aboedo: that wouldn't work, unfortunately, because if the date is nil the func will no-op. it could be a static method, but I'm not sure it would add much value

vegaro: oh got it

PR #39, opened 2020-05-27
Switch to use carthage instead of cocoapods

vegaro:
This doesn't mean that the single target in the project can build all the platforms. For example, trying to build for tvOS would show this error:

Screen Shot 2020-05-27 at 4 52 54 PM

Meaning that the linked Purchases.framework was built for iOS. I think the solution for this would be to create different targets, one per platform and link the Purchases.framework from the corresponding Carthage/Build folder.

I didn't add the `/usr/local/bin/carthage copy-frameworks` build phase since as far as I understand this is only required for Apps to circumvent an App Store submission bug where apps with frameworks that contain binary images for the simulator are automatically rejected. 

To build this framework we have to do `carthage build --no-skip-current --platform iOS`, that would give us a framework we can use with React Native. 

So unless I am missing something, we don't really get any benefit over using Cocoapods.


aboedo:
we should replace the pod install step in CI with carthage bootstrap for tests to run correctly

vegaro:
sure @aboedo , I didn't do it because as I said in the description I am not sure we should move forward with this since it doesn't really give us any benefit. What do you think?

aboedo:
this is tricky 🤔 
The target in the xcproject is really just for local development, since pods will only use the podspec to find classes and build. 
so one alternative is to have multiple targets (i.e.: 1 per platform) instead of a single, multiplatform target. 
This would allow each platform to link to the correct `Purchases.framework` through Cocoapods or carthage. 
We could have the productName be the same for all of them so that when carthage builds it outputs `PurchasesHybridCommon.framework` for all targets.

vegaro:
I did that yesterday, created a target per platform and added the Purchases.framework from each of the carthage build folders to each of the targets. But then I thought that if the sole purpose of doing this PurchasesHybridCommon.framework is for React Native to work, should we even doing this or just stay with what we have, which already works. We are only using the iOS Purchases.framework in React Native anyway

aboedo:
right, forgot about that. I mean, it's kind of a shame because the common part truly is multi-platform, but it's not really important to be able to generate one .framework per platform if no libraries actually use it. 

In that case, we should be able to just generate the ios framework and call it a day for now, right?

aboedo:
as for whether to use carthage or cocoapods, the only advantage of carthage is to generate the .framework file easily, but we can use xcbuild for that as well. happy to go with whatever is easiest.

vegaro:
Carthage can still generate the framework even if we use Cocoapods. I am closing this. Thanks for reviewing @aboedo :)

PR #46, opened 2020-07-07
Updates to Android SDK 3.3.0

vegaro:
[HOLD] Missing release of https://github.com/RevenueCat/purchases-android/commit/eeb7a478a127092285040b795eabf1e51e64899a

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/common/mappers.kt
+++ b/android/src/main/java/com/revenuecat/purchases/common/mappers.kt
@@ -143,8 +143,8 @@ private fun SkuDetails.mapIntroPrice(): Map<String, Any?> {
         // Check freeTrialPeriod first to give priority to trials
         // Format using device locale. iOS will format using App Store locale, but there's no way
         // to figure out how the price in the SKUDetails is being formatted.
-        val format = java.text.NumberFormat.getCurrencyInstance().apply {
-            currency = java.util.Currency.getInstance(priceCurrencyCode)
+        val format = NumberFormat.getCurrencyInstance().apply {
+            currency = Currency.getInstance(priceCurrencyCode)
         }
         mapOf(
             "price" to 0,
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/revenuecat/purchases/common/mappers.kt
```diff
@@ -70,7 +70,9 @@ fun PurchaserInfo.map(): Map<String, Any?> =
         "allPurchaseDates" to allPurchaseDatesByProduct.mapValues { it.value?.toIso8601() },
         "allPurchaseDatesMillis" to allPurchaseDatesByProduct.mapValues { it.value?.toMillis() },
         "originalApplicationVersion" to null,
-        "managementURL" to managementURL?.toString()
+        "managementURL" to managementURL?.toString(),
+        "originalPurchaseDate" to originalPurchaseDate?.toIso8601(),
+        "originalPurchaseDateMillis" to originalPurchaseDate?.toMillis()
```
aboedo, line 75: beat me to it. I've had a local stash for this for weeks and never got around to finishing it. 
I have tests stashed as well, posting here in case they're useful: 

PurchaserInfoMappersTests.kt
```kotlin
        context("with an originalPurchaseDate") {
            it("should map to an originalPurchaseDate") {
                val date: Date = Date(2020, 3, 23)
                every { mockPurchaserInfo.originalPurchaseDate } returns date

                val map = mockPurchaserInfo.map()
                assertThat(map["originalPurchaseDate"]).isEqualTo(date.toIso8601())
                assertThat(map["originalPurchaseDateMillis"]).isEqualTo(.isEqualTo(date.toMillis())
            }
        }
```

And I also noticed a bug on that same file in line 25
```kotlin
            it("should map to a null managementURL") {
```
should be

```kotlin
            it("should map to a non-null managementURL") {
```

vegaro: Sure, I will add that!

I just added it because I had it in my own list of things to do and I wanted to do something not too big before the weekend :)

PR #48, opened 2020-07-09
Prepares release 1.3.0

vegaro:
- Bumped Android to 3.3.0 ([Changelog here](https://github.com/RevenueCat/purchases-android/releases))
    https://github.com/RevenueCat/purchases-hybrid-common/pull/46
- Added parsing of original_purchase_date in Android
    https://github.com/RevenueCat/purchases-hybrid-common/pull/46
- Bumped iOS to 3.5.0 ([Changelog here](https://github.com/RevenueCat/purchases-ios/releases))

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,10 +1,10 @@
 ### Releasing a version: 
 
-- Make a release branch `release/x.x.x`
-- Create a CHANGELOG.latest.md with the changes for the current
+- Start a git-flow release/x.y.z
+- Create a CHANGELOG.latest.md with the changes for the current version (to be used by Fastlane for the github release notes)
 - Run `fastlane bump_and_update_changelog version:X.Y.Z` (where X.Y.Z is the new version) to update the version number in `android/build.gradle`, `android/gradle.properties` and `PurchasesHybridCommon.podspec`
 - Update purchases-android version in `android/build.gradle`
-- Update purchases-ios pod version in `PurchasesHybridCommon.podspec`, `ios/PurchasesHybridCommon/Podfile`, and `ios/PurchasesHybridCommon/PurchasdsHybridCommon.xcworkspace` (open this last one with Xcode)
+- Update purchases-ios pod version in `PurchasesHybridCommon.podspec` and `ios/PurchasesHybridCommon/Podfile`.
 - Open a PR, merge and tag master.
 - run `carthage build --archive --platform iOS`
 - Upload to the new release PurchasesHybridCommon.framework.zip
--- a/ios/PurchasesHybridCommon/Info.plist
+++ b/ios/PurchasesHybridCommon/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(MARKETING_VERSION)</string>
+	<string>1.3.0</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 </dict>
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Info.plist
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>1.3.0</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 </dict>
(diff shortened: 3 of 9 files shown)


RELEASING.md
```diff
@@ -1,10 +1,10 @@
 ### Releasing a version: 
 
-- Make a release branch `release/x.x.x`
-- Create a CHANGELOG.latest.md with the changes for the current
+- Start a git-flow release/x.y.z
+- Create a CHANGELOG.latest.md with the changes for the current version (to be used by Fastlane for the github release notes)
 - Run `fastlane bump_and_update_changelog version:X.Y.Z` (where X.Y.Z is the new version) to update the version number in `android/build.gradle`, `android/gradle.properties` and `PurchasesHybridCommon.podspec`
 - Update purchases-android version in `android/build.gradle`
-- Update purchases-ios pod version in `PurchasesHybridCommon.podspec`, `ios/PurchasesHybridCommon/Podfile`, and `ios/PurchasesHybridCommon/PurchasdsHybridCommon.xcworkspace` (open this last one with Xcode)
```
vegaro, line 7: @aboedo I didn't understand what this means by "`ios/PurchasesHybridCommon/PurchasdsHybridCommon.xcworkspace` (open this last one with Xcode)". Is it required? I didn't see anything changed in the xcworkspace the other PRs updating the iOS versions

ios/PurchasesHybridCommon/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(MARKETING_VERSION)</string>
+	<string>1.3.0</string>
```
vegaro, line 18: fastlane made this change, is this fine?

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>1.3.0</string>
```
vegaro, line 18: I assume this won't break anything since it's tests

RELEASING.md
aboedo: I'm not even sure 😂  I _believe_ I meant "you should probably open it with Xcode since it's not meant to be human-readable". 
I'm glad the step is gone, though. Long live fastlane.

ios/PurchasesHybridCommon/Info.plist
aboedo: 🤔 did it also update the version in the project file?

aboedo: are you using the latest version of fastlane? was this with xcode 11 or 12? maybe there's some incompatibility with the formats and the project doesn't update?

vegaro: What file is supposed to update, the xcproject? This was both in Xcode 11 and 12

vegaro: I looked at this https://github.com/RevenueCat/purchases-hybrid-common/pull/42/files and no xcproject was updated

vegaro: I see. So there's a Marketing Version in the Build Settings

vegaro: Is it fine if we let Fastlane increment the `CFBundleShortVersionString` like in this commit, and then we add something like the following to increment the `MARKETING_VERSION` in the Build Settings:

```
version=$(sed -n '/MARKETING_VERSION/{s/MARKETING_VERSION = //;s/;//;s/^[[:space:]]*//;p;q;}' {your project name}.xcodeproj/project.pbxproj)a
```

aboedo: 🤔 I'm surprised that fastlane bump doesn't do this automatically, it does for purchases-ios. 
also, I just noticed that the fastfile does bump by calling "sh fastlane ios ..." through the console, but it doesn't need to, it can just call the command directly as in https://github.com/RevenueCat/purchases-ios/blob/develop/fastlane/Fastfile#L34 or even call the method `bump` directly

vegaro: Please note that this Fastfile is in the root folder so it has both android and ios platform lanes. That's the only way I could find to specify the platform I wanted to do the bump on

vegaro: Also, the fastlane in iOS doesn't increment the `MARKETING_VERSION`, it also increments `CFBundleShortVersionString` like in this repo https://github.com/RevenueCat/purchases-ios/pull/274/files#diff-d998ce5a98d33edbdd2da993fd55f1db

PR #49, opened 2020-07-14
Fixes Period parsing crash

vegaro:
Should fix https://github.com/RevenueCat/purchases-flutter/issues/40

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -96,10 +96,8 @@ dependencies {
 
     testImplementation "io.mockk:mockk:$mockk_version"
     // Core library
-    testImplementation 'junit:junit:4.12'
     testImplementation 'androidx.test:core:1.2.0'
     // AndroidJUnitRunner and JUnit Rules
-//    testImplementation 'org.robolectric:robolectric:4.3.1'
     testImplementation 'androidx.test.ext:junit:1.1.1'
     testImplementation 'org.robolectric:robolectric:4.3.1'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -39,7 +39,6 @@ android {
         versionCode 1
         versionName "1.2.0"
 
-        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         consumerProguardFiles 'consumer-rules.pro'
     }
 
@@ -72,6 +71,9 @@ android {
         unitTests.all {
             testLogging.events = ["passed", "skipped", "failed"]
         }
+        unitTests {
+            includeAndroidResources = true
+        }
     }
 }
 
@@ -93,6 +95,13 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertj_version"
 
     testImplementation "io.mockk:mockk:$mockk_version"
+    // Core library
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'androidx.test:core:1.2.0'
+    // AndroidJUnitRunner and JUnit Rules
+//    testImplementation 'org.robolectric:robolectric:4.3.1'
+    testImplementation 'androidx.test.ext:junit:1.1.1'
+    testImplementation 'org.robolectric:robolectric:4.3.1'
 }
 
 repositories {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -71,9 +71,6 @@ android {
         unitTests.all {
             testLogging.events = ["passed", "skipped", "failed"]
         }
-        unitTests {
-            includeAndroidResources = true
-        }
     }
 }
 
@@ -95,11 +92,6 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertj_version"
 
     testImplementation "io.mockk:mockk:$mockk_version"
-    // Core library
-    testImplementation 'androidx.test:core:1.2.0'
-    // AndroidJUnitRunner and JUnit Rules
-    testImplementation 'androidx.test.ext:junit:1.1.1'
-    testImplementation 'org.robolectric:robolectric:4.3.1'
 }
 
 repositories {
--- a/android/src/main/java/com/revenuecat/purchases/common/PurchasesPeriod.java
+++ b/android/src/main/java/com/revenuecat/purchases/common/PurchasesPeriod.java
@@ -1,6 +1,9 @@
 package com.revenuecat.purchases.common;
 
+import android.util.Log;
+
 import androidx.annotation.NonNull;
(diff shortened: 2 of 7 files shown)


android/build.gradle
```diff
@@ -93,6 +95,13 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertj_version"
 
     testImplementation "io.mockk:mockk:$mockk_version"
+    // Core library
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'androidx.test:core:1.2.0'
+    // AndroidJUnitRunner and JUnit Rules
+//    testImplementation 'org.robolectric:robolectric:4.3.1'
+    testImplementation 'androidx.test.ext:junit:1.1.1'
+    testImplementation 'org.robolectric:robolectric:4.3.1'
```
vegaro, line 102: had to make this changes because I added a call to `Log.e`, which is part of the Android dependencies and the `android.jar` that is used with Unit tests don't have implementations so we need to use roboelectric, which has default implementations for all these calls.

android/src/main/java/com/revenuecat/purchases/common/PurchasesPeriod.java
```diff
@@ -84,24 +87,22 @@ static PurchasesPeriod parse(@NonNull CharSequence text) {
...
-                    throw new RuntimeException("Text cannot be parsed to a Period: " + text, ex);
+                } catch (NumberFormatException | ArithmeticException ex) {
+                    Log.e("PURCHASES", "Text cannot be parsed to a Period: " + text);
+                    return null;
                 }
             }
         }
-        throw new RuntimeException("Text cannot be parsed to a Period: " + text);
+        Log.e("PURCHASES", "Text cannot be parsed to a Period: " + text);
+        return null;
```
vegaro, line 97: Note that now, instead of crashing, we return null and log an error

android/src/main/java/com/revenuecat/purchases/common/PurchasesPeriod.java
```diff
@@ -84,24 +87,22 @@ static PurchasesPeriod parse(@NonNull CharSequence text) {
...
         if (str == null) {
             return 0;
         }
         int val = Integer.parseInt(str);
-        try {
-            return PurchasesMath.multiplyExact(val, negate);
-        } catch (ArithmeticException ex) {
-            throw new RuntimeException("Text cannot be parsed to a Period: " + text, ex);
-        }
+        return PurchasesMath.multiplyExact(val, negate);
```
vegaro, line 105: this, would throw `ArithmeticException` which will be capture by the catch in `parse`

android/src/main/java/com/revenuecat/purchases/common/mappers.kt
```diff
@@ -195,7 +195,10 @@ private fun String?.mapPeriodDeprecated(): Map<String, Any?> {
                     "intro_price_period_number_of_units" to 0
                 )
             }
-        }
+        } ?: mapOf(
+            "intro_price_period_unit" to null,
+            "intro_price_period_number_of_units" to null
+        )
```
vegaro, line 201: here's the deal, if `PurchasesPeriod.parse` returns `null` because the text is not a Period (i.e. the text is "365" like in the reported crash), `intro_price_period_unit` and `intro_price_period_number_of_units` will be null, but the period itself (like 152 and 159) will be the string "365". So in summary, it's possible to get something like:

```
            "price": 0,
            "priceString": "$0",	            
            "period": "365",
            "cycles" to 1,         
            "intro_price_period_unit": null,
            "intro_price_period_number_of_units": null
	           
```

Which doesn't make sense. Alternatively, we could try to parse the period beforehand, and if it fails, we could nullify the whole object. What do you think??

android/src/test/java/com/revenuecat/purchases/common/PeriodTests.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import android.os.Build
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(sdk = [Build.VERSION_CODES.P])
```
vegaro, line 11: Android 29 requires JDK 9 to run tests, that's why I specified this version here

android/src/test/java/com/revenuecat/purchases/common/PeriodTests.kt
```diff
@@ -0,0 +1,24 @@
...
+class PeriodTests {
```
vegaro, line 12: Couldn't use `Spek` because it's not compatible with `Roboelectric`. Or at least I couldn't get it to work https://github.com/spekframework/spek/issues/207

aboedo: so essentially anything that logs is just not compatible with Spek? if so, we might wanna stop using it altogether

aboedo: not for this pr though

android/src/main/java/com/revenuecat/purchases/common/mappers.kt
aboedo: kinda feels weirdly inconsistent to have a "365" in there with null period and number of units. I'd vote for nullifying the whole thing if it's in an unexpected format

android/src/test/java/com/revenuecat/purchases/common/PeriodTests.kt
vegaro: I will give another try another day to try to run RoboElectric + Spek. Either way, I am still not convinced about using Spek, I had another problem couple weeks back because their IDE plugin wasn't updated and was giving me issues with regular unit tests, and I had to run all tests from the command line. Anyhow, that's a discussion for another day

vegaro: Alternative to adding RoboElectric would be to mock the Log class, which we can also do. I can try that too

vegaro:
@aboedo sure, I will change that and re-request review

vegaro:
@aboedo I used another way of mocking `Log` since I didn't want to add Roboelectric just because of this.

I nullified the period if there's an exception and added tests. I also ran the tests on `develop` to make sure I didn't break anything.

PR #58, opened 2020-10-22
Fix integration errors

aboedo:
We have build issues when trying to run the tests against purchases-ios `3.7.4`. 

This seems to have started with version `3.7.2`, the first one with static linking. It looks like there are issues with obj-c categories and static linking, particularly with objective-c extensions. There's a chance that it can be solved by linker flags as an alternate solution. 

Note that this is specific to cocoapods. The reason is that cocoapods that on other package managers, we only statically link `PurchasesCoreSwift`, but cocoapods doesn't let us statically link only one dependency, the entire chain of dependencies needs to be static. 

This PR: 
- adds a single, empty swift file to fix issues with linking against the Swift runtime in Xcode 12
- removes static linking
- bumps the `purchases-ios` repo to use a version that doesn't enforce static linking for cocoapods

Note that tests pass correctly now. 

### Requirements: 
- [ ] purchases-ios 3.7.5 release with https://github.com/RevenueCat/purchases-ios/pull/374 included


aboedo:
this is ready for review, although I'll update the purchases-ios version number once `3.7.5` is released

PurchasesHybridCommon.podspec
```diff
@@ -15,23 +15,17 @@ Pod::Spec.new do |s|
 
   s.framework      = 'StoreKit'
 
-  s.dependency 'Purchases', '3.7.2'
+  s.dependency 'Purchases'
```
aboedo, line 18: I'll lock the version to `3.7.5` once it's released

PurchasesHybridCommon.podspec
```diff
@@ -15,23 +15,17 @@ Pod::Spec.new do |s|
...
 
   s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
 
-  s.source_files = [
-    'ios/PurchasesHybridCommon/PurchasesHybridCommon/PurchasesHybridCommon.h',
-    'ios/PurchasesHybridCommon/PurchasesHybridCommon/**/*.h',
-    'ios/PurchasesHybridCommon/PurchasesHybridCommon/**/*.m'
-  ]
-
+  s.source_files = ['ios/PurchasesHybridCommon/PurchasesHybridCommon/**/*.{h,m,swift}']
```
aboedo, line 28: I added a swift file so that hopefully customers won't get the swift compilation issues anymore

ios/PurchasesHybridCommon/Podfile
```diff
@@ -1,17 +1,19 @@
...
-platform :ios, '9.0'
+  platform :ios, '9.0'
   # Comment the next line if you don't want to use dynamic frameworks
   use_frameworks!
 
   # Pods for PurchasesHybridCommon
-  pod 'Purchases', '3.7.1'
+  pod 'Purchases', branch: 'feature/make_cocoapods_dynamic_again', git:'https://github.com/revenuecat/purchases-ios'
+  pod 'PurchasesCoreSwift', branch: 'feature/make_cocoapods_dynamic_again', git:'https://github.com/revenuecat/purchases-ios'
+
```
aboedo, line 11: will update to 
```ruby
pod 'Purchases', '3.7.5'
```
once it's available

vegaro:
Looks good, I will approve when 3.7.5 is released. So the errors when running the tests were due to the static linking?

aboedo:
yeah, the linker gets confused when using categories on statically-linked frameworks. It just can't find the symbols when linking (although it finds them when compiling, so no issue there). 
So when you try to use them at runtime, you get an error saying that the method isn't there.

aboedo:
there are linker flags that should help solve it, though. However, I'm just not sure static linking is even worth it for cocoapods, since it's forced for all targets. 
I'd done static linking mostly to get around issues where the customers would have to add each module separately (for example in carthage, you'd have to add one framework per module), but cocoapods takes care of that elegantly automatically anyway.

aboedo:
@vegaro updated with `3.7.5`

vegaro (approved):

PR #62, opened 2020-12-16
fix: dates in millis are actually in seconds in iOS

aboedo:
the dates in `millis` were actually in seconds, we were never doing a conversion to millis. 
This adds the conversion and uses it everywhere.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonTests/NSDateHybridAdditionsTests.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonTests/NSDateHybridAdditionsTests.swift
@@ -41,7 +41,7 @@ class NSDateHybridAdditionsTests: QuickSpec {
                 expect(date.millisecondsSince1970()) == date.timeIntervalSince1970 * 1000.0
 
                 let now = NSDate()
-                expect(now.millisecondsSince1970) == now.timeIntervalSince1970 * 1000.0
+                expect(now.millisecondsSince1970()) == now.timeIntervalSince1970 * 1000.0
             }
         }
     }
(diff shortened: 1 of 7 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommonTests/NSDateHybridAdditionsTests.swift
```diff
@@ -34,5 +34,15 @@ class NSDateHybridAdditionsTests: QuickSpec {
...
             }
         }
+
+        describe("millisecondsSince1970") {
+            it("correctly returns results in milliseconds") {
+                let date = NSDate(timeIntervalSince1970: 1588044611)
+                expect(date.millisecondsSince1970()) == date.timeIntervalSince1970 * 1000.0
+
+                let now = NSDate()
+                expect(now.millisecondsSince1970) == now.timeIntervalSince1970 * 1000.0
```
vegaro, line 44: Why isn't this `now.millisecondsSince1970()`

vegaro (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/NSDateHybridAdditionsTests.swift
aboedo: lol good question, why did tests pass?

aboedo: they run correctly locally, it does evaluate the function! i.e.: if I change the value to something wrong, like
```swift
expect(now.millisecondsSince1970) == now.timeIntervalSince1970 * 1001.0
```
it fails as you'd expect! I wonder if there's something within Nimble that evaluates a function if it has no params and is passed into the `expect`

aboedo: yeah, apparently it's Nimble's behavior. I'll still update to call the function for the sake of clarity. 
It must be doing something pretty clever, because it does realize that the type of the right-hand side of the expression needs to be the same as the function evaluated in `expect`

PR #63, opened 2021-01-22
release/1.5.1

aboedo:
Fixes a bug where dates that were supposed to be in milliseconds in Android were actually in seconds.


aboedo:
tests are failing because I'm not taking timezones into consideration yet, fixing

aboedo:
I realize that there are newer versions of `purchases-ios` and `purchases-android`, but I'm going for a quick hotfix, so I don't want to deviate a ton and integrate the rest of the changes yet.

android/src/main/java/com/revenuecat/purchases/common/mappers/mappersHelpers.kt
```diff
@@ -42,6 +42,6 @@ fun JSONObject.convertToMap(): Map<String, String?> =
         }
     }
 
-internal fun Date.toMillis(): Double = this.time.div(1000.0)
+public fun Date.toMillis(): Long = this.time
```
aboedo, line 45: I think this bug might have been around forever

aboedo: the iOS equivalent was fixed in 3.0.0, but since this function was called `toMillis` I assumed it was correct 🤦

vegaro (approved):

PR #68, opened 2021-03-22
Feature: identity v3

aboedo:
Adds: 
- `logIn`
- `logOut`

Deprecates: 
- `allowSharingAppStoreAccount`
- `identify`
- `reset`

### Requirements: 
- [ ] `purchases-ios` release with identity v3 support
- [ ] `purchases-android` release with identity v3 support


ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Mocks/MockPurchases.swift
```diff
@@ -0,0 +1,491 @@
+//
+// Created by Andrés Boedo on 4/7/21.
+// Copyright (c) 2021 RevenueCat. All rights reserved.
+//
+
+import Foundation
+import Purchases
+
+class MockPurchases: Purchases {
```
aboedo, line 9: big mock here. Mostly autogenerated, but I fixed some of the names by hand. Only a part of it is used, but I figured it's safer to do a full mock and not risk calling real methods. 
Also, this way we already have stuff set up for future tests.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchaserInfoHybridAdditionsTests.swift
```diff
@@ -78,7 +80,7 @@ class PartialMockPurchaserInfo: Purchases.PurchaserInfo {
     }
     
     override var firstSeen: Date {
-        return Date()
+        return _firstSeen
```
aboedo, line 83: having these return on-the-fly would make it impossible to compare the generated dictionaries of them with `==`, since each time `dictionary` gets called it has a new value

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests-Bridging-Header.h
```diff
@@ -3,3 +3,4 @@
 //
 
 #import <PurchasesHybridCommon/RCPurchases+HybridAdditions.h>
+#import <PurchasesHybridCommon/RCPurchases+ProtectedExtensions.h>
```
aboedo, line 6: protected extension so that we can mock the shared instance

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -186,6 +190,14 @@ fun reset(
     }
 }
 
+fun logOut(
+    onResult: OnResult
+) {
```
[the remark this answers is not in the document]
beylmk: it doesn't look like we have anything configured in purchases-android.... do you guys have a style config somewhere other than git?

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -33,4 +44,199 @@ internal class CommonKtTests {
...
+
+        configure(
+            context = mockContext,
+            apiKey = "api_key",
+            appUserID = "appUserID",
+            observerMode = true,
+            platformInfo = PlatformInfo("flavor", "version")
+        )
+
+        every { Purchases.sharedInstance } returns mockPurchases
```
beylmk, line 79: i'm new to mockk, but couldn't you pull the `every { Purchases.sharedInstance } returns mockPurchases` to setup() if you call it in each test?

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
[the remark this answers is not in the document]
vegaro: we do have it in purchases-android and it should work automatically https://github.com/RevenueCat/purchases-android/tree/develop/.idea/codeStyles but we just recently added this so it's possible it doesn't work as we thought. We should add the same to this repo. 

This indentation is fine though, if using multi line parameters this is the correct indentation. It can also be define as a single line:

```
fun logOut(onResult: OnResult) {
```

And the formatter should not complain. I am going to add to the Kotlin conventions Notion page to do mandatory multi-line for methods with multiple arguments, and optional multiline for single argument methods.

ios/PurchasesHybridCommon/PurchasesHybridCommon/RCCommonFunctionality.h
```diff
@@ -11,19 +11,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 typedef void (^RCHybridResponseBlock)(NSDictionary * _Nullable, RCErrorContainer * _Nullable);
 
+
 @interface RCCommonFunctionality : NSObject
 
-@property(class, nonatomic, nullable, copy) NSString *proxyURLString;
+@property (class, nonatomic, nullable, copy) NSString *proxyURLString;
```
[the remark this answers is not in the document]
vegaro: Thanks for fixing it!

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -303,7 +328,7 @@ private fun getProductChangeCompletedFunction(onResult: OnResult): (PurchaseDeta
     }
 }
 
-private fun PurchasesError.map(
+fun PurchasesError.map(
```
vegaro, line 331: Since this is only used in this module, it can be internal instead of public

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -33,4 +44,199 @@ internal class CommonKtTests {
...
     }
-}
\ No newline at end of file
+
+    private val mockApplicationContext = mockk<Application>(relaxed = true)
+    private val mockContext = mockk<Context>(relaxed = true)
+    private val mockPurchases = mockk<Purchases>()
+
+    @BeforeEach
+    fun setup() {
```
vegaro, line 53: I would add this `setup` at the top, since it will also be executed before the 3 tests that were already in this file

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -33,4 +44,199 @@ internal class CommonKtTests {
...
+        every {
+            mockPurchases.logIn(
+                newAppUserID = appUserID,
+                capture(logInCallback)
+            )
+        } just runs
+
+        val onResult = mockk<OnResult>()
+        every { onResult.onReceived(any()) } just runs
+        every { onResult.onError(any()) } just runs
```
vegaro, line 117: You could also do `mockk<OnResult>(relaxed = true)` here if you don't really care about any parameter. Your call. I tend to avoid `relaxed` because it can have unwanted results due to the default values it mocks, but in this case I think it's fine since the behaviour of the mock doesn't really matter for the test and you're just checking is being called in a way.

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
aboedo: great catch, I forgot that the default for Kotlin is `public` and assumed it'd be `internal` like swift

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
aboedo: agreed

aboedo:
@vegaro @beylmk I think that's all comments addressed

vegaro (approved):

aboedo:
merging in spite of the HOLD, since we only need to HOLD from releasing, but not really from merging

PR #70, opened 2021-03-27
Spek removal

vegaro:
Removes Spek from the project and migrates unit tests to JUnit5

### DIFF
--- a/android/src/test/java/com/revenuecat/purchases/common/CommonKtTests.kt
+++ b/android/src/test/java/com/revenuecat/purchases/common/CommonKtTests.kt
@@ -2,16 +2,19 @@ package com.revenuecat.purchases.common
 
 
 import com.revenuecat.purchases.Purchases
-import org.spekframework.spek2.Spek
-import org.spekframework.spek2.style.specification.describe
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
 import java.net.URL
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
-import kotlin.test.expect
 
-object CommonKtTests : Spek({
-    describe("setProxyURLString") {
-        it("sets the proxyURL correctly from a valid URL") {
+internal class CommonKtTests {
+    @Nested
+    @DisplayName("Calling setProxyURLString")
+    inner class SetProxyURLString {
+        @Test
+        fun `sets the proxyURL correctly from a valid URL`() {
             assertEquals(Purchases.proxyURL, null)
 
             val urlString = "https://revenuecat.com"
@@ -20,7 +23,8 @@ object CommonKtTests : Spek({
             assertEquals(Purchases.proxyURL.toString(), urlString)
         }
 
-        it("sets the proxyURL to null from a null string") {
+        @Test
+        fun `sets the proxyURL to null from a null string`() {
             Purchases.proxyURL = URL("https://revenuecat.com")
 
             setProxyURLString(null)
@@ -28,10 +32,11 @@ object CommonKtTests : Spek({
             assertEquals(Purchases.proxyURL, null)
         }
 
-        it("raises exception if url string can't be parsed into a URL") {
+        @Test
+        fun `raises exception if url string can't be parsed into a URL`() {
             assertFailsWith<java.net.MalformedURLException> {
                 setProxyURLString("this is not a url")
             }
         }
     }
-})
+}
--- a/android/src/test/java/com/revenuecat/purchases/common/CommonKtTests.kt
+++ b/android/src/test/java/com/revenuecat/purchases/common/CommonKtTests.kt
@@ -10,33 +10,29 @@ import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 
 internal class CommonKtTests {
-    @Nested
-    @DisplayName("Calling setProxyURLString")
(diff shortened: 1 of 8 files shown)


android/src/test/java/com/revenuecat/purchases/common/CommonKtTests.kt
```diff
@@ -2,16 +2,19 @@ package com.revenuecat.purchases.common
...
-import kotlin.test.expect
 
-object CommonKtTests : Spek({
-    describe("setProxyURLString") {
-        it("sets the proxyURL correctly from a valid URL") {
+internal class CommonKtTests {
+    @Nested
+    @DisplayName("Calling setProxyURLString")
+    inner class SetProxyURLString {
+        @Test
```
aboedo, line 16: this is interesting, we haven't been splitting up test cases into groups like this, right? I like it. 
a slight advantage that Spek had was that you'd be able to very quickly create setup and tear down code that was specific to each group, does this structure still allow for that? 

I'm asking out of sheer curiosity because we weren't really taking advantage of the grouped setup and tear down anyway

aboedo (commented):
thanks for doing this!

android/src/test/java/com/revenuecat/purchases/common/ConfiguringUnitTests.kt
```diff
@@ -7,20 +7,22 @@ import io.mockk.every
...
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
 
-object ConfiguringUnitTests : Spek({
-    describe("when setting up Purchases") {
+internal class ConfiguringUnitTests {
+    @Nested
+    @DisplayName("when setting up Purchases")
+    inner class ParsingInvalidPurchasesPeriod {
```
beylmk, line 18: i might just be missing something, but does this class name still make sense? i'm not seeing us parsing purchases periods in this class?

android/src/test/java/com/revenuecat/purchases/common/CommonKtTests.kt
vegaro: So we are using JUnit 5 for these tests, in `purchases-android` we still use JUnit 4, which doesn't let us do that. In this particular test suite doesn't really make sense to use this inner class, but in `SkuDetailsMapperTests` for example, it makes sense also for organization of the results. Android Studio will show the divided by groups which is nice. We can also do `@BeforeEach` and `@AfterEach` in each of these inner classes. I set it up following this article https://medium.com/@boonkeat/android-unit-testing-with-junit5-d1b8f9c620b6, in case you want to check it out.

beylmk: does the @nested with @displayname gets us any added functionality here (given that there's only one "nested" section in the file?)?  i see it's advantage for readability/future refactoring, but curious if there's any functionality or output associated with it..

beylmk: @vegaro posted that comment before seeing your response, feel free to ignore

vegaro: it shows the "nested" as a group, but being the single group, I don't see any benefit, so I am going to remove it. It would show similar to this:

android/src/test/java/com/revenuecat/purchases/common/ConfiguringUnitTests.kt
vegaro: no, it doesn't make sense. Fried Brain Fridays haha

android/src/test/java/com/revenuecat/purchases/common/MappersTests.kt
```diff
@@ -2,27 +2,27 @@ package com.revenuecat.purchases.common
...
+import org.junit.jupiter.api.Test
 import java.text.SimpleDateFormat
 import java.util.*
 
-object MappersHelpersTests : Spek({
+internal class MappersHelpersTests {
 
-    describe("when mapping dates to milliseconds") {
+    @Test
+    fun `PurchasesPeriod gets parsed correctly`() {
```
aboedo, line 14: this class doesn't really specify anything related to `PurchasesPeriod`, I think we should rename the tests accordingly - perhaps something like
```kotlin
fun `parsing simple date format returns the correct date`() {
```

android/src/test/java/com/revenuecat/purchases/common/PeriodTests.kt
```diff
@@ -2,23 +2,24 @@ package com.revenuecat.purchases.common
...
-object PeriodTests : Spek({
-    describe("when parsing an invalid PurchasesPeriod") {
-        beforeEachTest {
-            mockLogError()
-        }
-        it("there is no exception") {
-            val period = PurchasesPeriod.parse("365")
-            assertThat(period).isNull()
-        }
+internal class PeriodTests {
```
aboedo, line 10: FFTI: Should we name these `PurchasesPeriodTests` so they match the class they test more closely?

beylmk (approved):
👍  thank you!

aboedo (approved):

android/src/test/java/com/revenuecat/purchases/common/MappersTests.kt
vegaro: my brain was really fried

PR #64, opened 2021-01-29
Amazon support

vegaro:

PR #72, opened 2021-04-02
Android: SkuDetails -> ProductDetails

aboedo:
### Changes Summary (Android only)

- Updates `purchases-hybrid-common` on android for compatibility with the latest `develop` version. Copies the logic from the Amazon support PR (#64)
- Renames the output package of `purchases-hybrid-common` for android from `com.revenuecat.purchases.common` -> `com.revenuecat.purchases.hybridcommon` because there's already a `com.revenuecat.purchases.common` in the main SDK in `purchases-android`
- Updates the dependency to purchases-android to use a git dependency on `develop`. This is so that the changes from the Amazon PR get included, like SkuDetails. Documentation on the [method used can be found here.](https://blog.gradle.org/introducing-source-dependencies) 
- To achieve this, this PR also solves a couple of issues with the dependency management: we were previously leveraging [Maven's behavior of importing all transitive dependencies](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html) to be able to use just about anything in the main SDK. But this would mean that it only works when using Maven, but not when using a local Gradle build. To fix this, this PR explicitly adds dependencies to `purchases-android`'s submodules when they're needed.


android/src/main/java/com/revenuecat/purchases/common/mappers/mappersHelpers.kt
```diff
@@ -3,6 +3,8 @@ package com.revenuecat.purchases.common.mappers
 import com.revenuecat.purchases.utils.Iso8601Utils
```
aboedo, line 3: this is one of the lines that won't compile now for some reason

android/settings.gradle
```diff
@@ -1 +1,6 @@
-rootProject.name = 'purchases-hybrid-common'
\ No newline at end of file
+rootProject.name = 'purchases-hybrid-common'
+sourceControl {
+    gitRepository("https://github.com/revenuecat/purchases-android.git") {
+        producesModule("com.revenuecat.purchases:purchases")
+    }
```
aboedo, line 5: found this and it seems to work as expected. well, other than the fact that it doesn't compile XD but it seems unrelated

android/src/main/java/com/revenuecat/purchases/common/common.kt
```diff
@@ -279,7 +279,18 @@ private fun getPurchaseErrorFunction(onResult: OnResult): (PurchasesError, Boole
     return { error, userCancelled -> onResult.onError(error.map(mapOf("userCancelled" to userCancelled))) }
 }
 
-private fun getPurchaseCompletedFunction(onResult: OnResult): (Purchase?, PurchaserInfo) -> Unit {
```
aboedo, line 282: I can't add a comment on the line because it wasn't changed, but line 266 on this file is the other place where compilation now fails because the symbol can't be found

android/settings.gradle
vegaro: wutttttt this is great

vegaro: ok, so I think this is actually the reason why it's not working. It's related to that issue I mentioned where all the classes are being accessible when downloading the SDK from Maven. 

`PlatformInfo` and `Iso8601Utils` are not in the `:public` not in `:purchases` module. If you look at the build.gradle in `:purchases`:

```
dependencies {
    implementation fileTree(include: ['*.jar'], dir: 'libs')
    implementation project(":common")
    implementation project(":utils")
    implementation project(":feature:subscriber-attributes")
    implementation project(":feature:google")
    implementation project(":feature:identity")
    implementation project(":strings")
    api project(":public")
```

The only one we expose when someone adds the module `:purchases` is `:public`, so when building the project locally (and it looks like when downloading from git too) `purchases-hybrid-common` shouldn't have access to classes outside `:purchases` or `:public`. When adding `purchases` from maven, everything is exposed.

This is actually how I found out that everything was being exposed. I tried to add the local dependency and `PlatformInfo` was not accessible. But when downloading from Maven, it is.

aboedo: bummer, but that makes total sense. It also does mean that we don't have a great way of testing this while developing other than issuing a pre-release, right? 
let's sync up about it today

android/build.gradle
```diff
@@ -56,11 +57,33 @@ android {
...
+        version {
+            branch = 'develop'
+        }
+    }
+
+    implementation('com.revenuecat.purchases:utils') {
+        version {
+            branch = 'develop'
+        }
+    }
```
aboedo, line 84: maybe there's a nicer way to group these? I don't know how flexible Gradle is in terms of loops or lists

aboedo (commented):
no idea why GitHub shows an error in the checks, but the tests did pass correctly in CircleCI ¯\_(ツ)_/¯

android/build.gradle
```diff
@@ -8,15 +8,16 @@ buildscript {
...
+    // ext.purchases_version = '4.0.5'
 
     repositories {
         google()
         jcenter()
 
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:$android_tools"
+        classpath "com.android.tools.build:gradle:${android_tools}"
```
beylmk, line 20: just curious why this one needs curly braces?

android/build.gradle
```diff
@@ -56,11 +57,32 @@ android {
...
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    api "com.revenuecat.purchases:purchases:$purchases_version"
-    
+    implementation 'androidx.annotation:annotation:1.2.0'
+
+    // TODO: un-comment the api line and remove the implementation once purchases-android is released
+    // api "com.revenuecat.purchases:purchases:$purchases_version"
+    implementation('com.revenuecat.purchases:purchases') {
+        version {
+            branch = 'develop'
```
beylmk, line 70: this is really cool, thanks for sharing the article

android/build.gradle
aboedo: I actually don't think it does - this was left behind after changing a lot of stuff back and forth, so I don't even recall how it got there. I'll revert this line

android/build.gradle
beylmk: i'm realizing how much i need to brush up on gradle..

beylmk:
I just want to check my understanding on what’s happening here, since I’m only just starting to understand how hybrid SDKs work… 

* The Unity SDK interfaces with the `android` portion of the `purchases-hybrid-common` library, which maps objects to the format Unity expects
* As part of the Unity work, we updated SkuDetails in `purchases-android` to ProductDetails (appears to just be a rename?)
* We ran into dependency issues when consuming Unity work because it isn’t in an official release on Maven, you needed to build locally, and the way we were depending back onto `purchases-android` wasn’t exposing certain dependencies 

Is that all correct? And will we need to do the same updates for the iOS portion of `purchases-hybrid-common`? 

Sorry if these questions are silly, I definitely need to spend more time understanding how everything works together...

aboedo:
Absolutely not silly questions! And this situation is not a particularly simple one. Answers below: 

> The Unity SDK interfaces with the android portion of the purchases-hybrid-common library, which is a wrapper of the android native code adhering to a uniform interface across iOS/android

This is correct. Note that all of the issues described in this PR apply to all hybrids, not just Unity. 

> we updated SkuDetails in purchases-android to ProductDetails (appears to just be a rename?)

Yep, it's mostly a rename. The motivation was for Amazon support - SkuDetails is a Google Billing class, and we didn't wan to have Amazon apps reference Google Billing types. So ProductDetails is sort of a generalization of both SkuDetails and Amazon's products, which can be referenced without having to directly reference the Amazon or Google Billing libraries. 

> We ran into dependency issues when consuming Unity work because it isn’t in an official release on Maven, you needed to build locally, and the way we were depending back onto purchases-android wasn’t exposing certain dependencies. Is that all correct? 

Yep. On Maven, when you depend on a module, all of its dependencies get exposed as well, so even if we're not explicitly adding dependencies on sub-modules, we can use them. However, when trying to use them from a non-Maven build, we have to rely on Gradle directly. And Gradle dependencies won't expose everything, but rather expose only the public stuff. 
`implementation` will expose only the public stuff from the specified module, and `api` will expose the public stuff from transitive dependencies as well. But you still have to explicitly define a dependency to the sub-modules in Gradle.

> And will we need to do the same updates for the iOS portion of purchases-hybrid-common?
No... Mostly because our iOS dependencies work differently. We use Cocoapods, and it already provides a nice unified experience between local builds, git branch dependencies and published pods. 
So it's already taken care of there. Also, on iOS there are no changes for Amazon, so the `SkuDetails` -> `ProductDetails`
thing didn't happen. 

Hope this helps! Happy to answer more questions

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/mappersHelpers.kt
```diff
@@ -42,6 +44,12 @@ fun JSONObject.convertToMap(): Map<String, String?> =
         }
     }
 
-public fun Date.toMillis(): Long = this.time
+fun Date.toMillis(): Long = this.time
```
vegaro, line 47: can this be internal?

android/src/test/java/com/revenuecat/purchases/hybridcommon/PurchaserInfoMappersTests.kt
```diff
@@ -1,16 +1,14 @@
...
+package com.revenuecat.purchases.hybridcommon
 
 import android.net.Uri
 import com.revenuecat.purchases.PurchaserInfo
-import com.revenuecat.purchases.common.mappers.map
-import com.revenuecat.purchases.common.mappers.toIso8601
-import com.revenuecat.purchases.common.mappers.toMillis
+import com.revenuecat.purchases.hybridcommon.mappers.toIso8601
+import com.revenuecat.purchases.hybridcommon.mappers.toMillis
+import com.revenuecat.purchases.hybridcommon.mappers.*
```
vegaro, line 7: Is this needed?

aboedo: this might have been just me trying stuff out until something worked and forgetting to revert. I'll try reverting

vegaro (approved):

PR #74, opened 2021-04-27
Fix: add prefix to foundation extensions

aboedo:
purchases-hybrid-common equivalent of https://github.com/RevenueCat/purchases-ios/pull/500

adds `rc_` prefix to all foundation extensions to prevent naming collisions, renames `millisecondsSince1970` to prevent collisions in return type


ios/PurchasesHybridCommon/PurchasesHybridCommon/NSDate+HybridAdditions.m
```diff
@@ -25,11 +25,11 @@
 
 @implementation NSDate (RCExtensions)
 
-- (NSString *)formattedAsISO8601 {
+- (NSString *)rc_formattedAsISO8601 {
```
beylmk, line 28: is it worth adding asString

beylmk (approved):
📖

vegaro (approved):

PR #75, opened 2021-05-04
temporarily remove identity v3

aboedo:
I merged the identity v3 stuff a little too optimistically, assuming that the native SDKs would be deployed right away, but this forced us to point to a branch that's in development, and base all other branches off of that. 

This removes all identity v3 stuff temporarily until it gets deployed in the native SDKs. When that happens, we can just revert this PR and we'll be ready to go.


android/build.gradle
```diff
@@ -63,25 +62,9 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     implementation 'androidx.annotation:annotation:1.2.0'
 
-    // TODO: un-comment the api line and remove the implementation once purchases-android is released
-    // api "com.revenuecat.purchases:purchases:$purchases_version"
```
beylmk, line 67: maybe we can document how this works somewhere in notion? ideally it wouldn't be pushed again but i could see it being useful for local development

beylmk (approved):
◀️  looks good!

aboedo:
this is interesting: it looks like defining all of the dependencies actually causes them to be identified as duplicate in maven. it makes sense, but it's unfortunate. I'll just roll back to the old way of defining the dependency for now.

vegaro:
> this is interesting: it looks like defining all of the dependencies actually causes them to be identified as duplicate in maven. it makes sense, but it's unfortunate. I'll just roll back to the old way of defining the dependency for now.

I guess this is because they are already downloaded transitively when downloading `purchases` right?

vegaro (approved):

PR #77, opened 2021-05-11
Add canMakePayments method

beylmk:
Adds common functionality for calling native canMakePayments methods

Will need to update purchases-android dependency (once it's released) before merging


android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -258,6 +260,23 @@ fun invalidatePurchaserInfoCache() {
     Purchases.sharedInstance.invalidatePurchaserInfoCache()
 }
 
+fun canMakePayments(context: Context,
+                    features: List<String>,
+                    onResult: OnResultAny<Boolean>) {
```
aboedo, line 265: 🤔 
This would break when we add Amazon, right? 
I don't see much of a way around that, tbh, at least until the Amazon stuff gets merged. 

On another note, I feel like this might lead to weird usage from a customer's perspective: 
For example, they might be using react-native and have to call `Purchases.canMakePayments("SUBSCRIPTIONS")`, while making sure that they're passing in names that are compatible with Android's `BillingFeature`, right? 

Perhaps we can create an abstraction of that? Either by deciding which features we want to check ourselves, or by having an enum or something abstracting the Google Billing types? That way React-Native users on iOS won't have to look up Android's `BillingFeatures` to call, and it also gives us a clean way of separating Amazon vs Google Play features?

beylmk: @aboedo so i do have a BillingFeature enum on the ReactNative side, and was planning to do that for each hybrid. So the call for an end user would use an enum type. And then I have the BillingFeature enum in `purchases-android`, which is the one I mapped to here.

Re-declaring our purchases enums seemed to be the norm for other enums in ReactNative, I assume because enum types aren't compatible across all the languages? We do that for PURCHASE_TYPE, for instance, and pass a string to  `common` methods expecting it, and then the mapping back to the native enum happens there. 

So there really shouldn't be a reason that the InvalidArgumentException is thrown, since the interfaces in each hybrid will enforce the correct mappings. Let me know if I just missed something? I can open the ReactNative PR for reference, too 

Could you elaborate on why this would break Amazon?

aboedo: That makes sense. 

And yeah, enums wouldn't really work here, don't know what I was thinking. 

> Could you elaborate on why this would break Amazon?

🤔 I should clarify: I don't think this would break Amazon, rather, we'd have to add logic to make a different call for Amazon, since `BillingFeature` is a Google Play thing, right? I'm not sure Amazon even provides a method to check if purchases are allowed on the device. 

@vegaro might know more about that part, but we probably won't have to worry about it until Amazon is ready to ship

beylmk: ahhh i see. maybe we could have a check for the store and just return an error if amazon. let me know your thoughts @vegaro

vegaro: Good catch @aboedo. It's tricky. What we would normally do is `canMakePayments` would be an abstract function in [BillingAbstract](https://github.com/RevenueCat/purchases-android/blob/develop/common/src/main/java/com/revenuecat/purchases/common/BillingAbstract.kt) and then the Amazon module would have an implementation different to the one Google one.

Since `canMakePayments` lives in Purchases and it's actually a static call, it doesn't use an instance of BillingClient, it can't be a static abstract function. I also don't think Amazon provides an equivalent to `canMakePayments` 🤔 Maybe we could check if the package of the store is present? Probably error prone though...

vegaro: Re-reading this. @aboedo I don't think this code in `purchases-hybrid-common` would break in Amazon, since `BillingFeature` is an enum in `purchases-android`. The call to `Purchases.Companion.canMakePayments` will break though, because its implementation assumes that the BillingClient is  able to connect to play services

beylmk: @aboedo I discussed this with Cesar. Proposed plan is:
* add a new overload of this function that takes a store parameter
* default that store parameter to Play for now, and return false if it's not Play
* once that's in, i'll update purchases-hybrid-common and hybrid repos off the amazon branches

does that sound ok to you?

aboedo: Sounds great! 👌

android/build.gradle
```diff
@@ -63,7 +63,9 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.2.0'
 
     api("com.revenuecat.purchases:purchases:$purchases_version")
```
aboedo, line 65: should this one actually be `implementation`?

beylmk: i had the same question :) this is so hybrids can just add `purchases-hybrid-common` and don't also have to add purchases.

android/build.gradle
```diff
@@ -63,7 +63,9 @@ dependencies {
...
-    
+    implementation("com.revenuecat.purchases:purchases-core-common:$purchases_version")
```
beylmk, line 66: to get purchases-flutter working pointing to local hybrid-common, i actually had to make both of these api... 

i guess we can test purchases-flutter using a hybrid-common snapshot version before we release this

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -261,12 +261,13 @@ fun invalidatePurchaserInfoCache() {
...
 fun canMakePayments(context: Context,
-                    features: List<String>,
+                    features: List<Int>,
                     onResult: OnResultAny<Boolean>) {
     val billingFeatures = mutableListOf<BillingFeature>()
     try {
-        billingFeatures.addAll(features.map { BillingFeature.valueOf(it) })
-    } catch (e: IllegalArgumentException) {
+        val billingFeatureEnumValues = BillingFeature.values()
+        billingFeatures.addAll(features.map { billingFeatureEnumValues[it] })
```
beylmk, line 269: @aboedo Here's that refactor we talked about. When I make the update to `purchases-android` to include a store parameter, we could consider updating the BillingFeature enum there to take a `code` the way we did for `PurchasesError`. I'm not sure that's necessary though

beylmk:
@aboedo @vegaro when you get a chance could you re-review?

vegaro:
I want to put a comment in `.gradle/buildOutputCleanup/buildOutputCleanup.lock `but I can't because it's a binary 😆 Anyhow, I don't think you meant to commit that file

android/build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     ext.assertj_version = '3.13.2'
     ext.mockk_version = '1.10.0'
     ext.gradle_maven_publish = '0.11.1'
-    ext.purchases_version = '4.1.0'
+    ext.purchases_version = '4.2.0'
```
vegaro, line 11: We can update this to 4.2.1 now 😄

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -65,4 +66,113 @@ internal class CommonKtTests {
             setProxyURLString("this is not a url")
         }
     }
+
+    @Test
+    fun `calling canMakePayments correctly passes through to Purchases`() {
```
vegaro, line 71: This test is the same as `calling canMakePayments correctly passes through to Purchases` right? The latter is actually more complete

vegaro (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/RCCommonFunctionality.h
```diff
@@ -108,10 +108,10 @@ signedDiscountTimestamp:(nullable NSString *)discountTimestamp
 + (void)setAd:(nullable NSString *)ad;
 
 + (void)setKeyword:(nullable NSString *)keyword;
-
+
```
aboedo, line 111: this line can probably be reverted?

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -65,4 +66,113 @@ internal class CommonKtTests {
...
+        configure(
+                context = mockContext,
+                apiKey = "api_key",
+                appUserID = "appUserID",
+                observerMode = true,
+                platformInfo = PlatformInfo("flavor", "version")
+        )
+
+        every { Purchases.canMakePayments(mockContext, any(), any()) } just runs
```
aboedo, line 80: this will always return the same value, right? and we're not really checking that the value returned by `Purchases.canMakePayments` is the same one returned by `CommonKt.canMakePayments`. 
Perhaps we could have a random bool variable here, have `Purchases.canMakePayments` return the random bool, and check that the same value is received in `onResult.onReceived` (instead of matching to `any()`

aboedo (approved):
left small comments, but looks ready to go

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -65,4 +71,120 @@ internal class CommonKtTests {
             setProxyURLString("this is not a url")
         }
     }
+
+    @RepeatedTest(5)
```
beylmk, line 75: @aboedo found this "repeatedtest" thing so that this test will run 5 times. thought it was a good way to check that we don't just get lucky with the random bool, since i t's a 50/50 chance. other option would be to add this verification to each of the other unit tests and take advantage of them running anyway, but that's only 3x. let me know what you think!

PR #82, opened 2021-06-24
Adds readableErrorCode to UserInfo when creating ErrorContainer

vegaro:
Should fix one of the issues related to https://github.com/RevenueCat/react-native-purchases/issues/266

We create a dictionary with extra info and we add `readableErrorCode` to it. The issue is that we don't pass that dictionary when rejecting the promise in Rn, and we pass the original error when rejecting (https://github.com/RevenueCat/react-native-purchases/blob/develop/ios/RNPurchases.m#L354) and the original error doesn't have `readableErrorCode`, it has `readable_error_code` instead. We don't want the snake case version because it doesn't match the casing of the other properties of the error.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/RCErrorContainer.m
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/RCErrorContainer.m
@@ -30,8 +30,15 @@
     if (error.userInfo[RCReadableErrorCodeKey]) {
         NSString *readableErrorCode = error.userInfo[RCReadableErrorCodeKey];
         dict[@"readableErrorCode"] = readableErrorCode;
-        dict[@"readable_error_code"] = readableErrorCode;
-            
+        dict[RCReadableErrorCodeKey] = readableErrorCode;
+        
+        // Reason behind this is because React Native doens't let reject the promises passing more information
+        // besides passing the original error, but it passes the extra userInfo from that error to the JS layer.
+        // Since we want to pass both readable_error_code (deprecated) and readableErrorCode when building the
+        // error JS object, and the error coming from purchases-ios only has the snake case version, we need to
+        // add readableErrorCode to the userInfo of the error. In a future project, we will remove the
+        // deprecated version and also improve error handling so it's easier to detect which errors come
+        // from RevenueCat and which don't
         NSMutableDictionary *fixedUserInfo = [NSMutableDictionary dictionaryWithDictionary:error.userInfo];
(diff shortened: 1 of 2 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/RCErrorContainer.m
```diff
@@ -28,8 +28,14 @@ - (instancetype)initWithError:(NSError *)error extraPayload:(NSDictionary *)extr
     }
 
     if (error.userInfo[RCReadableErrorCodeKey]) {
-        dict[@"readableErrorCode"] = error.userInfo[RCReadableErrorCodeKey];
-        dict[@"readable_error_code"] = error.userInfo[RCReadableErrorCodeKey];
+        NSString *readableErrorCode = error.userInfo[RCReadableErrorCodeKey];
+        dict[@"readableErrorCode"] = readableErrorCode;
+        dict[@"readable_error_code"] = readableErrorCode;
```
aboedo, line 33: since this is the same as `RCReadableErrorCodeKey` [as declared here](https://github.com/RevenueCat/purchases-ios/blob/d0ac5b9b422207f7c710d04cb77899e4e77d9495/Purchases/Public/RCPurchasesErrorUtils.m#L22), should we be using that instead?

ios/PurchasesHybridCommon/PurchasesHybridCommon/RCErrorContainer.m
```diff
@@ -28,8 +28,14 @@ - (instancetype)initWithError:(NSError *)error extraPayload:(NSDictionary *)extr
...
+            
+        NSMutableDictionary *fixedUserInfo = [NSMutableDictionary dictionaryWithDictionary:error.userInfo];
+        fixedUserInfo[@"readableErrorCode"] = readableErrorCode;
```
aboedo, line 36: should we add a comment explaining _why_ this is a fix for context for future devs?

vegaro: Sure! I also wanted to clarify that I solved it this way because the other fix would be to change the `RCReadableErrorCodeKey` and that would require updating all hybrids. This way we can just fix it for now, and think about a better solution for errors in that project we have to fix error handling in the hybrids 😄

aboedo (approved):

taquitos:
Just noticed, you merged this into `develop` you’ll want to cherry-pick that into `main` :grin:

vegaro:
Oh develop is still the default branch in this repo

PR #83, opened 2021-07-07
ruby dependency updates

aboedo:
we had an automated security warning from GitHub because of a vulnerability in `rexml`. This PR updates dependencies, with `rexml` in particular being upgraded to `3.2.5` which has the fix.


taquitos (approved):
YOU GET AN UPDATE, YOU GET AN UPDATE, YOU GET AN UPDATE, YOU ALL GET UPDATES! 🐝🐝🐝🐝🐝🐝🐝🐝🐝🐝🐝
:shipit:

PR #85, opened 2021-07-29
Fix: error codes mismap android

aboedo:
Addresses https://github.com/RevenueCat/purchases-flutter/issues/222

We have a bug in the code that maps `PurchasesErrorCode` -> `ErrorContainer` -> `Map`. 
We're using `ordinal` to find the value of the error code enum, however, since there's no value for code `13`, this maps the wrong error code for all errors after `13`. 

Fixed by replacing usage of `ordinal` with `.code`.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -355,10 +355,10 @@ internal fun PurchasesError.map(
     extra: Map<String, Any?> = mapOf()
 ): ErrorContainer =
     ErrorContainer(
-        code.ordinal,
+        code.code,
         message,
         mapOf(
-            "code" to code.ordinal,
+            "code" to code.code,
             "message" to message,
             "readableErrorCode" to code.name,
             "readable_error_code" to code.name,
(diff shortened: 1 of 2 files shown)


android/src/test/java/com/revenuecat/purchases/hybridcommon/MappersTests.kt
```diff
@@ -23,4 +25,14 @@ internal class MappersHelpersTests {
         assertThat(date3?.toMillis()).isEqualTo(649382400000L)
     }
 
+    @Test
+    fun `purchasesErrors are mapped correctly`() {
```
aboedo, line 29: confirmed that this test fails on `develop`

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -355,10 +355,10 @@ internal fun PurchasesError.map(
     extra: Map<String, Any?> = mapOf()
 ): ErrorContainer =
     ErrorContainer(
-        code.ordinal,
+        code.code,
         message,
         mapOf(
-            "code" to code.ordinal,
```
beylmk, line 361: great find :)

beylmk (approved):
🔢

PR #88, opened 2021-08-17
Makes getPaymentDiscount return an error in android

vegaro:
When adding susbcription offers support for Unity, I want to return an error if calling `getPaymentDiscount` from Android. I don't want to create the error in the Unity code, but I want to create it in this library so the behavior is the same in all platforms. I also noticed Flutter no-ops in Android which is not ideal.

### DIFF
--- a/android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
+++ b/android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
@@ -7,6 +7,7 @@ import com.revenuecat.purchases.BillingFeature
 import com.revenuecat.purchases.Purchases
 import com.revenuecat.purchases.PurchaserInfo
 import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesErrorCode
 import com.revenuecat.purchases.common.PlatformInfo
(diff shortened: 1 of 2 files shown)


android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -360,4 +363,12 @@ internal class CommonKtTests {
         verify(exactly = 1) { onResult.onError(mockErrorMap) }
     }
 
+    @Test
+    fun `getPaymentDiscount returns an error`() {
+        val error = getPaymentDiscount()
+        assertNotNull(error)
```
taquitos, line 369: I usually try to leave out null check assertions if I'm going to be asserting the value of the object later on. Is there extra value to null checking in kotlin?

taquitos (approved):
🩰 🐐  

One question, but otherwise 🚢

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
vegaro: Good point, getPaymentDiscount will always return a not null object actually since it's marked as non-nullable in Kotlin. Removing!!

PR #89, opened 2021-08-17
Release 1.9.0

vegaro:
Made it a minor since we are changing the Android API but open to do it patch

### 1.9.0

- Adds `getPaymentDiscount` in Android, which returns an error
    https://github.com/RevenueCat/purchases-hybrid-common/pull/88

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 1.9.0
+
+- Adds `getPaymentDiscount` in Android, which returns an error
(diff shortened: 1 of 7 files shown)


taquitos (approved):
IMO, a minor makes sense, we're adding API, not just doing a bug fix/internal cleanup patch.
Also, we can increment minors for days, so I'm not worried about running out of numbers 😄 
👯 🐐 👯

PR #97, opened 2021-11-18
Upgrades publish plugin

vegaro:
Counterpart of https://github.com/RevenueCat/purchases-android/pull/310 so we don't have to go to Sonatype anymore


taquitos (approved):
🅾️🐐

RELEASING.md
```diff
@@ -10,10 +10,4 @@
 - run `./carthage.sh build --archive --platform iOS`
 - Upload to the new release PurchasesHybridCommon.framework.zip
 - Create a github release
-- Release Android:
```
beylmk, line 13: 👏

.circleci/config.yml
```diff
@@ -84,12 +84,12 @@ jobs:
...
             else
               echo "Deploying snapshot..."
-              ./gradlew androidSourcesJar androidJavadocsJar uploadArchives \
+              ./gradlew androidSourcesJar androidJavadocsJar publish closeAndReleaseRepository \
               -Psigning.keyId=$GPG_SIGNING_KEY_ID \
               -Psigning.password=$GPG_SIGNING_KEY_PW \
               -Psigning.secretKeyRingFile=./secring.gpg \
-              -PSONATYPE_NEXUS_USERNAME=$SONATYPE_NEXUS_USERNAME \
-              -PSONATYPE_NEXUS_PASSWORD=$SONATYPE_NEXUS_PASSWORD \
+              -PmavenCentralUsername=$SONATYPE_NEXUS_TOKEN_USERNAME \
```
beylmk, line 91: could we maybe get rid of the SONATYPE_NEXUS_USERNAME/PASSWORD variables now?

vegaro: I had to bring them back because of https://github.com/RevenueCat/purchases-hybrid-common/pull/104

I want to keep that branch with only amazon changes on top of 1.7.1

aboedo: @beylmk @vegaro FYI I ran into some issues with this, and updated it again in #105. I also added the new env variables, but I guess I should probably update the values we're reading here.

vegaro: I see so if I remember correctly, it would fail if it had the env SONATYPE_NEXUS_PASSWORD set even if it was not being accessed right? So I removed them, then that made my Amazon PR fail, so I added them, but that broke your PR lol. So the new version of the plugin doesn't have that issue right?

aboedo: lol
Yeah, or at least, I haven't run into the issue with the newer version... yet. But I didn't see anything in the changelog that would explain it

PR #98, opened 2021-11-19
Fix bump_and_update_changelog

vegaro:
`bump_and_update_changelog` was inside the ios platform, but it performs actions on both platforms. I also updated fastlane and fixed the attaching to changelog so it adds the version number


taquitos (approved):
🚤🐐

PR #100, opened 2021-11-19
Fix publish gradle tasks

vegaro:
The publish task failed and I went to look in purchases-android `main` branch and realized that there had been some changes after the PR that upgraded the publish plugin https://github.com/RevenueCat/purchases-android/blob/main/fastlane/Fastfile#L111

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,10 +121,6 @@ workflows:
             branches:
               ignore: /.*/
       - deploy-android:
-          requires:
-            - test-android
           filters:
-            tags:
-              only: /^(\d+\.)(\d+\.)(\d+)$/
             branches:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -84,7 +84,14 @@ jobs:
               echo "Skipping snapshot deployment: wrong JDK. Expected '$JDK' but was '$CIRCLE_JDK_VERSION'."
             else
               echo "Deploying snapshot..."
```
vegaro, line 86: We should move all this to fastlane, but that's a bigger task and I just want to ship this so we can enjoy this asap

taquitos (approved):
🏃‍♂️🐐

PR #101, opened 2021-11-22
Adds ownershipType to EntitlementInfo

vegaro:
Maps `ownershipType` in the `EntitlementInfo`

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/EntitlementInfoMapper.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/EntitlementInfoMapper.kt
@@ -20,5 +20,6 @@ fun EntitlementInfo.map(): Map<String, Any?> =
         "unsubscribeDetectedAt" to this.unsubscribeDetectedAt?.toIso8601(),
         "unsubscribeDetectedAtMillis" to this.unsubscribeDetectedAt?.toMillis(),
         "billingIssueDetectedAt" to this.billingIssueDetectedAt?.toIso8601(),
-        "billingIssueDetectedAtMillis" to this.billingIssueDetectedAt?.toMillis()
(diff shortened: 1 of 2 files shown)


taquitos (approved):
Owner🚢 'it

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/EntitlementInfoMapper.kt
```diff
@@ -20,5 +20,6 @@ fun EntitlementInfo.map(): Map<String, Any?> =
         "unsubscribeDetectedAt" to this.unsubscribeDetectedAt?.toIso8601(),
         "unsubscribeDetectedAtMillis" to this.unsubscribeDetectedAt?.toMillis(),
         "billingIssueDetectedAt" to this.billingIssueDetectedAt?.toIso8601(),
-        "billingIssueDetectedAtMillis" to this.billingIssueDetectedAt?.toMillis()
+        "billingIssueDetectedAtMillis" to this.billingIssueDetectedAt?.toMillis(),
+        "ownershipType" to "UNKNOWN"
```
aboedo, line 24: I don't think we should leave this as `UNKNOWN` for Android. 
I could still make purchases on iPhone that grant me access on Android if I'm using an accounts system, so I'd have a status `purchased`. 
And someone in my family could make a family purchase for a subscription, which grants access on my iPhone account... which I may also be using on Android. 

I don't think this is necessarily a far-fetched scenario, since a lot of people have Android phones but iPad tablets, or even macOS computers. 
I think we should read the value in `purchases-android`, and continue from there. 
Also, I think it's probably a matter of time until Google implements something akin to family sharing on iOS, and we designed the entire system so that it'd be platform-agnostic with that consideration in mind. 
@vegaro @taquitos

PR #103, opened 2021-11-29
Get the ownershipType from the android sdk

daentech:
Description:

This PR exposes `ownershipType` from `purchases-android` via the Hybrid Common Purchases SDK

I know that the play store does not provide this information, but it is possible that a user id exists in multiple platforms and as such may have their entitlement purchased by themselves, or shared to them, for example on iOS, and then be using it on Android.

This information is stored in the revenucat API and as such, the entitlement could be returned to Android with this data and a store type of e.g. `APP_STORE`

I'm making use of this in the purchases-flutter SDK, so a further PR will be forthcoming to update `purchases-flutter`

`purchases-android` PR: https://github.com/RevenueCat/purchases-android/pull/382

Related to Zendesk ticket: 12668

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,3 @@
-- Get the ownershipType from the purchases-android SDK via the API
-
 ## 1.11.0
 
 Add ownershipType to EntitlementInfo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Get the ownershipType from the purchases-android SDK via the API
+
 ## 1.11.0
 
 Add ownershipType to EntitlementInfo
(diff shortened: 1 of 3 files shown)


aboedo (commented):
thanks for opening this one as well!! 
We'll have to wait for the purchases-android PR so we can use this, since we have to update the version of `purchases-android` in `android/build.gradle` to use the property

daentech:
I can update that build.gradle ref once the purchaes-android has been updated, and then I can do the same for the flutter PR. Thanks!

aboedo (approved):

aboedo:
I guess tests here are set up not to run on forks. This should be good to go, though.

PR #104, opened 2021-12-02
Amazon support (Billing Client 3 and IAP 2)

vegaro:
Adds amazon support but on top of the 1.7.1 release, which is the last one supported in observer mode alongside Unity IAP since it's the last one that has the billing client 3


vegaro:
Closing since this has already been released as with the `-unityiap` artifacts

PR #106, opened 2022-01-14
fix inconsistency in `introPrice` mapping

aboedo:
mapping of null `introPrice` is inconsistent between iOS and Android: 

### iOS:
```json
"introPrice": null
```

[Code here](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/ios/PurchasesHybridCommon/PurchasesHybridCommon/SKProduct+HybridAdditions.m#L39)

### Android:
```json
"introPrice": { 
    "price": null,
    "priceString": null,
    "period": null,
    "cycles": null,
    "periodUnit": null,
    "periodNumberOfUnits": null
}
```

[Code here](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/android/src/main/java/com/revenuecat/purchases/common/mappers/SkuDetailsMapper.kt#L94)

This is the root cause of this [flutter issue](https://github.com/RevenueCat/purchases-flutter/issues/293)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.11.2
+
+- Fixes an inconsistency between `null` `introPrice` mapping in iOS and Android
+    https://github.com/RevenueCat/purchases-hybrid-common/pull/106
+
 ## 1.11.1
 
 - Adds compatibility for `ownershipType` for Android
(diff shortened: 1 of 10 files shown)


aboedo:
this also seems more consistent with our [typescript declarations in react-native-purchases](https://github.com/RevenueCat/react-native-purchases/blob/main/src/offerings.ts#L130), which consider `introPrice` as optionals, but not its fields.

aboedo:
for cordova, we're using the deprecated fields 🤦 but those shouldn't be an issue and are unaffected by this PR

joshdholtz (approved):
This looks good to me! 💪 🚀

PR #107, opened 2022-01-26
Release 2.0.0

vegaro:
### Breaking changes
- Updated `configureWithAPIKey:` method to accept a DangerousSettings object

### Other
- Bump purchases-ios to 3.14.0 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/3.14.0)


ios/PurchasesHybridCommon/PurchasesHybridCommon/RCPurchases+HybridAdditions.h
```diff
@@ -18,14 +18,16 @@ NS_ASSUME_NONNULL_BEGIN
                        observerMode:(BOOL)observerMode
               userDefaultsSuiteName:(nullable NSString *)userDefaultsSuiteName
                      platformFlavor:(nullable NSString *)platformFlavor
-              platformFlavorVersion:(nullable NSString *)platformFlavorVersion;
+              platformFlavorVersion:(nullable NSString *)platformFlavorVersion
+                  dangerousSettings:(nullable RCDangerousSettings *)dangerousSettings;
```
aboedo, line 22: this would technically make it a major release

vegaro: That makes sense!

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.12.0
+
+- Bump purchases-ios to 3.14.0 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/3.14.0)
```
aboedo, line 3: I think the change in the init should be documented here as a breaking change. 
And the version number should be 2.0.0 if we're following semver. I think that's fine, this is an internal library, so there's really no problem in it being major. But we should still stick to semver. 

It might be confusing otherwise if you're making a release of something like `cordova-plugin-purchases`, you bump to the latest hybrid-common version as per the RELEASING.md instructions, then it stops working and you don't understand why

aboedo (changes requested):

CHANGELOG.md
vegaro: Alternatively I can do a 3.14.1 fixing purchases-ios and have this be a minor. What do you think?

vegaro: Yes. It feels scary to do a major, but yes, it's not a big deal in this case. I of course had to break something with that iOS release lol

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.0.0
+
+- Updated init function to accept a DangerousSettings object
+- Bump purchases-ios to 3.14.0 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/3.14.0)
```
aboedo, line 4:
```suggestion
### Breaking changes
- Updated `configureWithAPIKey:` method to accept a DangerousSettings object

### Other
- Bump purchases-ios to 3.14.0 [Changelog here](https://github.com/RevenueCat/purchases-ios/releases/tag/3.14.0)
```

aboedo (approved):

CHANGELOG.md
vegaro: Thanks!

aboedo (approved):

PR #112, opened 2022-02-15
Updates Android versions

vegaro:
- Removed jcenter
- Updated to Gradle 7.0.2
- Upgraded JUnit5 because the one we were using is not available in maven central
- Upgraded build tools version because the Android plugin 7.0.2 uses that version anyway

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    ext.android_tools = '7.1.1'
+    ext.android_tools = '7.0.2'
     ext.kotlin_version = '1.4.10'
     ext.android_junit5_version = '1.8.2.0'
     ext.junit5 = '5.8.2'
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -92,7 +84,9 @@ jobs:
           command: bundle exec fastlane ios release
 
   test-android:
-    executor: android-executor
+    executor:
+      name: android/android-machine
+      resource-class: large
```
taquitos, line 89: 💪

android/build.gradle
```diff
@@ -13,7 +13,6 @@ buildscript {
     repositories {
         google()
         mavenCentral()
-        jcenter()
```
taquitos, line 16: ❤️

taquitos (approved):
Awesome.

vegaro:
@taquitos I've realized there's actually a 7.1.1, so I just updated that and gradle

taquitos:
> @taquitos I've realized there's actually a 7.1.1, so I just updated that and gradle

Oh, nice!

vegaro:
Ugh reverting that change because it requires to increase the target version and that requires some testing 👎

aboedo (approved):

PR #113, opened 2022-02-16
[WIP] iOS v4 migration

joshdholtz:
An early review on this would be appreciated 🙌  

## Maybe iOS SDK Tasks
- [ ] Add number of periods/cycles to `StoreProductDiscount` (if we still need that)
- [ ] Add way to get currency on `StoreProduct` and `StoreProductDiscount`

- [ ] Which StoreProduct fields to add…
    * Should we be normalizing the product type/category across android and ios?
- [ ] Pass through localizedprice string directly from purchases-ios instead of re-formatting it in purchases-hybrid-common (we are currently formatting it twice for ios)
- [ ] promotional offer flow

PR #115, opened 2022-02-28
Update to Android v5

beylmk:
Android equivalent of https://github.com/RevenueCat/purchases-hybrid-common/pull/113

This was branched off `update-to-rc2` and then main was merged in... 

Adds:
* setLogHandler
* dangerousSettings option in configuration

Also:
* pulls fixes from https://github.com/RevenueCat/purchases-hybrid-common/pull/106 into the renamed StoreProductMappers file
* defaults the purchases-android dependency to the `latestDependencies` flavor... not sure if we want to make flavors for purchases-hybrid-common, but figured that would be in a separate PR if so


.circleci/config.yml
```diff
@@ -156,3 +156,40 @@ workflows:
...
+            branches:
+              ignore: /.*/
+      - deploy-android:
+          requires:
+            - test-android
+          filters:
+            tags:
+              only: /^(\d+\.)(\d+\.)(\d+)$/
+            branches:
+              ignore: /.*/
```
aboedo, line 175: how does this differ from the `deploy-android` job?

.circleci/config.yml
```diff
@@ -156,3 +156,40 @@ workflows:
...
+  android-deploy-amazon:
+    jobs:
+      - test-android:
+          filters:
+            branches:
+              only: amazon
+      - hold:
+          type: approval
+          requires:
+            - test-android
```
aboedo, line 186: why does this one require approval vs the others?

aboedo, line 186: why does this one require approval vs the others?

android/build.gradle
```diff
@@ -39,6 +39,7 @@ android {
         versionName "2.0.1"
 
         consumerProguardFiles 'consumer-rules.pro'
+        missingDimensionStrategy 'dependencyVersions', 'latestDependencies', 'unityIAP'
```
aboedo, line 42: Could we add a comment here specifying what it does? 

I figure that this will try to choose the latest dependencies first, but if there's an incompatibility with other requirements for the same app, it'll try the flavor `unityIAP`, is that correct?

aboedo, line 42: Could we add a comment here specifying what it does? 

My understanding is that this will try to choose the latest dependencies first, but if there's an incompatibility with other requirements for the same app, it'll try the flavor `unityIAP`, is that correct?

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -0,0 +1,73 @@
+package com.revenuecat.purchases.hybridcommon.mappers
+
+import com.revenuecat.purchases.models.StoreProduct
+
+fun StoreProduct.map(): Map<String, Any?> =
+    mapOf(
+        "identifier" to sku,
+        "description" to description,
+        "title" to title,
+        "price" to priceAmountMicros / 1000000.0,
```
aboedo, line 10: I think this was here before but could we have this be `1_000_000.0` for readability?

aboedo, line 10: I think this was here before but could we have this be `1_000_000.0` for readability?

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -0,0 +1,73 @@
...
+                    "priceString" to formatUsingDeviceLocale(priceCurrencyCode, 0),
+                    "period" to freeTrialPeriod,
+                    "cycles" to 1
+                ) + periodFields
+            } ?: null
+        }
+        introductoryPrice != null -> {
+            introductoryPricePeriod!!.mapPeriod()?.let { periodFields ->
+                mapOf(
+                    "price" to introductoryPriceAmountMicros / 1000000.0,
```
aboedo, line 37: same here

aboedo, line 37: same here

.circleci/config.yml
vegaro: we don't need these, removing. I think they are leftovers from manually triggered releases we've been doing in the past year

android/build.gradle
vegaro: I am not sure what this is for either @beylmk . Can you add some context?

.circleci/config.yml
vegaro: Same as the other comment. We were triggering releases using the approval step, I don't remember why. I removed it because it's not needed

android/build.gradle
beylmk: @aboedo yes, that's correct. we have to specify the flavor of purchases-android that we want to depend on, and this defaults to latestDependencies. 

we could instead make purchases-hybrid-common flavors, which in turn pull a specific purchases-android flavor. i figured that could be a follow-up ticket

beylmk: see points 2/3 in the TLDR at this link: https://kiranrao.in/blog/2020/03/31/gradle-missing-flavors/

according to that, we probably do want to make PHC flavors

vegaro: I am actually working on adding the flavors so we can improve Unity by A LOT

vegaro: I think maybe this was required because you were doing local development and pointing to your copy of purchases-android 🤔

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -194,9 +194,9 @@ fun logIn(
 ) {
```
aboedo, line 194: we should also rename `restoreTransactions` -> `restorePurchases`, but that could be a separate PR

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -322,14 +288,21 @@ fun configure(
...
-        Purchases.configure(context, apiKey, appUserID, observerMode)
-    } else {
-        Purchases.configure(context, apiKey, appUserID)
+        builder.observerMode(observerMode)
     }
+
+    Purchases.configure(builder.build())
 }
 
 fun getPaymentDiscount() : ErrorContainer {
```
aboedo, line 308: this should be renamed to getPromotionalOffer, but it can also be done as a separate PR

aboedo (approved):
Looks great! I left a couple of notes w/regards to other APIs that need a rename, but we can do that as a separate thing. 
Have you already tested this on a hybrid?

.circleci/config.yml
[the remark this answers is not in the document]
aboedo: lol I guess I had that comment for weeks and hadn't posted the PR 🤦‍♂️

vegaro:
will open follow up PRs. I have tested this while working on updating Flutter and so far it's compiling. Some things to change on the Flutter side though 😄

PR #116, opened 2022-04-07
[On top of amazon-unity-iap-compatible] Increase version to make Amazon observer mode fix adding price and currency

vegaro:
Increase version number for when https://github.com/RevenueCat/purchases-android/pull/519/ is released, so we can release this and point unity 3.2.0-amazon.alpha.6 to this version


beylmk (approved):
👍

PR #119, opened 2022-04-13
Migrate to swift part 1: PurchasesHybridCommonSwift

aboedo:
This is the first step in the migration to Swift for HybridCommon. 
It adds a new project and pod, that will hold the Swift files. 
Just like with `Purchases` and `PurchasesCoreSwift`, the idea is that `PurchasesHybridCommonSwift` is considered lower-level and used by `PurchasesHybridCommon` under the hood.


ios/PurchasesHybridCommon/PurchasesHybridCommonSwift/PurchasesHybridCommonSwift.docc/PurchasesHybridCommonSwift.md
```diff
@@ -0,0 +1,13 @@
+# ``PurchasesHybridCommonSwift``
```
aboedo, line 1: this gets killed in another PR

ios/PurchasesHybridCommon/PurchasesHybridCommonSwiftTests/PurchasesHybridCommonSwiftTests.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  PurchasesHybridCommonSwiftTests.swift
+//  PurchasesHybridCommonSwiftTests
+//
+//  Created by Andrés Boedo on 4/13/22.
+//  Copyright © 2022 RevenueCat. All rights reserved.
+//
+
+import XCTest
```
aboedo, line 9: I didn't add new tests here yet, we might just stick to the ones in PurchasesHybridCommon

vegaro (approved):
I believe tests are failing because this line from the second PR is missing, so that's fine https://github.com/RevenueCat/purchases-hybrid-common/pull/120/files#diff-11764a582a962cb9b10bf9f90c548ee35dd14b260b29bb7044ac1d45def4a4caR10

aboedo:
Two updates here: 

1. I'm changing the base to `swift_migration` so we never leave `main` in an intermediate state.2. 
2. Tests are actually not gonna work entirely in CI because `pod lib lint` needs the `PurchasesHybridCommonSwift` pod to be published. I'm going to ignore this for now, then remove the CI check in a separate PR, and re-enable the test as soon as we finish the migration. However, if this gets prolonged I'll add a private pod so that we can continue to test with `pod lib lint`.

PR #120, opened 2022-04-13
Migrate to swift part 2: API Tester

aboedo:
Creates an API Tester that we can use to ensure we don't make unintended changes to the API in `PurchasesHybridCommon` during the migration.


ios/PurchasesHybridCommon/Podfile
```diff
@@ -1,12 +1,13 @@
 # Uncomment the next line to define a global platform for your project
 
 target 'PurchasesHybridCommon' do
-  platform :ios, '9.0'
+  platform :ios, '11.0'
```
aboedo, line 4: took the opportunity to update the deployment targets. This simplifies code in a few places where we were checking for older OS versions. 11.0 matches purchases-ios v4

ios/PurchasesHybridCommon/Podfile
```diff
@@ -15,3 +16,31 @@ target 'PurchasesHybridCommon' do
   end
 
 end
+
```
aboedo, line 19: a lot of boilerplate in this one, but it can be summed up as: 
- `PurchasesHybridCommon` uses `PurchasesHybridCommonSwift`
- `ObjCApiTester` uses `PurchasesHybridCommon` and `PurchasesHybridCommonSwift`
- They use local paths because we're not going to actually publish `PurchasesHybridCommonSwift` as a pod this time around, since it won't live long enough for that

beylmk:
will we eventually have this build in ci?

vegaro (approved):

aboedo:
@beylmk 

> will we eventually have this build in ci?

Yes, [see my comment in the initial PR](https://github.com/RevenueCat/purchases-hybrid-common/pull/119#issuecomment-1103065104)

PR #130, opened 2022-04-22
remove @testable imports

aboedo:
This updates the codebase so that we no longer need `@testable` imports. 
Before shipping this we will need to make a purchases-ios release [including the changes from this PR](https://github.com/RevenueCat/purchases-ios/pull/1523). 

Note that `pod lib lint` will continue to fail on this PR, sadly, because we can't point a `Podspec` to a specific branch, only a `Podfile`. And `pod lib lint` uses a `Podspec`.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-@testable import RevenueCat
+import RevenueCat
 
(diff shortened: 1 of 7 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
```diff
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-@testable import RevenueCat
+import RevenueCat
```
aboedo, line 10: this one wasn't needed at all

PR #134, opened 2022-04-29
Adds @JvmOverloads to configure function

vegaro:
Adds @JvmOverloads to configure function since most hybrids are still in Java


beylmk (approved):
🌄

vegaro:
Yes, I think the plan is to actually remove a lot of functions that we don't need from purchases-hybrid-common. We can call Purchases directly from the hybrids. A good example is this function or `invalidateCustomerInfoCache` which doesn't really do anything the hybrids can't do directly.

PR #136, opened 2022-05-04
Update cocoapods downloader and fastlane

vegaro:
We have a dependabot alert regarding cocoapods-downloader having a vulnerability


taquitos (approved):
👍 🐐

PR #141, opened 2022-05-20
Fix: auto release iOS

aboedo:
Auto release for iOS failed for a few reasons: 
- the API tester was using Cocoapods to get the actual `PurchasesHybridCommon.framework`, which was a leftover from a complicated intermediate setup during the swift migration. I updated it so that it uses the project from xcode instead. This also means that you don't need to call `pod install` when modifying the API in order to test the API tester, which was annoying. 
- with the API tester fixed, carthage would still complain because it was trying to build multiple slices and the whole fun apple silicon incompatibility thing, which annoyed me so I killed Carthage with fire. I replaced it with `fastlane gym`.


carthage.sh
```diff
@@ -1,23 +0,0 @@
-# This script is needed to use Carthage in Xcode 12, since it currently has a compatibility issue
```
aboedo, line 1: die, hacky workaround, die

fastlane/Fastfile
```diff
@@ -166,18 +166,33 @@ def push_pods
...
+  build_app(
+    workspace: "ios/PurchasesHybridCommon/PurchasesHybridCommon.xcworkspace",
+    configuration: "Release",
+    scheme: "PurchasesHybridCommon",
+    silent: true,
+    clean: true,
+    derived_data_path: "derived_data/",
+    output_name: framework_filename,
+    destination: 'generic/platform=iOS'
+  )
```
aboedo, line 180: this is also faster, since it's only building the actual framework we want, in the actual platform we want.

fastlane/Fastfile
```diff
@@ -166,18 +166,33 @@ def push_pods
...
+
+  # fastlane gym doesn't fully support .frameworks, so we need to look for the output
+  # in derived data manually.
```
aboedo, line 183: this was the reason we weren't using `fastlane gym` before, but I'd much rather do this than deal with carthage's issues

fastlane/Fastfile
```diff
@@ -166,18 +166,33 @@ def push_pods
...
-  end
+    output_framework_path = Dir["derived_data/**/#{framework_filename}"].first
+    UI.error("framework not found!") if output_framework_path.nil?
 
-  def is_snapshot_version?(version_name)
-    version_name.end_with?("-SNAPSHOT")
+    # compress using ditto because zip can cause issues with symlinks
+    sh("ditto", "-c", "-k", "--sequesterRsrc", "--keepParent", 
+      output_framework_path, 
+      "#{framework_filename}.zip")
```
aboedo, line 191: This was a big issue for purchases-ios with catalyst at one point, so I'm doing it just in case. 
`ditto` seems to be what macOS's Finder uses under the hood when you right-click and select "Compress" and produces identical output

fastlane/Fastfile
```diff
@@ -166,18 +166,33 @@ def push_pods
...
+  framework_filename = "PurchasesHybridCommon.framework"
+  build_app(
+    workspace: "ios/PurchasesHybridCommon/PurchasesHybridCommon.xcworkspace",
+    configuration: "Release",
+    scheme: "PurchasesHybridCommon",
+    silent: true,
+    clean: true,
+    derived_data_path: "derived_data/",
+    output_name: framework_filename,
+    destination: 'generic/platform=iOS'
```
NachoSoto, line 179: 👍🏻

NachoSoto (approved):
Seems reasonable 👍🏻

PR #142, opened 2022-05-24
fix: add amazon the EntitlementInfo dictionary's store value

fire-at-will:
This PR adds `Amazon` to the `Store` dictionaries in both the iOS and Android common hybrid SDKs.

### Motivation
Resolves [CF-653](https://revenuecats.atlassian.net/browse/CF-653).

#### iOS
- Added the `RCAmazon` value to the `RCEntitlementInfo.jsonDict`'s `store` value. 
- Update `purchases-ios` version to `4.4.0`
- Adopts the new `Configuration` builder pattern to avoid warnings during the `pod lib lint` CI phase.

#### Android
No changes are required to the main code since:
1. The [EntitlementInfoMapper automatically consumes](https://github.com/RevenueCat/purchases-hybrid-common/blob/ed689b6ad230a2e2c94faf87f5bf9eb7194c1f1e/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/EntitlementInfoMapper.kt#L17) changes to the Android SDK's `Store` enum.
2. The Android SDK [already has](https://github.com/RevenueCat/purchases-android/blob/e40aac4487aaf385ec26eeb843b0bc73a6b61140/public/src/main/java/com/revenuecat/purchases/EntitlementInfo.kt#L150) the `Amazon` `Store` case.

#### iOS
- Adds tests for the all possible `store` values of the `EntitlementInfo+HybridAdditions`'s `dictionary` property.
- Fixes an unrelated broken function call in the `MockPurchases` test file.
- Fix CustomerInfo tests broken by upstream `purchases-ios` changes (specifically, [this commit](https://github.com/RevenueCat/purchases-ios/commit/29c034dfc80690c21323c30880d65027dd97223a))

#### Android
- Adds tests for all possible `store` values of the `EntitlementInfoMapper`'s `map` function.


ios/PurchasesHybridCommon/PurchasesHybridCommonTests/EntitlementInfo+HybridAdditionsTests.swift
```diff
@@ -0,0 +1,166 @@
...
+                        return
+                    }
+
+                    let dictionary = mockEntitlementInfo.dictionary
+                    expect(dictionary["store"] as? String).to(
+                        equal(expectedDictionaryValue),
+                        description: "Expected \(String(describing: dictionary["store"] as? String)) to become \(expectedDictionaryValue)."
+                    )
+                }
+                it("equals PLAY STORE when the store is play store") {
```
NachoSoto, line 63: All these tests are essentially the same with different inputs / outputs. Especially with Quick you could extract the core of the test then instantiate multiple tests (basically calls to `it`) inside of a loop to avoid the duplicate code.

fire-at-will: Good catch! I've updated the iOS & Android tests to be more compact.

fire-at-will:
@NachoSoto I've fine-tuned those unit tests, updated the `purchases-ios` dependency to `4.4.0`, and updated the `CustomerInfo` unit tests! Let me know what you think 👀

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/CustomerInfo+HybridAdditionsTests.swift
```diff
@@ -10,37 +10,54 @@ import Quick
...
-                    let mockCustomerInfo = try CustomerInfo(data: [
-                        "request_date": "2019-08-16T10:30:42Z",
-                        "subscriber": [
-                            "first_seen": "2019-07-17T00:05:54Z",
-                            "management_url": "https://revenuecat.com",
-                            "original_app_user_id": "",
-                            "subscriptions": [:],
-                            "other_purchases": [:]
-                        ]])
+                    guard let mockCustomerInfo = CustomerInfo.fromJSON(
```
NachoSoto, line 21: I think `it` accepts throwing, so you could keep this as `throws`, and then the `try` will automatically fail the test.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/CustomerInfo+TestExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+import RevenueCat
+
+extension CustomerInfo {
+    static func fromJSON(_ jsonString: String) -> CustomerInfo? {
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        decoder.dateDecodingStrategy = .iso8601
+
+        guard let data = jsonString.data(using: .utf8) else { return nil }
```
NachoSoto, line 18: You can do `Data(jsonString.utf8)` and that won't return an optional.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/EntitlementInfo+HybridAdditionsTests.swift
```diff
@@ -0,0 +1,85 @@
...
+                            description: "Expected \(String(describing: dictionary["store"] as? String)) to become \(expectedDictionaryValue)."
+                        )
+                    }
+                }
+            }
+        }
+    }
+
+    private func mockEntitlementInfo(store: Store) -> EntitlementInfo? {
+        return EntitlementInfo(
```
NachoSoto, line 58: Does this return an optional?

fire-at-will: It doesn't! I think it did before the `4.4.0` versions, I've removed the optional now :)

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/CustomerInfo+TestExtensions.swift
fire-at-will: Oh, nice! I've used that initializer and removed the `guard let`s where this helper is being called.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/CustomerInfo+TestExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+import RevenueCat
+
+extension CustomerInfo {
+    static func fromJSON(_ jsonString: String) -> CustomerInfo {
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        decoder.dateDecodingStrategy = .iso8601
+
+        let data = Data(jsonString.utf8)
+        return try! decoder.decode(CustomerInfo.self, from: data)
```
NachoSoto, line 19: :warning: This will make the tests _crash_ instead of _fail_, which means none of the other tests will run and execution stops, instead of reporting each individual failure.
Instead, make this function `throws`, and use `try`. At the call site, you can just do `try CustomerInfo.fromJSON("")` inside of an `it`, and if it throws, the test fails instead.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/EntitlementInfo+HybridAdditionsTests.swift
```diff
@@ -0,0 +1,81 @@
...
+                    ]
+
+                    for expectation in expectations {
+                        let (store, expectedDictionaryValue) = expectation
+                        let mockEntitlementInfo = self.mockEntitlementInfo(store: store)
+
+                        let dictionary = mockEntitlementInfo.dictionary
+                        expect(dictionary["store"] as? String).to(
+                            equal(expectedDictionaryValue),
+                            description: "Expected \(String(describing: dictionary["store"] as? String)) to become \(expectedDictionaryValue)."
```
NachoSoto, line 45: Nit: you don't need to reference `dictionary["store"]` because that will be part of the assertion message already (since that's the value you're checking. I would just do:
```suggestion
                            description: "Expected \(store) to be encoded as \(expectedDictionaryValue)."
```

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
```diff
@@ -10,17 +10,23 @@ import Quick
...
 
-    private let mockCustomerInfo = try! CustomerInfo(data: [
-        "request_date": "2019-08-16T10:30:42Z",
-        "subscriber": [
-            "first_seen": "2019-07-17T00:05:54Z",
-            "original_app_user_id": "",
-            "subscriptions": [:],
-            "other_purchases": [:]
-        ]])
+    private var mockCustomerInfo = CustomerInfo.fromJSON(
```
NachoSoto, line 17: This doesn't need to be mutable right?
```suggestion
    private static let mockCustomerInfo = try! CustomerInfo.fromJSON(
```

Here you can do `try!` since this belongs to all tests, and gets created at initialization.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/CustomerInfo+HybridAdditionsTests.swift
```diff
@@ -10,37 +10,48 @@ import Quick
...
                 it ("contains null when the management url doesn't exist") {
-                    let mockCustomerInfo = try CustomerInfo(data: [
-                        "request_date": "2019-08-16T10:30:42Z",
-                        "subscriber": [
-                            "first_seen": "2019-07-17T00:05:54Z",
-                            "original_app_user_id": "",
-                            "subscriptions": [:],
-                            "other_purchases": [:]
-                        ]])
+                    let mockCustomerInfo = try! CustomerInfo.fromJSON(
```
NachoSoto, line 42: This would still crash. I think `try` should work?

taquitos:
YAY!

fire-at-will:
Thanks for all of the feedback, @NachoSoto! I really like the improvements that this PR is bringing to the SDK and its tests :)

NachoSoto (approved):
Awesome!

PR #143, opened 2022-06-01
Version 3.0.1

vegaro:
- Added Amazon case to the store value in the EntitlementInfo map in both the iOS and Android https://github.com/RevenueCat/purchases-hybrid-common/pull/142/
- Bumped `purchases-ios` to `4.4.0` https://github.com/RevenueCat/purchases-hybrid-common/pull/142
- Bumped `purchases-ios` to `4.4.1`


aboedo:
adding an enum case would make this a `3.1.0` release, right?

vegaro:
That's what I thought at first, but we are not adding an enum case, we are adding a new value possibility to a map, not to an enum https://github.com/RevenueCat/purchases-hybrid-common/pull/142/files#diff-a22b288a94eab820bdfc8a03ce19b11f43906ea4ceed783e44c032e3c13f8aa0R80

So I think it's a patch because it's really not changing the API, just one of the values in the map, but I can be convinced of the opposite.

PR #146, opened 2022-06-15
Fix call to purchase(product in iOS

vegaro:
With the refactor to Swift, we broke purchaseProduct and the purchase was never initiated. 

I believe we would need to setup StoreKit test in order to be able to test this.


aboedo (approved):
LOL I guess that's what you meant by "found a big bug". I'm glad you caught it before it shipped

aboedo:
the good news is that setting up StoreKit tests for hybrid common is scheduled for next sprint 🎉

PR #149, opened 2022-06-16
Fix deploys

vegaro:


fastlane/Fastfile
```diff
@@ -40,38 +74,10 @@ platform :ios do
   desc "Release to CocoaPods, create Carthage archive, and create GitHub release"
   lane :release do |options|
     version_number = current_version_number
-    push_pods
+    # push_pods
```
taquitos, line 77: I'm guessing this is an accident 😄

ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
```diff
@@ -563,10 +563,12 @@
 			baseConfigurationReference = 460EB988CBB68111D4CF19A2 /* Pods-PurchasesHybridCommon.debug.xcconfig */;
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_IDENTITY = "Apple Development";
+				"CODE_SIGN_IDENTITY[sdk=macosx*]" = "Apple Development";
+				CODE_SIGN_STYLE = Manual;
```
taquitos, line 568: ᕙ(⇀‸↼‶)ᕗ

taquitos (changes requested):

ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
aboedo: I think it has to be manual, right?

aboedo: we should be able to use the other projects as a reference

.circleci/config.yml
```diff
@@ -89,9 +90,17 @@ jobs:
...
+          name: Install pods
+          command: pod install
+          working_directory: ios/PurchasesHybridCommon
       - run:
           name: Deploy new version
           command: bundle exec fastlane ios release
+      - persist_to_workspace:
+          root: .
+          paths:
+            - PurchasesHybridCommon.framework.zip
```
aboedo, line 103: do we need this? it'd already be added to the workspace through the lane, right?

aboedo (approved):

.circleci/config.yml
vegaro: This is needed for `make-github-release` which is another job that needs that file. Afaik, just because the lane adds it to the filesystem it doesn't mean it is accessible in a downstream job. Take a look at https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/

taquitos: Yeah, this is a problem I had- the file wasn't persisted. Makes sense.

fastlane/Fastfile
```diff
@@ -25,6 +25,40 @@ lane :bump_and_update_changelog do |options|
   increment_version_in(previous_version_number, new_version_number, "../.version")
 end
 
+desc "Make GitHub release for current version"
+lane :github_release_current do |options|
+  version_number = current_version_number
+  github_release(version: version_number)
+end
+
+desc "Make GitHub release"
```
taquitos, line 34: Maybe `Make GitHub release with specific version`

taquitos (approved):
Desc was a little confusing for me, I know it's been there a while, but I had to read the code to understand the difference between the two release methods.

PR #154, opened 2022-06-23
Removed `FatalErrorUtil` in favor of `purchases-ios` version

NachoSoto:


ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -8,8 +8,8 @@
 
 import Foundation
 import StoreKit
-import RevenueCat
 
+@testable import RevenueCat
```
aboedo, line 12: Wouldn't this cause compilation issues on `RELEASE` builds, where `RevenueCat` wouldn't be built for testing?

NachoSoto: Oh that's right this is not only in tests 🤦‍♂️

PR #157, opened 2022-06-28
Started Integration Tests

NachoSoto:
For [CSDK-146].

### TODO:

- [x] Change `Podfile` to point to `4.6.1` and not my path.
- [x] Set up CI
- [x] Avoid duplicating `.storekit` file from main repo?
- [x] Fix test run on CI
- [x] Make new CircleCI job _required_?
- [x] Add more tests: #158 

[CSDK-146]: https://revenuecats.atlassian.net/browse/CSDK-146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #158, opened 2022-06-28
Added Integration Tests

NachoSoto:
Follow up to #157.
Fixes [CSDK-146].

### TODO:

- [x] Look into flaky login/logout test
- [x] Figure out incorrect SK1/SK2 `product_type`:
```diff
26c26
<           "product_type" : "NON_CONSUMABLE",
---
>           "product_type" : "AUTO_RENEWABLE_SUBSCRIPTION",
53c53
```
_This is the reason: https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/StoreKitAbstractions/SK1StoreProduct.swift#L36-L38_
- [x] Handle changing dates

[CSDK-146]: https://revenuecats.atlassian.net/browse/CSDK-146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/CommonFunctionality+async.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/CommonFunctionality+async.swift
@@ -6,8 +6,7 @@
 //  Copyright © 2022 RevenueCat. All rights reserved.
 //
 
-import PurchasesHybridCommon
-
+@testable import PurchasesHybridCommon
 @testable import RevenueCat
 
 extension CommonFunctionality {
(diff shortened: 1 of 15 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -49,10 +54,48 @@ class StoreKit1IntegrationTests: BaseIntegrationTests {
...
+
+        do {
+            try await self.purchaseMonthlyOffering()
+            fail("Expected error")
+        } catch {
+            expect(error).to(matchError(ErrorCode.invalidPromotionalOfferError))
+        }
+    }
+
+    func testLogInAndLogOut() async throws {
```
NachoSoto, line 82: This test is flaky and I have no idea why :/ I think it may be exposing a bug.

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/CommonFunctionality+async.swift
```diff
@@ -20,4 +19,57 @@ extension CommonFunctionality {
         }
     }
 
+    static func purchase(product productIdentifier: String,
```
aboedo, line 22: we could have these live as part of the public API, but as of right now we wouldn't really be able to use them because they're only compatible with iOS 13. So having them here makes sense 👍

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -49,10 +54,51 @@ class StoreKit1IntegrationTests: BaseIntegrationTests {
...
+        // Fetch receipt to make sure there aren't race conditions
```
aboedo, line 83: let's add another line here like
```swift
// since the receipt might not be available on device on the first run, 
// Fetch receipt to make sure there aren't race conditions dependent on the order of test runs
```

aboedo (approved):
:chefskiss:

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: Thanks! That's much clearer

PR #160, opened 2022-06-29
add comment to clarify `canMakePayments` behavior on iOS

aboedo:
Added comment to clarify behavior of `canMakePayments` on iOS. 

On iOS, we ignore the `features` value passed in, because those are for Google Play only. 
But there's no reason a maintainer should know that when reading the code, so this adds a comment to explain what's going on.


aboedo:
shoutout to @taquitos for bringing this up!

NachoSoto (approved):

PR #161, opened 2022-06-29
Add api tests for android code

tonidero:
https://revenuecats.atlassian.net/browse/CSDK-140

As mentioned in the ticket, I added the API tests as part of the test suite of the module. While they are not unit tests, they will be compiled so we will catch any issues when running tests normally, so android api tests don't require any extra steps to run, just the usual `./gradlew lint test`.

### DIFF
--- a/android/src/test/java/apitests/java/CommonApiTests.java
+++ b/android/src/test/java/apitests/java/CommonApiTests.java
@@ -1,0 +1,149 @@
+package apitests.java;
+
+import android.app.Activity;
+import android.content.Context;
+
+import com.revenuecat.purchases.DangerousSettings;
+import com.revenuecat.purchases.LogHandler;
+import com.revenuecat.purchases.Store;
+import com.revenuecat.purchases.common.PlatformInfo;
+import com.revenuecat.purchases.hybridcommon.CommonKt;
+import com.revenuecat.purchases.hybridcommon.ErrorContainer;
+import com.revenuecat.purchases.hybridcommon.OnResult;
+import com.revenuecat.purchases.hybridcommon.OnResultAny;
+import com.revenuecat.purchases.hybridcommon.OnResultList;
+
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings({"unused", "deprecation"})
+class CommonApiTests {
+    private void checkCheckSetAllowSharingAppStoreAccount(Boolean enabled) {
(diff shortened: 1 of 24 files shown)


android/src/test/java/apitests/java/mappers/CustomerInfoMapperApiTests.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.revenuecat.purchases.CustomerInfo;
+import com.revenuecat.purchases.hybridcommon.mappers.CustomerInfoMapperKt;
+
+import java.util.Map;
+
+@SuppressWarnings("unused")
+class CustomerInfoMapperApiTests {
+    private void checkMap(CustomerInfo customerInfo) {
+        Map<String, Object> map = CustomerInfoMapperKt.map(customerInfo);
```
tonidero, line 11: Honestly, for all these I thought about joining the api tests into a single file... But I like better that the tests follow the original file structure of the project, even if these are not unit tests but api tests.

tonidero (commented):
Note that `purchases-hybrid-common` also makes a bunch of public classes in `purchases-android` available, but I didn't think it made sense to api-test them here since they are already api-tested there.

android/src/test/java/apitests/java/CommonApiTests.java
```diff
@@ -0,0 +1,149 @@
...
+import com.revenuecat.purchases.hybridcommon.OnResult;
+import com.revenuecat.purchases.hybridcommon.OnResultAny;
+import com.revenuecat.purchases.hybridcommon.OnResultList;
+
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings({"unused", "deprecation"})
+class CommonApiTests {
+    private void checkCheckSetAllowSharingAppStoreAccount(Boolean enabled) {
```
NachoSoto, line 21: Why not `boolean`?

NachoSoto (approved):
Awesome!

android/src/test/java/apitests/java/CommonApiTests.java
tonidero: Well, the compiler won't really help in this case, since the conversion from Kotlin's `Boolean` doesn't make the java side not nullable. But it's true that strictly speaking, this can be `boolean` since it shouldn't ever be null, so I will make the change everywhere except where we have nullable booleans (1 place only). Thanks!

PR #163, opened 2022-06-30
`Podspec`: updated `RevenueCat` dependency to 4.6.1

NachoSoto:
Missed that this was repeated there when I did #155.

I can't think of a way to enforce that these remain up-to-date though.

### DIFF
--- a/PurchasesHybridCommon.podspec
+++ b/PurchasesHybridCommon.podspec
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
 
   s.framework      = 'StoreKit'
 
-  s.dependency 'RevenueCat', '4.4.1'
+  s.dependency 'RevenueCat', '4.6.1'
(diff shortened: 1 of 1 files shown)


aboedo:
> I can't think of a way to enforce that these remain up-to-date though.

We should probably use fastlane scripts to update the version numbers together, i.e.: create a lane `bump_hybrid_common version:3.2.0`

aboedo (approved):

aboedo:
we list out the relevant files in [RELEASING.md](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/RELEASING.md) but it's not exactly bullet proof

NachoSoto:
Oh it is in there 👍🏻 

I didn't see it because I wasn't making a new release, but at least that ensures that when we do, those versions are consistent.

PR #178, opened 2022-07-22
Creates lane to update native dependencies to latest versions

vegaro:
[CSDK-346](https://revenuecats.atlassian.net/browse/CSDK-346)

Creates a lane that will:

- Get latest releases of purchases-android and purchases-ios in GitHub
- Updates `PurchasesHybridCommon.podspec`, `ios/PurchasesHybridCommon/Podfile`, `ios/PurchasesHybridCommon/Podfile.lock`, `DEVELOPMENT.md`, `android/build.gradle`

The plan is to schedule this to run every wednesday and create an automatic PR with the changes

Some of this code is already present in other Fastfiles, so it should be able to be simplified with @tonidero work on our own Fastlane plugin


fastlane/actions/get_latest_github_release.rb
```diff
@@ -0,0 +1,170 @@
+module Fastlane
```
vegaro, line 1: We could publish this action. It's based on `get_github_release`, but that one requires passing a tag name,

fastlane/Fastfile
```diff
@@ -59,6 +59,98 @@ lane :github_release do |options|
...
+  update_android_native_versions(version: latest_android_release_name)
+
+  run_pod_install
+end
+
+desc "Update purchases-ios dependency version"
+lane :update_ios_native_versions do |options|
+  if options[:dry_run]
+    dry_run = true
+  end
```
NachoSoto, line 83: I think this could be simply
```suggestion
  dry_run = options[:dry_run]
```

fastlane/Fastfile
```diff
@@ -59,6 +59,98 @@ lane :github_release do |options|
...
+        increment_build_number(previous_version_number, new_version_number, file_to_update)
+      end
+    end
+  end
+end
+
+desc "Update purchases-android dependency version"
+lane :update_android_native_versions do |options|
+  if options[:dry_run]
+    dry_run = true
```
NachoSoto, line 117: Ditto

NachoSoto (commented):
I think this looks good 👍🏻

fastlane/Fastfile
```diff
@@ -59,6 +59,94 @@ lane :github_release do |options|
   )
 end
 
+desc "Update dependencies to latest GitHub releases"
+lane :update_native_dependencies_to_latest do |options|
+  latest_ios_release_info = get_latest_github_release(url: "RevenueCat/purchases-ios")
```
tonidero, line 64: Hmm more than a `url` this seems more a `repo` or `repo_name`... What do you think?

fastlane/Fastfile
```diff
@@ -59,6 +59,94 @@ lane :github_release do |options|
...
+  latest_android_release_info = get_latest_github_release(url: "RevenueCat/purchases-android")
+
+  latest_ios_release_name = latest_ios_release_info["name"]
```
tonidero, line 67: Do we foresee using other info from the latest release other than the name? If not, I suggest renaming `get_latest_github_release` to `get_latest_github_release_name` and return that directly in the action.

fastlane/Fastfile
```diff
@@ -59,6 +59,94 @@ lane :github_release do |options|
...
+  UI.message("ℹ️  Latest RevenueCat Android version: #{latest_android_release_name}")
+
+  update_ios_native_versions(version: latest_ios_release_name)
+  update_android_native_versions(version: latest_android_release_name)
+
+  run_pod_install
+end
+
+desc "Update purchases-ios dependency version"
+lane :update_ios_native_versions do |options|
```
tonidero, line 80: Hmm I feel this should be singular? As in `update_ios_native_version`. Same for the android lane.

fastlane/Fastfile
```diff
@@ -59,6 +59,94 @@ lane :github_release do |options|
...
+  if options[:version]
+    new_version_number = options[:version]
+  else
+    UI.user_error!("Missing `version` argument")
+  end
+
+  previous_version_number = parse_previous_android_native_version
+
+  # Add to options so if we're chaining lanes together, the other lanes get it
+  options[:previous_ios_version] = previous_version_number
```
tonidero, line 124: This should be `previous_android_version`?

fastlane/Fastfile
```diff
@@ -59,6 +59,94 @@ lane :github_release do |options|
...
+  if options[:version]
+    new_version_number = options[:version]
+  else
+    UI.user_error!("Missing `version` argument")
+  end
+
+  previous_version_number = parse_previous_ios_native_version
+
+  # Add to options so if we're chaining lanes together, the other lanes get it
+  options[:previous_ios_version] = previous_version_number
```
tonidero, line 91: I wonder if this (and the android lane version) are really needed... Do you foresee it being necessary for something?

fastlane/actions/get_latest_github_release.rb
```diff
@@ -0,0 +1,170 @@
...
+    module SharedValues
+      GET_LATEST_GITHUB_RELEASE_CUSTOM_VALUE = :GET_LATEST_GITHUB_RELEASE_CUSTOM_VALUE
+    end
+
+    class GetLatestGithubReleaseAction < Action
+      def self.run(params)
+        UI.message("Getting latest release on GitHub (#{params[:server_url]}/#{params[:url]})")
+
+        GithubApiAction.run(
+          server_url: params[:server_url],
```
tonidero, line 12: Hmm unless we pretend to customize this for any reason, I would just hardcode things that will not change like the `server-url`. Also, we probably will use an `api_token` only, so I think we can remove the `api_bearer` parameter

fastlane/actions/get_latest_github_release.rb
```diff
@@ -0,0 +1,170 @@
...
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "This will get latest release version that is available on GitHub"
+      end
+
+      def self.details
+        sample = <<-SAMPLE.markdown_sample
```
tonidero, line 53: If we end up returning the latest release `name` directly, instead of this blob, we can probably remove this.

fastlane/actions/get_latest_github_release.rb
```diff
@@ -0,0 +1,170 @@
...
+              UI.error("GitHub responded with #{result[:status]}:#{result[:body]}")
+              return nil
+            end
+          }
+        ) do |result|
+          json = result[:json]
+          json.each do |current|
+            Actions.lane_context[SharedValues::GET_GITHUB_RELEASE_INFO] = current
+            UI.message("Version #{current['tag_name']} is latest release live on GitHub.com 🚁")
+            return current
```
tonidero, line 36: Hmm I guess this code assumes the endpoint `/repos/repo-name/releases` returns releases in order... I guess this is true and I doubt it will change but not sure I like trusting on a 3rd party service like that... I checked the API docs and I don't see a sorting parameter, but it doesn't mention anything about the order either (https://docs.github.com/en/rest/releases/releases#list-releases)... Maybe we should sort the releases by date? 

That said, the number of releases increases over time, so it might get to a point where it's inefficient, so I'm ok if we want to trust github to not change this. We can probably notice when it's creating the changelog anyway. In that case, I would change the `each` on the `json` with a `first` assuming it's an `Array` which I'm not sure (https://apidock.com/ruby/Array/first)

tonidero (commented):
A few things but looking good

fastlane/Fastfile
vegaro: Removing. This actually doesn't work and I ended up doing it in a different way in https://github.com/RevenueCat/purchases-hybrid-common/pull/179 (which is based on top of this PR)

fastlane/actions/get_latest_github_release.rb
vegaro: I based this action from `get_github_release` https://docs.fastlane.tools/actions/get_github_release/ and thought we could maybe publish it initially, but I think it's better to simplify it and just do as you say and remove all this stuff we are not using

fastlane/actions/get_latest_github_release.rb
vegaro: Hmmm that's true. I actually just realized too that sometimes the latest published release is actually not the latest release, like for example when publishing a hotfix on an older version. I think we should order by major, minor, patch instead. There's probably code written somewhere

fastlane/Fastfile
```diff
@@ -59,6 +59,95 @@ lane :github_release do |options|
...
+    new_version_number = options[:version]
+  else
+    UI.user_error!("Missing `version` argument")
+  end
+
+  previous_version_number = parse_previous_ios_native_version
+
+  if previous_version_number == new_version_number
+    UI.message("ℹ️  RevenueCat iOS is already in its latest version: #{previous_version_number}")
+    next
```
tonidero, line 89: I would use `return` here and the android version

fastlane/Fastfile
```diff
@@ -59,6 +59,95 @@ lane :github_release do |options|
...
+  UI.message("ℹ️  Latest RevenueCat Android version: #{latest_android_release}")
+
+  update_ios_native_version(version: latest_ios_release)
+  update_android_native_version(version: latest_android_release)
+
+  run_pod_install
+end
+
+desc "Update purchases-ios dependency version"
+lane :update_ios_native_version do |options|
```
tonidero, line 77: I think we can probably extract most of this logic to a function in this Fastfile so it can be shared with the android lane.

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,105 @@
...
+            if (Gem::Version.new(highest['tag_name']) < Gem::Version.new(current['tag_name']))
+              highest = current
+            end
+          end
+          Actions.lane_context[SharedValues::GET_LATEST_GITHUB_RELEASE_INFO] = highest
+          UI.message("Version #{highest['tag_name']} is latest release live on GitHub.com 🚁")
+          return highest['tag_name']
+        end
+
+        UI.important("Couldn't find latest GitHub release")
```
tonidero, line 45: Should we throw an error instead? I think that can be done with `UI.user_error!("error message")`

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,105 @@
...
+          json = result[:json]
+          highest = json.last
+          json.each do |current|
+            next if current["prerelease"]
+            if (Gem::Version.new(highest['tag_name']) < Gem::Version.new(current['tag_name']))
+              highest = current
+            end
+          end
+          Actions.lane_context[SharedValues::GET_LATEST_GITHUB_RELEASE_INFO] = highest
+          UI.message("Version #{highest['tag_name']} is latest release live on GitHub.com 🚁")
```
tonidero, line 41: Nitpick. We are using `highest['tag_name']` twice, should extract to a local variable.

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,105 @@
...
+              return nil
+            end,
+            '*' => proc do |result|
+              UI.error("GitHub responded with #{result[:status]}:#{result[:body]}")
+              return nil
+            end
+          }
+        ) do |result|
+          json = result[:json]
+          highest = json.last
```
tonidero, line 33: Nitpick. Don't think it's an issue, but we can throw an error if there are no releases.

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,105 @@
...
+      def self.output
+        [
+          ['GET_LATEST_GITHUB_RELEASE_INFO', 'Contains all the information about this release']
+        ]
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :repo_name,
+                                       env_name: "FL_GET_GITHUB_RELEASE_URL",
```
tonidero, line 70: Looking at moving this to the plugin later, since the org will always be the same here, we can probably just pass the name of the repo (`purchases-ios` or `purchases-android` for example). In the plugin, I already have a similar parameter: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/blob/create-plugin/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb#L82-L86. In any case, I can do this later once I get to moving this to the plugin.

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,105 @@
...
+              UI.error("GitHub responded with #{result[:status]}:#{result[:body]}")
+              return nil
+            end
+          }
+        ) do |result|
+          json = result[:json]
+          highest = json.last
+          json.each do |current|
+            next if current["prerelease"]
+            if (Gem::Version.new(highest['tag_name']) < Gem::Version.new(current['tag_name']))
```
tonidero, line 36: Nice!

tonidero (commented):
Looking great! Just left a few more comments but nothing major

fastlane/Fastfile
```diff
@@ -219,3 +308,24 @@ def replace_in(previous_text, new_text, path, allow_empty=false)
   backup_extension = '.bck'
   sh("sed", '-i', backup_extension, sed_regex, path)
 end
+
+def parse_previous_android_native_version
+  Dir.chdir(get_root_folder) do
+    return (sh 'cat android/build.gradle | grep "ext.purchases_version = " | awk \'{print($3)}\' | sed "s/\'//g"').strip
```
aboedo, line 314: maybe we can do this in ruby instead? 
it might be slightly longer but will definitely be much more readable

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,105 @@
+module Fastlane
+  module Actions
+    module SharedValues
+      GET_LATEST_GITHUB_RELEASE_INFO = :GET_LATEST_GITHUB_RELEASE_INFO
+    end
+
+    class GetLatestGithubReleaseNameAction < Action
```
aboedo, line 7: awesome

fastlane/Fastfile
```diff
@@ -59,6 +59,95 @@ lane :github_release do |options|
   )
 end
 
+desc "Update dependencies to latest GitHub releases"
+lane :update_native_dependencies_to_latest do |options|
+  latest_ios_release = get_latest_github_release_name(repo_name: "RevenueCat/purchases-ios")
+  latest_android_release = get_latest_github_release_name(repo_name: "RevenueCat/purchases-android")
```
aboedo, line 65: I think we should filter for "latest within the same major", otherwise we might be auto-bumping into a breaking version

fastlane/Fastfile
```diff
@@ -219,3 +308,24 @@ def replace_in(previous_text, new_text, path, allow_empty=false)
...
+def parse_previous_android_native_version
+  Dir.chdir(get_root_folder) do
+    return (sh 'cat android/build.gradle | grep "ext.purchases_version = " | awk \'{print($3)}\' | sed "s/\'//g"').strip
+  end
+end
+
+
+def parse_previous_ios_native_version
+  Dir.chdir(get_root_folder) do
+    return (sh 'cat PurchasesHybridCommon.podspec | grep "\'RevenueCat\', " | awk \'{print($3)}\' | sed "s/\'//g"').strip
```
aboedo, line 321: we should extract the common logic here for these two

fastlane/Fastfile
aboedo: +1

aboedo (commented):
Looking great!! I left a few comments. I'm super excited about this

fastlane/Fastfile
vegaro: I don't think it's possible to call return like that in Ruby, it gives a `LocalJumpError`. This next will skip the rest of the execution and go back to the line that executed. According to their docs https://docs.fastlane.tools/advanced/lanes/


The next keyword can be used to stop executing a lane before it reaches the end.

lane :build do |options|
  if cached_build_available?
    UI.important 'Skipping build because a cached build is available!'
    next # skip doing the rest of this lane
  end
  match
  gym
end

private_lane :cached_build_available? do |options|
  # ...
  true
end
When next is used during a lane switch, control returns to the previous lane that was executing.

lane :first_lane do |options|
  puts "If you run: `fastlane first_lane`"
  puts "You'll see this!"
  second_lane
  puts "As well as this!"
end

private_lane :second_lane do |options|
  next
  puts "This won't be shown"
end
When you stop executing a lane early with next, any after_each and after_all blocks you have will still trigger as usual :+1:

fastlane/actions/get_latest_github_release_name.rb
vegaro: I did it! Thanks for the feedback 🤟

fastlane/Fastfile
```diff
@@ -59,6 +59,95 @@ lane :github_release do |options|
...
+  latest_android_release = get_latest_github_release_name(repo_name: "RevenueCat/purchases-android")
+
+  UI.message("ℹ️  Latest RevenueCat iOS version: #{latest_ios_release}")
+  UI.message("ℹ️  Latest RevenueCat Android version: #{latest_android_release}")
+
+  update_ios_native_version(version: latest_ios_release)
+  update_android_native_version(version: latest_android_release)
+
+  run_pod_install
+end
```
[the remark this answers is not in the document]
vegaro: Thanks!!! 🤟

fastlane/Fastfile
vegaro: true. I thought in that case we just wouldn't merge it, but it's true that the PR will basically have broken code. I am going to add that check

fastlane/Fastfile
vegaro: I like this challenge. I am going to try haha because this is very ilegible

fastlane/Fastfile
tonidero: Ahh I didn't know that about fastlane. Makes sense!

fastlane/Fastfile
vegaro: Ok I think I did it haha [`0fa8d37` (#178)](https://github.com/RevenueCat/purchases-hybrid-common/pull/178/commits/0fa8d3749074a8af9c75dcb5770d1748f5a2386a)

fastlane/Fastfile
```diff
@@ -219,3 +308,32 @@ def replace_in(previous_text, new_text, path, allow_empty=false)
...
+
+def parse_previous_ios_native_version
+  parse_version_in_file(/(?<=s\.dependency ["']RevenueCat["'], ["'])(.*)(?=["'])/, "PurchasesHybridCommon.podspec")
+end
+
+def parse_version_in_file(pattern, path)
+  Dir.chdir(get_root_folder) do
+    File.open path do |file|
+      line = file.find { |line| line =~ pattern }
+      unless line == nil
```
aboedo, line 324: can't recall, but I think the ruby way of doing this was `unless line.nil?`

fastlane/Fastfile
```diff
@@ -219,3 +308,32 @@ def replace_in(previous_text, new_text, path, allow_empty=false)
...
+        match = line.match(pattern)
+        return match[1]
+      end
+    end
+  end
+  UI.user_error!("Couldn't extract version from #{path}")
+end
+
+def get_root_folder
+  return File.expand_path('../../', __FILE__)
```
aboedo, line 334: this can just be 
```suggestion
  File.expand_path('../../', __FILE__)
```
since explicit returns aren't needed for ruby, it'll just always return the last value

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,110 @@
...
+        GithubApiAction.run(
+          server_url: "https://api.github.com",
+          api_token: ENV["GITHUB_API_TOKEN"],
+          api_bearer: nil,
+          http_method: 'GET',
+          path: "repos/RevenueCat/#{params[:repo_name]}/releases",
+          error_handlers: {
+            404 => proc do |result|
+              UI.error("Repository #{params[:repo_name]} cannot be found, please double check its name and that you provided a valid API token (if it's a private repository).")
+              return nil
```
aboedo, line 20: I think this is the same as just doing `return`, since ruby always returns a value, which defaults to `nil`

aboedo: with these being error handlers, they probably don't need the return line at all, though, right? Like, the code won't continue after the `UI.error` line, right?

aboedo: so you could simplify this block to 
```suggestion
            404 => proc do |result|
              UI.error("Repository #{params[:repo_name]} cannot be found, please double check its name and that you provided a valid API token (if it's a private repository).")
            end,
            401 => proc do |result|
              UI.error("You are not authorized to access #{params[:repo_name]}, please make sure you provided a valid API token.")
            end,
            '*' => proc do |result|
              UI.error("GitHub responded with #{result[:status]}:#{result[:body]}")
            end
          end
```

fastlane/actions/get_latest_github_release_name.rb
```diff
@@ -0,0 +1,110 @@
...
+            end,
+            '*' => proc do |result|
+              UI.error("GitHub responded with #{result[:status]}:#{result[:body]}")
+              return nil
+            end
+          }
+        ) do |result|
+          json = result[:json]
+          highest = json.last
+          unless highest == nil
```
aboedo, line 34: you can unnest this block by checking the precondition first, like
```ruby
if highest.nil?
  UI.user_error!("Couldn't find latest GitHub release for #{params[:repo_name]}")
  return
end

highest_tag_name = highest['tag_name']
...
```

fastlane/Fastfile
vegaro: Done in [`b7a5bac` (#178)](https://github.com/RevenueCat/purchases-hybrid-common/pull/178/commits/b7a5baca32136d03e9e89d0fee40c754c6138bf6)

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -15,38 +15,34 @@ def self.run(params)
...
-              if (highest_version < item_version)
-                highest_version = item_version
-              end
+          if json.count == 0
+            UI.user_error!("Couldn't find latest GitHub release for #{params[:repo_name]}")
+            return
+          end
+          highest_version = current_version
+          json.each do |item|
+            next if item["prerelease"]
```
aboedo, line 35: you could probably one-line this by doing
```suggestion
          json.reject {  |item| item.key?("prerelease") } each do |item|
```

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -0,0 +1,100 @@
...
+          }
+        ) do |result|
+          current_version = Gem::Version.new(params[:current_version])
+          json = result[:json]
+          if json.count == 0
+            UI.user_error!("Couldn't find latest GitHub release for #{params[:repo_name]}")
+            return
+          end
+          highest_version = current_version
+          json.reject {  |item| item.key?("prerelease") }.each do |item|
```
tonidero, line 34: There are 2 spaces here

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -0,0 +1,100 @@
...
+      end
+
+      def self.example_code
+        [
+          'release_name = get_latest_github_release_within_same_major(repo_name: "fastlane/fastlane")
+          puts release_name'
+        ]
+      end
+
+      def self.sample_return_value
```
tonidero, line 89: I guess this would need to be updated or maybe removed... I believe this is useful if we are publishing the action somewhere but considering this will likely remain internal (probably just moved to our internal plugin)

tonidero (approved):
LGTM!

fastlane/Fastfile
```diff
@@ -219,3 +307,32 @@ def replace_in(previous_text, new_text, path, allow_empty=false)
   backup_extension = '.bck'
   sh("sed", '-i', backup_extension, sed_regex, path)
 end
+
+def parse_previous_android_native_version
+  parse_version_in_file(/(?<=ext\.purchases_version = ["'])(.*)(?=["'])/, "android/build.gradle")
```
tonidero, line 312: One thing I was thinking about is that we can maybe use the `'fastlane-plugin-versioning_android'` plugin to get the current version, same way we do in the `purchases-android` repo... That said, I think it's better to keep it as you have it, since it's the only usage, and that wouldn't work in the case of iOS.

fastlane/Fastfile
```diff
@@ -219,3 +307,32 @@ def replace_in(previous_text, new_text, path, allow_empty=false)
...
+end
+
+def parse_version_in_file(pattern, path)
+  Dir.chdir(get_root_folder) do
+    File.open path do |file|
+      line = file.find { |line| line =~ pattern }
+      unless line.nil?
+        match = line.match(pattern)
+        return match[1]
+      end
```
aboedo, line 326: this block could also be simplified as 
```suggestion
      return line.match(pattern)[1] unless line.nil?
```

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -0,0 +1,100 @@
...
+          json = result[:json]
+          if json.count == 0
+            UI.user_error!("Couldn't find latest GitHub release for #{params[:repo_name]}")
+            return
+          end
+          highest_version = current_version
+          json.reject {  |item| item.key?("prerelease") }.each do |item|
+            item_version = Gem::Version.new(item['tag_name'])
+            # Skip if it's not in the same major range. Ex. 3.0.0 vs 4.2.0 
+            next if item_version.canonical_segments[0] != current_version.canonical_segments[0]
```
aboedo, line 37: you could also extract the line into a function for readability, i.e.: 
```ruby
next if !same_major_range?(item_version, current_version)

...

def same_major_range?(version_a, version_b)
    version_a.canonical_segments[0] == version_b.canonical_segments[0]
end

```

aboedo: furthermore, since this is still filtering items on the list, you could add it to the filters, i.e.: 
```ruby

def item_version(item)
  Gem::Version.new(item['tag_name'])
end

highest_version = json
  .reject { |item| item.key?("prerelease") }
  .map { |item| item_version(item) }
  .select { |item_version| same_major_range?(item_version, current_version) }
  .max
```

aboedo: disclaimer: I didn't test any of that code 😅 so apologies if it doesn't work as is, but you get the idea

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -0,0 +1,100 @@
...
+            if (highest_version < item_version)
+              highest_version = item_version
+            end
+          end
+          UI.message("Version #{highest_version.version} is latest release live on GitHub.com 🚁")
```
aboedo, line 42: the latest release in the same major*

aboedo (commented):
I'm super excited about this 😃 

Auto-upgrading the hybrids isn't just gonna save a lot of work (and it definitely will!!), it's also going to guarantee that our customers get bug fixes and improvements constantly with minimal wait time.

fastlane/Fastfile
vegaro: that gives you the current version of the project, but not the current version of the dependency, which is what we want in this case. So I don't think we can use that plugin

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -0,0 +1,99 @@
...
+            .map { |item| item_version(item) }
+            .select { |item_version| same_major_range?(item_version, current_version) }
+            .max
+
+          UI.message("Version #{highest_version.version} is latest release in the same major live on GitHub.com 🚁")
+          return highest_version.version
+        end
+      end
+
+      private_class_method def self.item_version(item)
```
vegaro, line 44: I think this is the correct way of defining a private method but I am not sure lol

tonidero: Yeah this is fine, this works for "static" methods such as these. For normal methods, you need to add them after a `private` delimiter (http://ruby-for-beginners.rubymonstas.org/advanced/private_methods.html). In any case, this is correct 👍

fastlane/actions/get_latest_github_release_within_same_major.rb
```diff
@@ -0,0 +1,99 @@
+module Fastlane
+  module Actions
+
+    class GetLatestGithubReleaseWithinSameMajorAction < Action
```
tonidero, line 4: You should rename the file to `get_latest_github_release_within_same_major_action.rb`

fastlane/actions/get_latest_github_release_within_same_major_action.rb
```diff
@@ -0,0 +1,99 @@
...
+  module Actions
+
+    class GetLatestGithubReleaseWithinSameMajorAction < Action
+      def self.run(params)
+        repo_name = params[:repo_name]
+        UI.message("Getting latest release for #{repo_name} on GitHub within same major as the current version")
+
+        GithubApiAction.run(
+          server_url: "https://api.github.com",
+          api_token: ENV["GITHUB_API_TOKEN"],
```
tonidero, line 11: This could be set as an optional parameter, similar to how I do it here: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/blob/main/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb#L89-L93

Also, I think it's not really needed since our repos are public so not sure if we want to remove it entirely (we should test if we decide to remove it)

fastlane/actions/get_latest_github_release_within_same_major_action.rb
```diff
@@ -0,0 +1,99 @@
...
+          path: "repos/RevenueCat/#{repo_name}/releases",
+          error_handlers: {
+            404 => proc do |result|
+              UI.error("Repository #{repo_name} cannot be found, please double check its name and that you provided a valid API token (if it's a private repository).")
+            end,
+            401 => proc do |result|
+              UI.error("You are not authorized to access #{repo_name}, please make sure you provided a valid API token.")
+            end,
+            '*' => proc do |result|
+              UI.error("GitHub responded with #{result[:status]}:#{result[:body]}")
```
tonidero, line 23: Hmm I wonder if these should be `UI.user_error!` instead... I think with these, the lane will continue even if the request fails...

fastlane/actions/get_latest_github_release_within_same_major_action.rb
```diff
@@ -0,0 +1,99 @@
...
+            end
+          }
+        ) do |result|
+          current_version = Gem::Version.new(params[:current_version])
```
tonidero, line 27: Nitpick. Should we do this before performing the `GithubApiAction`? We have the information before, so I think it makes sense to do it early... NABD though

tonidero (approved):
A couple comments but nothing major. Looks good!

tonidero:
Oh actually, it seems it can't find the action after the file rename... You might have to revert that @vegaro... I do name the files with `_action` in the plugin, weird...

vegaro:
> Oh actually, it seems it can't find the action after the file rename... You might have to revert that @vegaro... I do name the files with _action in the plugin, weird...

yeah, weird!

fastlane/actions/get_latest_github_release_within_same_major_action.rb
vegaro: Note this is me calling another action, not part of the action's parameters. In any case, I thought it was required by `github_action` but it doesn't look like it is? https://docs.fastlane.tools/actions/github_api/ 

Maybe we can just remove it?

fastlane/actions/get_latest_github_release_within_same_major_action.rb
vegaro: I agree, that way we can also print it

fastlane/actions/get_latest_github_release_within_same_major_action.rb
vegaro: Hmm, maybe yes. I got inspired by this official action https://github.com/fastlane/fastlane/blob/master/fastlane/lib/fastlane/actions/get_github_release.rb#L19

And they do `error` there. But happy to change it to `user_error`.

vegaro: Ok, yes I am going to change them. Looking at fastlane docs for `user_error`:

```
 fastlane "crash" because of a user error everything that is caused by the user and is not unexpected
 ```
 
 It looks like most of these errors are caused by a bad configuration, and we want to crash, so I will change them.

fastlane/actions/get_latest_github_release_within_same_major_action.rb
vegaro: Ok, I see what you mean. What would be the advantage of setting it as an optional parameter versus removing it? What should we test if we decide to remove it?

tonidero: Well setting it as an optional parameter would allow callers to pass the token either as a parameter or through an environment variable. This would be needed if we wanted to access private repos if we ever need to in the future. Besides that, authenticated requests have some advantages such as increased rate limits. 

That said, i don't think those factors should affect us... So i might just remove it to keep it simple. As long as it works without any token.

vegaro: I removed it along with the bearer. We can always bring them back if we need to

PR #179, opened 2022-07-25
Schedule automatic PR to update native dependencies

vegaro:
Depends on https://github.com/RevenueCat/purchases-hybrid-common/tree/support-fastlane-plugin-revenuecat_internal

CSDK-347


fastlane/Fastfile
```diff
@@ -300,12 +358,13 @@ def is_snapshot_version?(version_name)
...
+  # temporary until we have shared plugin in place
   if new_text.to_s.strip.empty? and not allow_empty
     UI.user_error!("Missing `new_text` in call to `replace_in`, looking for replacement for #{previous_text} 😵.")
   end
-  sed_regex = 's|' + previous_text.sub(".", "\\.") + '|' + new_text + '|'
-  backup_extension = '.bck'
-  sh("sed", '-i', backup_extension, sed_regex, path)
+  original_text = File.read(path)
+  replaced_text = original_text.gsub(previous_text, new_text)
+  File.write(path, replaced_text)
```
vegaro, line 367: I took this from the plugin https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal, which will be added in https://github.com/RevenueCat/purchases-hybrid-common/pull/186. The `sed` command wasn't working in the docker machine.

vegaro:
Just to prove that it works: https://app.circleci.com/pipelines/github/RevenueCat/purchases-hybrid-common/1045/workflows/83088db1-4469-4eba-894f-e0fdc3d95ac6/jobs/2368

fastlane/Fastfile
```diff
@@ -61,6 +61,58 @@ lane :github_release do |options|
   )
 end
 
+desc "Update dependencies to latest GitHub releases and opens a PR"
+lane :open_pr_upgrading_dependencies do |options|
+  if options[:dry_run]
+    dry_run = true
+  end
+  
+  new_versions = update_native_dependencies_to_latest()
```
tonidero, line 70: No need to add the parentheses if there are no parameters
```suggestion
  new_versions = update_native_dependencies_to_latest
```

fastlane/Fastfile
```diff
@@ -61,6 +61,58 @@ lane :github_release do |options|
...
+  
+  repo_status = Actions.sh("git status --porcelain")
```
tonidero, line 72: In the `Fastfile`, you shouldn't need to add the `Actions`. Same in the other sh scripts
```suggestion
  repo_status = sh("git status --porcelain")
```

fastlane/Fastfile
```diff
@@ -61,6 +61,58 @@ lane :github_release do |options|
...
+
+  if repo_clean
+    UI.success('Git status is clean, nothing to update! 💪')
+    next
+  end
+
+  new_ios_version = new_versions[:new_ios_version]
+  new_android_version = new_versions[:new_android_version]
+
+  branch = "update-#{new_ios_version}-#{new_android_version}"
```
tonidero, line 84: Just a suggestion: 
```suggestion
  branch = "update-ios-#{new_ios_version}-android-#{new_android_version}"
```

fastlane/Fastfile
```diff
@@ -61,6 +61,58 @@ lane :github_release do |options|
...
+  
+  result = Actions.sh("git checkout -b #{branch}")
+  if result =~ /error:(.)/
+    UI.error("Git couldn't change to `#{branch}`. Reason: #{result}.")
```
tonidero, line 88: Note that the script would continue when using `UI.error`. Either use `user_error!` or throw an exception. Same in the errors below

fastlane/Fastfile
```diff
@@ -61,6 +61,58 @@ lane :github_release do |options|
...
+    UI.success('Git status is clean, nothing to update! 💪')
+    next
+  end
+
+  new_ios_version = new_versions[:new_ios_version]
+  new_android_version = new_versions[:new_android_version]
+
+  branch = "update-#{new_ios_version}-#{new_android_version}"
+  
+  result = Actions.sh("git checkout -b #{branch}")
```
tonidero, line 86: Hmm I would think creating a new branch shouldn't be done in a `dry-run`, same with committing changes. Maybe not do those in a `dry-run` and just print the versions?

fastlane/Fastfile
tonidero: Also you can change this if block and others with a single statement with:
```suggestion
    UI.error("Git couldn't change to `#{branch}`. Reason: #{result}.") if result =~ /error:(.)/
```
It's common to put it in one line if there is only one statement

fastlane/Fastfile
```diff
@@ -61,6 +61,58 @@ lane :github_release do |options|
...
+  if result =~ /error:(.)/
+    UI.error("Git couldn't create commit. Reason: #{result}.")
+  end
+
+  if dry_run
+    UI.message("ℹ️  Nothing more to do, dry_run: true")
+  else
+    push_to_git_remote(
+      remote: "origin",
+      set_upstream: true
```
tonidero, line 103: Hmm are these parameters mandatory? I think these were the default but don't fully remember

fastlane/Fastfile
tonidero: Also, this will checkout from the current branch.... I wonder if we should ensure this is executed from the `main` branch and that it's updated... I guess we can add those validations later if needed.

fastlane/Fastfile
```diff
@@ -73,6 +125,12 @@ lane :update_native_dependencies_to_latest do |options|
   update_android_native_version(new_version: latest_android_release, current_version: current_android_version)
 
   run_pod_install
+
+  new_versions = {
+    new_ios_version: latest_ios_release,
+    new_android_version: latest_android_release
+  }
+  new_versions
```
tonidero, line 133: Hmm what's the reason to return this hash? Is this lane going to be used from a different lane?

fastlane/Fastfile
```diff
@@ -98,7 +156,7 @@ private_lane :update_android_native_version do |options|
   new_version_number = options[:new_version]
   previous_version_number = options[:current_version]
   files_to_update = [
-    'android/build.gradle',
+    File.join(get_root_folder, 'android/build.gradle'),
```
tonidero, line 159: Hmm is this needed?

tonidero (commented):
Looking good!

fastlane/Fastfile
vegaro: I am going to change the command to be `git checkout -b #{branch} main` so that the branch is off main

fastlane/Fastfile
vegaro: `remote: "origin",` is the default, I will remove that one. `set_upstream` defaults to `false`

fastlane/Fastfile
vegaro: It's used in https://github.com/RevenueCat/purchases-hybrid-common/pull/179/files/f007a2a475bfda2d75d463cb1e1109b255f7c6aa#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R70

fastlane/Fastfile
```diff
@@ -61,6 +61,56 @@ lane :github_release do |options|
...
+
+  new_ios_version = new_versions[:new_ios_version]
+  new_android_version = new_versions[:new_android_version]
+
+  if dry_run
+    UI.message("ℹ️  Nothing more to do since dry_run: true.")
+    UI.message("ℹ️  New iOS dependency version is #{new_ios_version}")
+    UI.message("ℹ️  New Android dependency version is #{new_android_version}")
+  else
+    branch = "update-ios-#{new_ios_version}-android-#{new_android_version}"
```
tonidero, line 87: Note that it's possible this only updates one of the SDKs. I don't think it's a big deal though

fastlane/Fastfile
```diff
@@ -61,6 +61,56 @@ lane :github_release do |options|
...
+    next
+  end
+
+  new_ios_version = new_versions[:new_ios_version]
+  new_android_version = new_versions[:new_android_version]
+
+  if dry_run
+    UI.message("ℹ️  Nothing more to do since dry_run: true.")
+    UI.message("ℹ️  New iOS dependency version is #{new_ios_version}")
+    UI.message("ℹ️  New Android dependency version is #{new_android_version}")
```
tonidero, line 85: I guess on a `dry-run` it would still update the files, but I think it's NABD... we could remove all changes I guess, but then we should ensure the repo is clean before starting

fastlane/Fastfile
```diff
@@ -61,6 +61,56 @@ lane :github_release do |options|
...
+  else
+    branch = "update-ios-#{new_ios_version}-android-#{new_android_version}"
+
+    result = sh("git checkout -b #{branch} origin/main")
+  
+    UI.user_error!("Git couldn't change to `#{branch}`. Reason: #{result}.") if result =~ /error:(.)/
```
tonidero, line 91: I believe if the `sh` command fails it should also cause the whole lane to fail. This is mostly to make the error better looking right? (which is great! 😄)

tonidero (approved):
LGTM!

fastlane/Fastfile
vegaro: Oh, so you think it won't execute?

tonidero: I think so but we should probably test it. If you've tested and it does execute it, I think it's fine

fastlane/Fastfile
vegaro: I am going to add the clean up if it's a dry run

fastlane/Fastfile
vegaro: you're right, it just fails and it doesn't execute it, so we can remove these!

fastlane/Fastfile
```diff
@@ -61,6 +61,52 @@ lane :github_release do |options|
...
+  end
+
+  new_ios_version = new_versions[:new_ios_version]
+  new_android_version = new_versions[:new_android_version]
+
+  if dry_run
+    UI.message("ℹ️  Nothing more to do since dry_run: true.")
+    UI.message("ℹ️  New iOS dependency version is #{new_ios_version}")
+    UI.message("ℹ️  New Android dependency version is #{new_android_version}")
+    reset_git_repo
```
tonidero, line 86: Since you added this, I suggest we check that the git status is clean before starting this lane to make sure we don't reset anything accidentally.

vegaro:
Alright I added a `stash` and `stash pop` around the change of branches so it doesn't fail

PR #180, opened 2022-07-25
iOS Integration Tests: retry on failure

NachoSoto:
Equivalent to https://github.com/RevenueCat/purchases-ios/pull/1767
Also made the maximum execution time 3 minutes instead of the default 10.


taquitos (approved):
🕺🎈🐐

PR #186, opened 2022-08-01
Support fastlane plugin revenuecat internal

tonidero:
This PR changes how we bump versions in `purchases-hybrid-common` to make use of the new fastlane plugin https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal. 

### TODO
- [x] Wait until https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/8 is merged and change to use a git reference instead of path references for the plugin
- [ ] Check whether we want to add a `prepare_next_version` lane and introduce snapshots as part of the release process for PHC.


tonidero (commented):
There is an open question for the team but that can be addressed in a future PR so this is now final

fastlane/Fastfile
```diff
@@ -145,31 +135,6 @@ platform :android do
   end
 end
 
-def attach_changelog_to_main(options)
```
aboedo, line 148: look at all that duplicated code go :boom:

aboedo (approved):

fastlane/Fastfile
```diff
@@ -74,13 +76,13 @@ lane :update_native_dependencies_to_latest do |options|
 end
 
 desc "Update purchases-ios dependency version"
-lane :update_ios_native_version do |options|
+private_lane :update_ios_native_version do |options|
```
tonidero, line 79: I changed these to `private_lane` since I think it's better to keep a cleaner API for fastlane, and having the `update_native_dependencies_to_latest` should be enough I think.

fastlane/Fastfile
```diff
@@ -74,13 +76,13 @@ lane :update_native_dependencies_to_latest do |options|
...
   new_version_number = options[:new_version]
   previous_version_number = options[:current_version]
   files_to_update = [
-    'PurchasesHybridCommon.podspec',
-    'ios/PurchasesHybridCommon/Podfile',
-    'DEVELOPMENT.md',
+    File.join(get_root_folder, 'PurchasesHybridCommon.podspec'),
+    File.join(get_root_folder, 'ios/PurchasesHybridCommon/Podfile'),
+    File.join(get_root_folder, 'DEVELOPMENT.md')
```
tonidero, line 85: So I changed this because I was finding some weird behavior with fastlane when replacing the versions. 

Basically we were doing `Dir.chdir(get_root_folder)` during the `update_native_version ` lane. However, after calling the `replace_text_in_files` action from the plugin, the path changed to the parent of the project path. This did not happen if we did not call `Dir.chdir(get_root_folder)`. In order to solve this, while keeping the ability to run this from anywhere, I changed it to passing absolute paths like this. That said I do wonder if we want to support running this lane from different directories, since in most cases it will be run from the root folder and this wouldn't be needed and we can just use relative paths.

tonidero (commented):
@vegaro I had to solve some conflicts and had to modify your code a bit. Please take a look when you have the chance!

fastlane/Fastfile
vegaro: I think we can just keep it simple and use relative paths like we do in iOS

vegaro (approved):

PR #190, opened 2022-08-03
Upgrade iOS to 4.9.1 and Android to 5.4.0

RCGitBot:
Upgrade iOS to 4.9.1 and Android to 5.4.0


fastlane/Fastfile
```diff
@@ -59,6 +59,58 @@ lane :github_release do |options|
...
+
+  if dry_run
+    UI.message("ℹ️  Nothing more to do, dry_run: true")
+  else
+    push_to_git_remote(
+      remote: "origin",
+      set_upstream: true
+    )
+    create_pull_request(
+      api_token: ENV["GITHUB_TOKEN"],
```
vegaro, line 104: This has to be explicitly set otherwise it doesn't work, I think it might be trying to get another ENV variable

PR #194, opened 2022-08-08
Update iOS 4.10.0 (and add AdServices)

joshdholtz:
- Update iOS to 4.10.0 with AdServices
  - New public method `enableAdServicesAttributionTokenCollection`

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -66,6 +66,12 @@ import RevenueCat
         Purchases.automaticAppleSearchAdsAttributionCollection = enabled
     }
 
+    @objc static func enableAdServicesAttributionTokenCollection() {
+        if #available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *) {
+            Purchases.shared.attribution.enableAdServicesAttributionTokenCollection()
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -63,7 +63,15 @@ import RevenueCat
     }
 
     @objc public static func setAutomaticAppleSearchAdsAttributionCollection(_ enabled: Bool) {
-        Purchases.automaticAppleSearchAdsAttributionCollection = enabled
+        // Using `setValue` because `automaticAppleSearchAdsAttributionCollection` is now deprecated and we get a warning
+        Purchases.setValue(enabled, forKey: "automaticAppleSearchAdsAttributionCollection")
```
taquitos, line 67: CatBobbing

taquitos (approved):
🎈🐐

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
NachoSoto: ⚠️  this seems super fragile.
Why don't we mark `setAutomaticAppleSearchAdsAttributionCollection` as deprecated as well? That way we don't need the hack, and we propagate the deprecation.

NachoSoto: #199

PR #197, opened 2022-08-12
added missing availability check for catalyst

aboedo:
Added missing availability check to fix react native issue


taquitos (approved):
download-39
🎈🐐

PR #199, opened 2022-08-15
`automaticAppleSearchAdsAttributionCollection`: changed implementation to call method directly

NachoSoto:
Follow up to #194. Instead of the hacky `setValue:forKey:`, this marks the method as deprecated instead.


joshdholtz (approved):
Oh this so much better!

joshdholtz (approved):
Oh this so much better!

PR #202, opened 2022-08-17
Release train

vegaro:
Creates a new workflow that will run every time a new commit is merged into `main` and that will perform the following:
- Determine next version to release using PR labels
- If next version to release is not a major release
- Open a PR preparing the release

[CSDK-137]

[CSDK-137]: https://revenuecats.atlassian.net/browse/CSDK-137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -234,5 +275,20 @@ workflows:
...
+        equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
-      - revenuecat/danger
\ No newline at end of file
+      - automatic-release:
+          filters:
+            tags:
+              ignore: /.*/
+            branches:
+              only: main
```
tonidero, line 294: Hmm I wonder if we would like to do this only once a week or possibly a day... Doing it for every merge might be noisy? That said, this should be mostly just ios/android bumps which will happen weekly I think, so this might be ok.

tonidero (approved):

.circleci/config.yml
vegaro: That made me think... merging this PR will also trigger an automatic release. Do we want that? I guess we can close the PR if we don't want to release at that moment. Maybe we can check the labels of the PRs to release and skip the release if it's only test, CI, and those that are not really public facing changes?

tonidero: Ahh good catch! And yeah, if we can filter out those, it should save time.

PR #207, opened 2022-08-19
Upgrade iOS to 4.10.2

NachoSoto:
Upgrade iOS to 4.10.2.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -5,7 +5,7 @@
 In `ios/PurchasesHybridCommon/Podfile` replace:
 
 ```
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Hmm this doesn't actually update Android at all... I think @vegaro fixed that so only the version updated showed... Did you do this PR automatically or manually? In any case, looks good.

tonidero:
Oh tests seems to be failing though

NachoSoto:
I did this manually. 

I still need to fix the iOS build though.

NachoSoto:
Updated the tests for the `internal` API change in the SDK.

PR #211, opened 2022-08-22
Rename revenueCatId and productId to transactionIdentifier and productIdentifier

tonidero:
### Description
This relates to the work in https://revenuecats.atlassian.net/browse/CSDK-409. We want to rename `revenueCatId` and `productId` to `transactionIdentifier` and `productIdentifier` respectively. This adds those properties to the mappings of the `Transaction` class and adds a comment to deprecate the old properties to be removed in a future release.


vegaro (approved):
I think it was a mistake, but this PR has the perf label too

tonidero (commented):
Agh forgot to submit this self-review.

PR #215, opened 2022-08-22
Update common fastlane plugin

vegaro:


tonidero (approved):
It would be great to get automatic updates of this plugin in all SDKs as well, same as we are planning to do on PHC and the hybrids... However, there might still be things that break, specially now that we are still working on it, so maybe we can wait a bit longer before doing something like that.

PR #213, opened 2022-08-22
Replace build with dependencies on automatic upgrade PRs

vegaro:
https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/27#discussion_r950498680

Changing PRs that update iOS/Android to be labelled with `dependencies` instead of `build`

[CSDK-425]

[CSDK-425]: https://revenuecats.atlassian.net/browse/CSDK-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #216, opened 2022-08-22
Release/4.1.4

RCGitBot:
### Other Changes
* Update common fastlane plugin (#215) via Cesar de la Vega (@vegaro)
* Rename revenueCatId and productId to transactionIdentifier and productIdentifier (#211) via Toni Rico (@tonidero)
* Replace build with dependencies on automatic upgrade PRs (#213) via Cesar de la Vega (@vegaro)
* `IntegrationTests`: actually fail test if tests aren't configured (#210) via NachoSoto (@NachoSoto)


CHANGELOG.latest.md
```diff
@@ -1,5 +1,7 @@
+### API Changes
+* Rename revenueCatId and productId to transactionIdentifier and productIdentifier. Old values still exist but are deprecated (#211) via Toni Rico (@tonidero)
```
vegaro, line 2: Makes sense to have it in a separate section. I wonder if either it could have been tagged differently, or if we need a special label for phc. Anyways, we will always have the flexibility to manually edit the Changelog, and I think it's not possible to cover all cases automatically.

vegaro (approved):

CHANGELOG.latest.md
tonidero: Yeah agreed... In this case, I could have labeled it differently, but none of the tags fit exactly... So I think just having the option to update the changelog manually is enough for now.

tonidero (approved):

PR #217, opened 2022-08-22
Fix CircleCI caches

vegaro:
Looks like we had a couple of jobs running on ruby machines that were sharing Gem caches with other type of executors. This was causing permission issues.

Look at the `Restoring caches` step in https://app.circleci.com/pipelines/github/RevenueCat/purchases-hybrid-common/1183/workflows/daae7bec-2246-4a9a-ba84-5a933d32fabf/jobs/2884

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,13 +173,13 @@ jobs:
     steps:
       - checkout
       - revenuecat/install-rubydocker-dependencies
-      - android/accept-licenses
-      - restore_cache:
-          key: jars-{{ checksum "android/build.gradle" }}
-      - android/restore-build-cache
-      - run:
-          name: Deployment
-          command: bundle exec fastlane android deploy
+      # - android/accept-licenses
+      # - restore_cache:
+          # key: jars-{{ checksum "android/build.gradle" }}
+      # - android/restore-build-cache
+      # - run:
+          # name: Deployment
+          # command: bundle exec fastlane android deploy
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -188,7 +188,7 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
-      - install-gems
+      - revenuecat/install-rubydocker-dependencies
```
tonidero, line 191: Hmm I might be wrong but do we also need to update the android jobs? Not sure if the `android-machine` uses mac or linux, but if it's using linux, that should also use `install-rubydocker-dependencies`

vegaro: I think you're correct! The problem comes from the Apple executor storing the caches then the Android machine trying to restore them, not because of this change I just did 😮‍💨

vegaro: Also, we should update to just be `install-docker-dependencies` right?

tonidero: yeah, I think that or even just rename to `install-gem-mac-dependencies` and `install-gem-unix-dependencies` or something like that, since the difference is due to the OS.

vegaro: I can update that and move the install-gem-mac-dependencies to the orb. Also adding a version for invalidation as a parameter of the job will be convenient

tonidero (approved):
Looks good!

PR #218, opened 2022-08-22
Skip next version if there are no public changes

vegaro:
`determine_next_version_using_labels` now returns `:skip` if there were no public changes and next version shouldn't be released


fastlane/Fastfile
```diff
@@ -51,6 +51,7 @@ lane :automatic_bump do |options|
     github_rate_limit: options[:github_rate_limit]
   )
   options[:next_version] = next_version
+  UI.user_error!('Skipping automatic bump since the next version doesn\'t include public facing changes') if type_of_bump == :skip
```
aboedo, line 54: 🤔 public facing changes here might feel like it means a minor release. But this is for anything that isn't CI, right? 
Maybe "non-infrastructure changes"?

tonidero, line 54: Hmm we will get an error on CI if we do it like this... I think we should just log this normally and finish (without executing the bump) instead. The same can be said for the next line, we probably should just log it and finish without errors. I think otherwise we will get a lot of unnecessary errors in CI (even if they are almost invisible, since they are generated automatically).

fastlane/Fastfile
```diff
@@ -51,7 +51,14 @@ lane :automatic_bump do |options|
     github_rate_limit: options[:github_rate_limit]
   )
   options[:next_version] = next_version
-  UI.user_error!('Skipping automatic bump since the next version is a major release') if type_of_bump == :major
+  if type_of_bump == :skip
+    UI.error('Skipping automatic bump since the next version doesn\'t include public facing changes')
```
tonidero, line 55: I think it's better to use `UI.message` for these
```suggestion
    UI.message('Skipping automatic bump since the next version doesn\'t include public facing changes')
```

fastlane/Fastfile
```diff
@@ -51,7 +51,14 @@ lane :automatic_bump do |options|
...
+    next
+  end
+  if type_of_bump == :major
+    UI.error('Skipping automatic bump since the next version is a major release')
```
tonidero, line 59:
```suggestion
    UI.message('Skipping automatic bump since the next version is a major release')
```

tonidero (approved):
LGTM! Just a suggestion on how the messages should appear.

fastlane/Fastfile
vegaro: Sure, I will update. I don't have a preference. Thanks for the suggestion!

PR #223, opened 2022-08-30
Adds hold jobs to manual trigger automatic dependency updates

vegaro:
Adds two hold jobs to `main`, one will trigger `automatic-dependency-update` to update the iOS and Android dependencies, and another one that would trigger `automatic-release`


taquitos (approved):
NOICE.

PR #228, opened 2022-09-04
`Integration Tests`: added test for `promotionalOffer`

NachoSoto:


ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -116,6 +116,34 @@ class StoreKit1IntegrationTests: BaseIntegrationTests {
...
+        }
+    }
+
+    @available(iOS 12.2, macOS 10.14.4, tvOS 12.2, *)
+    func testPromotionalOffer() async throws {
+        try await self.purchaseMonthlyOffering()
+        try self.testSession.expireSubscription(productIdentifier: Self.productIdentifier)
+
+        var result = try await CommonFunctionality.promotionalOffer(
+            for: Self.productIdentifier,
```
tonidero, line 139: We could also add a test that checks that it fails getting a promotional offer if passing an incorrect `productIdentifier` or `discountIdentifier`

tonidero (approved):
❤️ Love this! I wish we had such test infrastructure on Android...

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
tonidero: Thinking about it more, it might be too much to try to cover all edge cases in integration tests, so I think it's ok to cover only the major cases like this.

NachoSoto: It doesn't hurt though, it's trivial. Just added it.

PR #238, opened 2022-09-26
Trigger purchases-flutter dependency upgrade when release is made

vegaro:
Remotely triggers a pipeline in purchases-flutter to upgrade the dependency in that repository.

Other repositories will follow after this is merged


fastlane/Fastfile
```diff
@@ -234,6 +234,43 @@ lane :run_pod_install do |options|
...
+  req_options = {
+    use_ssl: uri.scheme == "https",
+  }
+
+  response = Net::HTTP.start(uri.hostname, uri.port, req_options) do |http|
+    http.request(request)
+  end
+
+  puts response.code
+  puts response.body
```
tonidero, line 269: Is this debugging code? If it isn't, we should make sure no hidden information shows up in the logs in CircleCI.

fastlane/Fastfile
```diff
@@ -234,6 +234,43 @@ lane :run_pod_install do |options|
...
+
+  UI.error("Error triggering the pipeline") unless response.code == "201"
```
tonidero, line 271: It would be great to also print what error happened, but that might be included if we print the code and body above.

fastlane/Fastfile
```diff
@@ -234,6 +234,43 @@ lane :run_pod_install do |options|
...
+  bump = current_version_number
+  repo_name = options[:repo_name],
+
+  UI.message("Triggering pipeline to bump #{repo_name} dependency to #{bump}")
+
+  require 'net/http'
+  require 'uri'
+  require 'json'
+
+  uri = URI.parse("https://circleci.com/api/v2/project/github/RevenueCat/purchases-flutter/pipeline")
```
tonidero, line 248: This is hardcoding `purchases-flutter`

fastlane/Fastfile
```diff
@@ -234,6 +234,43 @@ lane :run_pod_install do |options|
   end
 end
 
+desc "Run automatic bump in other repos"
+lane :bump_hybrid_dependencies do |options|
+  bump = current_version_number
```
tonidero, line 239: Maybe rename to `version_number_for_bump` or `bump_version_number`?

fastlane/Fastfile
```diff
@@ -234,6 +234,43 @@ lane :run_pod_install do |options|
...
+  repo_name = options[:repo_name],
+
+  UI.message("Triggering pipeline to bump #{repo_name} dependency to #{bump}")
+
+  require 'net/http'
+  require 'uri'
+  require 'json'
```
tonidero, line 246: Hmm not sure if it's best practice to import these at the top of the file... But then again, these are only used here so it seems better... I guess I'm ok with this.

.circleci/config.yml
```diff
@@ -245,6 +260,7 @@ workflows:
       - test-ios
       - integration-test-ios
       - test-android
+      - trigger-dependent-updates
```
tonidero, line 263: I guess this needs to be moved to the `deploy` workflow right? And make sure all steps finish correctly before triggering this.

fastlane/Fastfile
vegaro: Screen Shot 2022-09-27 at 1 14 41 PM

I left it to print the response from GitHub. Do you think we should remove it just in case?

fastlane/Fastfile
vegaro: I think it will be helpful to print the error code in the message and remove the response.code log

fastlane/Fastfile
vegaro: ohh yes good catch

.circleci/config.yml
vegaro: oh yes sorry, I wash in a rush to open the PR and forgot this here

fastlane/Fastfile
tonidero: Hmm I'm not a fan of printing the whole response, specially since our test logs are open... I guess I would say to only print the response on errors maybe? And if we can parse the error message, instead of printing the whole response that would be even better.

tonidero (approved):
Just that thing about printing the whole API response, other than that looks good!

PR #235, opened 2022-09-26
Upgrade iOS to 4.12.1 and Android to 5.6.1

RCGitBot:
Upgrade iOS to 4.12.1 and Android to 5.6.1

PR #241, opened 2022-09-27
[Automatic] Release/4.3.0

RCGitBot:
### Other Changes
* Upgrade iOS to 4.13.0 and Android to 5.6.1 (#235) via RevenueCat Git Bot (@RCGitBot)
* Remove step from automatic-release to trigger Flutter update. Use its own job instead (#240) via Cesar de la Vega (@vegaro)
* Trigger purchases-flutter dependency upgrade when release is made (#238) via Cesar de la Vega (@vegaro)
* Remove upload of the PurchasesHybridCommon.framework (#239) via Cesar de la Vega (@vegaro)
* Update Dangerfile to use repository (#237) via Cesar de la Vega (@vegaro)
* Store PurchasesHybridCommon.framework.zip artifact (#236) via Cesar de la Vega (@vegaro)


.version
```diff
@@ -1 +1 @@
-4.2.2
\ No newline at end of file
+4.2.3
```
NachoSoto, line 1: If one of the underlying dependencies (iOS and/or Android) had a minor version bump, should this be a minor version bump too? @tonidero @vegaro
Then if PHC gets a minor version bump, the hybrids that use it could too.

NachoSoto (approved):

NachoSoto:
This didn't get the `[Automatic]` in the PR title?

.version
tonidero: Hmm right... that would increase the amount of minors/majors in PHC and hybrids, since they would have to be increased for the corresponding minors and majors of iOS and Android, but I think that's probably desired behavior. Currently, that's not implemented though, it would only look at the labels of the PRs since the last release. I think for now, if we want to make this a minor, we can tag the PR with the iOS bump with `feat`, close this PR and rerun the job to create a minor. We would need to update the Changelog as well.

And about how to implement this in the future, we can think about it since I don't think it's trivial, specially the autogeneration of the CHANGELOG.

vegaro: That's an interesting one. I actually think if the public API doesn't change, it should still be a patch. Last time we increased BillingClient to version 4, we released a minor I think, not a major.

I actually found this comment https://github.com/semver/semver/issues/148#issuecomment-21970506 which pretty much summarizes what I think about it.

This is an interesting definition of the problem:

```
It isn't a bug fix/patch so 4.2.3 seems wrong.
It isn't an addition of functionality so 4.3.0 seems wrong.
It is still backwards compatible so 5.0.0 seems wrong.
```

vegaro:
> This didn't get the [Automatic] in the PR title?

Yeah... I am not sure why. I will check

vegaro:
@NachoSoto https://github.com/RevenueCat/purchases-hybrid-common/pull/242

.version
NachoSoto: Yeah it's tricky. In most cases I agree, but in this one for example we made SK2 the default. If we don't override that in some of the hybrids (not sure if all set it to false / true explicitly) then that's definitely not a patch...

vegaro: ah, in that case yes, it makes sense. Do you want me to take care of rerunning it as a 4.3.0?

NachoSoto: I can manually change it, I just wanted to hear what your opinions.
Being conservative and doing what I suggested in all cases might be best, even if in some cases a "new feature" doesn't affect the hybrids until it's been exposed. What do you think?

vegaro: Yes, I think it makes sense to be conservative. I think even for purchases-android being updated to Billing Client 5 we would want to release a new major for observer mode folks.

NachoSoto:
`deploy-android` failed?

NachoSoto:
Should we retry that job?

vegaro:
@NachoSoto let me take a look. Sonatype kept timing out the other day for some weird reason

vegaro:
I finished the release manually.

PR #243, opened 2022-09-28
Adds parameters to run bump and dependency updates manually instead of approval jobs

vegaro:
This way `automatic-dependency-update` and `automatic-release` jobs can be run from CircleCI UI by passing `dependency-update` or `bump` as parameters when triggering a pipeline

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -321,4 +321,4 @@ workflows:
     when:
       equal: [ bump, << pipeline.parameters.action >> ]
     jobs:
-      - revenuecat/automatic-bump
+      - automatic-release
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -321,18 +317,8 @@ workflows:
...
-          type: approval
-          filters:
-            tags:
-              ignore: /.*/
-            branches:
-              only: main
-      - automatic-release:
-          requires:
-            - trigger-automatic-release
+      - automatic-release
```
tonidero, line 324: It's a bit strange that a "manual" workflow to trigger an "automatic" job... Wonder if this should just be `release`... Same with the dependency update job. NABD though.

tonidero (approved):

PR #245, opened 2022-09-29
Update maven-publish-plugin

vegaro:
Deploying Android kept timing out. I increased the version of our publishing plugin since I saw in the latest version the timeout has been increased. 

Other changes are:
- `closeAndReleaseRepository` is ran automatically now
- `SONATYPE_HOST` needs to be specified
- AGP plugin had to be increased for this new version to work so I updated it to the latest version
- package name in the AndroidManifest.xml is now deprecated and should be declared in the build.gradle 
- `mavenPublish` has now been deprecated. The Android variant to publish can be defined using a gradle property now

Their changelog is here https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/CHANGELOG.md#version-0220-2022-09-09

[CSDK-480]

[CSDK-480]: https://revenuecats.atlassian.net/browse/CSDK-480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -324,7 +324,7 @@ platform :android do
     )
 
     UI.verbose("Creating special version for Unity IAP with BillingClient 3 and Amazon 2 for version: #{version}")
-    gradleProperties["PUBLISH_VARIANT"] = "unityIAPRelease"
+    gradleProperties["ANDROID_VARIANT_TO_PUBLISH"] = "unityIAPRelease"
 
     UI.verbose("Adding -unityiap to artifact ids")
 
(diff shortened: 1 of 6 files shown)


fastlane/Fastfile
```diff
@@ -335,16 +335,6 @@ platform :android do
       properties: gradleProperties,
       project_dir: 'android'
     )
-
-    unless is_snapshot_version?(version)
```
tonidero, line 339: Hmm for PHC it's not an issue since we don't have snapshots but once we update the plugin in `purchases-android` we will have to think about how to avoid releasing snapshots unnecessarily.

tonidero (approved):
Thanks for updating these! I think we would want to also update it in `purchases-android` if this improves the issue with the releases timeout. Do we have a ticket for it?

fastlane/Fastfile
vegaro: I thought doing `closeAndReleaseRepository` on a SNAPSHOT didn't have any effect and this check was just to save time, but maybe we should confirm it haha

vegaro:
@tonidero https://revenuecats.atlassian.net/browse/CSDK-485

PR #248, opened 2022-10-06
Adds missing repositories to trigger-dependent-updates workflow

vegaro:
Now that we see that Flutter is working, this PR adds the rest of the repositories to notify after making a new release

Some repos don't have the workflow yet, and they will fail, but that's ok


.circleci/config.yml
```diff
@@ -246,6 +246,15 @@ jobs:
...
           command: bundle exec fastlane bump_hybrid_dependencies repo_name:purchases-flutter
+      - run:
+          name: Kick off React Native automatic dependency update
+          command: bundle exec fastlane bump_hybrid_dependencies repo_name:react-native-purchases
+      - run:
+          name: Kick off Cordova automatic dependency update
+          command: bundle exec fastlane bump_hybrid_dependencies repo_name:cordova-plugin-purchases
+      - run:
+          name: Kick off Unity automatic dependency update
+          command: bundle exec fastlane bump_hybrid_dependencies repo_name:purchases-unity
```
tonidero, line 257: If one of these fails, it won't execute the next one right? I guess we should follow this order to add the required workflows in those repos.

tonidero (approved):

.circleci/config.yml
vegaro: Yeah, it won't execute the next one... Maybe there's a flag to prevent that? Btw, all the workflows are already in the hybrid repositories. I actually called executed these fastlane lanes from my machine yesterday and triggered the updates successfully

PR #249, opened 2022-10-06
Improvements on automatic upgrading iOS and Android version depending type of change

vegaro:
Good amount of improvements here:

- It will only indicate in the branch name and in the PR title which platform is being updated. This means that it won't show anymore both platforms version, even if one of them is not being updated. For example, in this PR https://github.com/RevenueCat/purchases-hybrid-common/pull/246/files, the title indicated Android was being updated, but the only platform being updated was actually iOS
- If the update is a minor change in version, it will add the label `minor`. For example, if going from 4.12.0 to 4.13.1, the `minor` label will be added. This is to force a `minor` update on dependent SDKs in future PRs.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -96,9 +96,6 @@ lane :open_pr_upgrading_dependencies do |options|
   type_of_bump_ios = detect_bump_type(current_ios_version, new_ios_version)
   type_of_bump_android = detect_bump_type(current_android_version, new_android_version)
 
-  ios_version_changed = current_ios_version != new_ios_version
-  android_version_changed = current_android_version != new_android_version
-
   repo_status = sh("git status --porcelain")
 
   repo_clean = repo_status.empty?
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -86,7 +86,18 @@ lane :open_pr_upgrading_dependencies do |options|
 
   dry_run = options[:dry_run]
 
+  current_ios_version = parse_previous_ios_native_version
+  current_android_version = parse_previous_android_native_version
+
   new_versions = update_native_dependencies_to_latest
+  new_ios_version = new_versions[:new_ios_version]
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -120,14 +147,22 @@ lane :open_pr_upgrading_dependencies do |options|
       set_upstream: true,
       remote_branch: branch
     )
+
+    message = "[AUTOMATIC] #{message_ios}#{message_android}".strip
+
+    labels = ["dependencies"]
+    if type_of_bump_ios == :minor || type_of_bump_android == :minor
```
tonidero, line 154: Hmm just to check, if it's a major are we still creating the automatic PRs? And if so, should we also add a `major` label here?

fastlane/Fastfile
```diff
@@ -86,7 +86,18 @@ lane :open_pr_upgrading_dependencies do |options|
...
+
   new_versions = update_native_dependencies_to_latest
+  new_ios_version = new_versions[:new_ios_version]
+  new_android_version = new_versions[:new_android_version]
+
+  type_of_bump_ios = detect_bump_type(current_ios_version, new_ios_version)
+  type_of_bump_android = detect_bump_type(current_android_version, new_android_version)
+
+  ios_version_changed = current_ios_version != new_ios_version
+  android_version_changed = current_android_version != new_android_version
```
tonidero, line 100: These seem unused

tonidero (approved):
Nice improvements here!

fastlane/Fastfile
vegaro: We are skipping if it's not within the same major, that's why I didn't bother checking here. We do that in `update_native_dependencies_to_latest`, where we only get the latest release within the same major

fastlane/Fastfile
vegaro: oh yes, this is from a bad rebase

Issue #252, opened 2022-10-21
Unexpectedly found nil while implicitly unwrapping an Optional value


SteveMB1:
## Common project specific issues

There are certain project specific issues that are commonly misinterpreted as bugs.

- [Offerings, products, or available packages are empty](https://support.revenuecat.com/hc/en-us/articles/360041793174)
- [Invalid Play Store credentials errors](https://support.revenuecat.com/hc/en-us/articles/360046398913)
- [Unable to connect to the App Store (STORE_PROBLEM) errors](https://support.revenuecat.com/hc/en-us/articles/360046399333)

For support I'd recommend our [online community](https://spectrum.chat/revenuecat), [StackOverflow](https://stackoverflow.com/tags/revenuecat/) and/or [Help Center](https://support.revenuecat.com/hc/en-us) 👍

If you have a clearly defined bug (with a [Minimal, Complete, and Reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)) that is not specific to your project, follow the steps below to file it with RevenueCat. For SDK-related bugs, make sure they can be reproduced on a physical device, not a simulator (there are simulator-specific problems that prevent purchases from working).

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue.

1. Environment
    1. Platform: `Flutter`
    2. SDK version: `4.3.0`
 
`Flutter 3.5.0-10.0.pre.17 • channel master • https://github.com/flutter/flutter.git
Framework • revision 59d26c481a (5 hours ago) • 2022-10-20 20:05:08 -0400
Engine • revision f24ea1a04d
Tools • Dart 2.19.0 (build 2.19.0-324.0.dev) • DevTools 2.18.0`
 
   5. OS version: `macOS 12.6`
    6. XCode/Android Studio version: `XCode`
    7. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

The emulator / iPhone 14 Pro crashes after running this. `Purchases.purchaseProduct('GTM1');` 

`2022-10-21 00:42:26.558652-0400 Runner[91897:4576069] flutter: time since last location update:0
[...]

PR #253, opened 2022-10-21
re-add error message if SDK hasn't been configured

aboedo:
Added a wrapper around `sharedInstance` so we can check whether the SDK has been configured and error out if it hasn't.


NachoSoto (approved):
👏🏻

vegaro (approved):

vegaro:
I think this fixes https://github.com/RevenueCat/purchases-hybrid-common/issues/252 too right?

PR #278, opened 2022-12-16
[AUTOMATIC] iOS 4.15.3 => 4.15.4

RCGitBot:
[AUTOMATIC] iOS 4.15.3 => 4.15.4

PR #279, opened 2022-12-19
`Integration Tests`: enabled receipt fetch retry mechanism

NachoSoto:
See https://github.com/RevenueCat/purchases-ios/pull/1945.
Hopefully this will fail the tests in https://github.com/RevenueCat/purchases-hybrid-common/pull/278.


vegaro (approved):
Makes sense. Do you know why it started failing?

NachoSoto:
@vegaro no idea why now, or why it hadn't been failing here, but this was essential in the iOS repo to get tests to stop failing.
I think this further points to the fact that we shouldn’t be using SK2 to purchase and SK1 to get the receipt in the backend, which leads to this race condition on Apple's side.

PR #285, opened 2022-12-28
Add StoreProduct mapper iOS tests

vegaro:
We were missing tests for mapping the StoreProduct

PR #284, opened 2022-12-28
Add StoreProduct Android Tests 

vegaro:
We were missing tests for the StoreProduct mapper

PR #286, opened 2022-12-28
Adds subscriptionPeriod

vegaro:
Tests depend on #284 and #285 

[CSDK-614]

[CSDK-614]: https://revenuecats.atlassian.net/browse/CSDK-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -14,7 +14,8 @@ fun StoreProduct.map(): Map<String, Any?> =
         "introPrice" to mapIntroPrice(),
         "discounts" to null,
         "productCategory" to mapProductCategory(),
-        "productType" to mapProductType()
+        "productType" to mapProductType(),
+        "subscriptionPeriod" to subscriptionPeriod
```
tonidero, line 18: I think we don't have many tests on these mappers but I wonder if we should add them... Not a blocker for this PR though I guess.

tonidero (approved):

android/src/test/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapperTests.kt
```diff
@@ -105,6 +105,19 @@ internal class StoreProductMapperTest {
...
+    @Test
+    fun `maps subscription period correctly`() {
+        stubStoreProduct().map().let {
+            assertThat(it["subscriptionPeriod"]).isNull()
+        }
+        stubStoreProduct(subscriptionPeriod = "P1M").map().let {
+            assertThat(it["subscriptionPeriod"]).isEqualTo("P1M")
+        }
+        stubStoreProduct(subscriptionPeriod = "P1Y").map().let {
+            assertThat(it["subscriptionPeriod"]).isEqualTo("P1Y")
```
tonidero, line 118: Nice thanks! It could also be interesting to add a test that checks the number of keys on the map so we detect if something changed.

tonidero (approved):
A suggestion but looks good!

PR #290, opened 2023-01-02
Add `beginRefundRequest` APIs in iOS

tonidero:
PHC PR for https://revenuecats.atlassian.net/browse/CSDK-94

This PR adds 3 new methods to the iOS API related to `beginRefundRequest`. Users can initiate that request using a `productId`, an `entitlementId` or just refund the current active entitlement. This API doesn't apply to Android so it hasn't been added there.

### DIFF
--- a/ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
+++ b/ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
@@ -93,6 +93,17 @@ NS_ASSUME_NONNULL_BEGIN
     }
     [RCCommonFunctionality invalidateCustomerInfoCache];
     BOOL canMakePayments __unused = [RCCommonFunctionality canMakePaymentsWithFeatures:@[]];
+
+    if (@available(iOS 15.0, *)) {
+        [RCCommonFunctionality beginRefundRequestProductId:@""
+                                           completionBlock:^(RCErrorContainer * _Nullable error) {
+        }];
+        [RCCommonFunctionality beginRefundRequestEntitlementId:@""
+                                               completionBlock:^(RCErrorContainer * _Nullable error) {
+        }];
+        [RCCommonFunctionality beginRefundRequestForActiveEntitlementCompletion:^(RCErrorContainer * _Nullable error) {
+        }];
+    }
 }
 
 - (void)testDeprecatedAPI {
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -13,7 +13,9 @@ import RevenueCat
 
 @objc(RCCommonFunctionality) public class CommonFunctionality: NSObject {
 
-    static var sharedInstance: PurchasesType {
+    typealias InstanceType = PurchasesType & PurchasesSwiftType
+
+    static var sharedInstance: InstanceType {
         get {
             guard let purchases = Self._sharedInstance else {
                 fatalError("Purchases has not been configured. Please configure the SDK before calling this method")
@@ -26,7 +28,7 @@ import RevenueCat
         }
     }
 
-    private static var _sharedInstance: PurchasesType?
+    private static var _sharedInstance: InstanceType?
 
     // MARK: properties and configuration
 
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -120,6 +120,7 @@ import RevenueCat
 
 // MARK: Refund request
 @objc public extension CommonFunctionality {
+
 #if os(iOS)
     @available(iOS 15.0, *)
     @available(tvOS, unavailable)
@@ -129,11 +130,7 @@ import RevenueCat
     @objc(beginRefundRequestProductId:completionBlock:)
     static func beginRefundRequest(productId: String,
                                    completion: @escaping (ErrorContainer?) -> Void) {
-        guard let purchases = Self.sharedInstance as? PurchasesSwiftType else {
-            completion(Self.refundRequestError(description: "Invalid Purchases type."))
-            return
-        }
-        purchases.beginRefundRequest(forProduct: productId) { result in
+        Self.sharedInstance.beginRefundRequest(forProduct: productId) { result in
             Self.processRefundRequestResultWithCompletion(refundRequestResult: result, completion: completion)
         }
     }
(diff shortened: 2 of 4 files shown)


ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
```diff
@@ -93,6 +93,17 @@ - (void)testAPI {
...
+    if (@available(iOS 15.0, *)) {
+        [RCCommonFunctionality beginRefundRequestProductId:@""
+                                           completionBlock:^(RCErrorContainer * _Nullable error) {
+        }];
+        [RCCommonFunctionality beginRefundRequestEntitlementId:@""
+                                               completionBlock:^(RCErrorContainer * _Nullable error) {
+        }];
+        [RCCommonFunctionality beginRefundRequestForActiveEntitlementCompletion:^(RCErrorContainer * _Nullable error) {
+        }];
+    }
```
tonidero, line 106: I don't think we have API tests for swift. Something to consider in the future.

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -114,7 +114,61 @@ import RevenueCat
...
+#if os(iOS)
+    @available(iOS 15.0, *)
+    @available(tvOS, unavailable)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(macCatalyst, unavailable)
+    @objc(beginRefundRequestProductId:completionBlock:)
+    static func beginRefundRequest(productId: String,
+                                   completion: @escaping (ErrorContainer?) -> Void) {
+        guard let purchases = Self.sharedInstance as? PurchasesSwiftType else {
```
tonidero, line 130: `PurchasesSwiftType` already had completion block methods for these in `purchases-ios` but we are using the `PurchasesType` protocol in PHC. I'm guessing the separation between `PurchasesType` and `PurchasesSwiftType` was made so people could use purchases from ObjC without any issues. I'm not sure if there is a better way to access those methods though, and I thought we should use them rather than write another implementation in PHC. Lmk if you have other ideas though!

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -114,7 +114,61 @@ import RevenueCat
...
+@objc public extension CommonFunctionality {
+#if os(iOS)
+    @available(iOS 15.0, *)
+    @available(tvOS, unavailable)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(macCatalyst, unavailable)
+    @objc(beginRefundRequestProductId:completionBlock:)
+    static func beginRefundRequest(productId: String,
+                                   completion: @escaping (ErrorContainer?) -> Void) {
```
tonidero, line 129: In iOS, the result is an enum with 3 values `valid`, `userCancelled` and `error`. So to keep it simple, here I'm just returning an optional `ErrorContainer` which would indicate if it was `userCancelled` or another error, and if it's missing, it would mean it was successful. Alternatively, I could map it to another enum from PHC which can be accessed by the hybrids. Thoughts on that?

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -509,4 +563,35 @@ private extension CommonFunctionality {
...
+            }
+        case let .failure(error):
+            completion(ErrorContainer(error: error, extraPayload: [:]))
+        }
+    }
+
+    static func refundRequestError(description: String, userCancelled: Bool? = nil) -> ErrorContainer {
+        var extraPayload: [String: Any] = [:]
+        if let userCancelled = userCancelled {
+            extraPayload["userCancelled"] = userCancelled
```
tonidero, line 588: I copied this from another piece of code in this class. I wasn't sure whether we should always set this property, even when it's false, but I thought we can leave this for now.

ios/PurchasesHybridCommon/PurchasesHybridCommon.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,16 +1,14 @@
...
+      "state" : {
+        "revision" : "f29e2014f6230cf7d5138fc899da51c7f513d467",
+        "version" : "1.10.0"
       }
-    ]
-  },
-  "version": 1
+    }
+  ],
+  "version" : 2
```
NachoSoto: This breaks compatibility with Xcode 13 ⚠️  you'll need to revert this file.

tonidero: Yup just did 👍

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -114,7 +114,61 @@ import RevenueCat
         // See https://sdk.revenuecat.com/android/5.1.1/purchases/com.revenuecat.purchases/-purchases/-companion/can-make-payments.html
         return Purchases.canMakePayments()
     }
+}
+
+// MARK: Refund request
+@objc public extension CommonFunctionality {
+#if os(iOS)
```
NachoSoto, line 121: Nit: space around declarations (see https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/SwiftStyleGuide.swift#L20)

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
```diff
@@ -197,5 +197,221 @@ class PurchasesHybridCommonTests: QuickSpec {
...
+                        CommonFunctionality.sharedInstance = mockPurchases
+
+                        var completionCallCount = 0
+                        var completionError: ErrorContainer? = nil
+
+                        CommonFunctionality.beginRefundRequest(productId: "mock-product-id") { error in
+                            completionCallCount += 1
+                            completionError = error
+                        }
+                        mockPurchases.invokedBeginRefundRequestForProductParameters?.1(Result(.userCancelled, nil))
```
NachoSoto, line 245: This is equivalent to
```suggestion
                        mockPurchases.invokedBeginRefundRequestForProductParameters?.1(.success(.userCancelled))
```

I wouldn't use the `Result(value, error)` overload just to set it to a success.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
```diff
@@ -197,5 +197,221 @@ class PurchasesHybridCommonTests: QuickSpec {
...
+                        CommonFunctionality.sharedInstance = mockPurchases
+
+                        var completionCallCount = 0
+                        var completionError: ErrorContainer? = nil
+
+                        CommonFunctionality.beginRefundRequest(productId: "mock-product-id") { error in
+                            completionCallCount += 1
+                            completionError = error
+                        }
+                        mockPurchases.invokedBeginRefundRequestForProductParameters?.1(Result(.success, nil))
```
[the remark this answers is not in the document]
tonidero: Hmm these tests are making sure that when the `Purchases.beginRefundRequest...` calls return `success`, PHC maps that correctly to calling the PHC completion block with a nil error. So this test does go through the real implementation of the mapping of this function. 

This specific line is there to call the completion block for the `purchases-ios` function call with the value we want (note that we don't have access to that completion block directly from the test but it's stored in the mock). Alternatively, we could setup stubbed return values for this function beforehand and always call the completion block with the stubbed value, but this is the format we have for the other mock methods in `MockPurchases`.

Do lmk if you meant something else!

NachoSoto (approved):

PR #294, opened 2023-01-05
`CommonFunctionality.sharedInstance`: changed type to `PurchasesType & PurchasesSwiftType`

NachoSoto:
This will help simplify #290.


tonidero (approved):
Ah nice! Makes sense, thanks!

PR #307, opened 2023-01-26
Adds setLogHandler

vegaro:
Adde setLogHandler as part of [CSDK-548](https://revenuecats.atlassian.net/browse/CSDK-548)

[CSDK-548]: https://revenuecats.atlassian.net/browse/CSDK-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -90,6 +90,34 @@ import RevenueCat
...
+            case .debug:
+                logLevelName = "debug"
+                break
+            case .verbose:
+                logLevelName = "verbose"
+                break
+            case .warn:
+                logLevelName = "warn"
+                break
+            }
```
vegaro, line 112: Not sure it's worth it to extract this into an extension or something like that? What do you think

vegaro:
forgot to add tests, adding now

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
NachoSoto: You could just use `logLevel.description`

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -90,6 +90,34 @@ import RevenueCat
         Purchases.logLevel = level
     }
 
+    @objc public static func setLogHander(completion: @escaping ([String: String]) -> Void) {
```
NachoSoto, line 93: Maybe document the format of the dictionary?

vegaro:
> You could just use logLevel.description
Oh good point. I thought about it but then I worried about making a change in the iOS side and not in Android and breaking this mapping. But I guess if we make a breaking change in purchases-ios to the loglevels we will also have to update this mapping, so I think we are good

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
NachoSoto: Actually why a dictionary instead of 2 parameters?

vegaro: I want it to be called by all hybrids like this https://github.com/RevenueCat/purchases-flutter/pull/591/files#diff-a84fbe19c95708058ab36b565eb31eb4a3071f224431bdb07085001b0f6271bbR546-R552

Otherwise I would have to do the mapping to a json in every single hybrid, this way it's done in phc and we don't duplicate code. Similar to what we do with other objects, like CustomerInfo

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/LogHandlerWithMapping.kt
```diff
@@ -0,0 +1,30 @@
+package com.revenuecat.purchases.hybridcommon.mappers
+
+class LogHandlerWithMapping(
+    private val callback: (logData: Map<String, String>) -> Unit
+) : com.revenuecat.purchases.LogHandler {
```
tonidero, line 5: Could we add an import instead of using the full name?

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/LogHandlerWithMapping.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun v(tag: String, msg: String) {
+        invokeCallback("verbose", msg)
+    }
+
+    override fun w(tag: String, msg: String) {
+        invokeCallback("warn", msg)
+    }
+
+    private fun invokeCallback(c: String, msg: String) {
+        callback(mapOf("logLevel" to "debug", "message" to msg))
```
tonidero, line 28: I think you meant this:
```suggestion
        callback(mapOf("logLevel" to c, "message" to msg))
```

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/LogHandlerWithMapping.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    override fun i(tag: String, msg: String) {
+        invokeCallback("info", msg)
+    }
+
+    override fun v(tag: String, msg: String) {
+        invokeCallback("verbose", msg)
+    }
+
+    override fun w(tag: String, msg: String) {
```
tonidero, line 23: We are ignoring `tag` in all this... I wonder if we should embed it into the message...

vegaro: I didn't because iOS doesn't have one, so it would be weird Android logs have a [PURCHASES] in front. But if you think it's worth it to add it I will

tonidero: Hmm not a big deal either way I would say. I think it's fine to leave as is.

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/LogHandlerWithMapping.kt
vegaro: Oh yes, I had named this class LogHandler and this is a leftover

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
```diff
@@ -361,5 +361,20 @@ class PurchasesHybridCommonTests: QuickSpec {
                 }
             }
         }
+
+        context("setLogHandler") {
+            let expectedMessage = "a message"
+            for level in LogLevel.levels {
+                it("\(level) logs work") {
+                    let originalHandler = Purchases.verboseLogHandler
```
vegaro, line 369: Ok so I noticed that doing this:

```
let originalHandler = Purchases.logHandler
Purchases.logHandler = originalHandler
```

Makes the code loop forever until it crashes. The getter creates a new function (function `a`) that calls the `verboseLogHandler`. Then we set `logHandler` to that new function `a`. But the setter of `logHandler` actually sets `verboseLogHandler` to a new function that calls function `a`. The problem is that `function a` was set to call `verboseLogHandler`. So when calling `verboseLogHandler` we will call `function a`, which will call `verboseLogHandler` again, and a loop is created

vegaro: So I decided to just set the `verboseLogHandler` instead.

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -90,6 +90,21 @@ import RevenueCat
...
+     *
+     * - Parameter completion: Gets a map with two keys, a `logLevel` which  is one of the ``LogLevel``  name lowercased, and a `message`, with the log message.
+     */
+    @objc public static func setLogHander(completion: @escaping ([String: String]) -> Void) {
+        Purchases.logHandler = { logLevel, message in
+            let logDetails = [
+                "logLevel": logLevel.description.uppercased(),
+                "message": message
+            ]
+            completion(logDetails)
```
tonidero, line 104: Hmm I think we should rename the lambda since it's not really a completion but a `callback` or an `onLogReceived` or something like that.

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -232,6 +232,16 @@ fun setLogHandler(
     Purchases.logHandler = logHandler
 }
 
+/**
+ * Sets a log handler and forwards all logs to completion function.
+ *
+ * @param callback Gets a map with two keys, a `logLevel` which  is one of the ``LogLevel`` name lowercased,
+ * and a `message`, with the log message.
+ */
+fun setLogHandler(callback: (logDetails: Map<String, String>) -> Unit) {
```
tonidero, line 241: Hmm should we deprecate/remove the old `setLogHandler` method? I believe it only exists for Android.

tonidero (approved):
Nothing major, LGTM!

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
NachoSoto: Okay that makes sense 👍🏻 

~~Curious, why lowercase though? I've made the associated `String` be uppercase everywhere to match the one in iOS too. Should we use uppercase?~~

Never mind, I see it's already uppercase, but the comment is backwards :P

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
NachoSoto: 👍🏻

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
NachoSoto: Oh yeah that makes sense :/ that's a consequence of the fact that `logHandler` is derived from `verboseLogHandler`.

This implementation makes sense though 👍🏻 

To clean it up I would move it out of the test though:
```swift
context("setLogHandler") {
  var originalHandler: VerboseLogHandler!
  beforeEach {
    originalHandler = Purchases.verboseLogHandler
  }

  afterEach {
    Purchases.verboseLogHandler = originalHandler
  }

  it("...") {}
}
```

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
```diff
@@ -361,5 +361,20 @@ class PurchasesHybridCommonTests: QuickSpec {
...
+                    CommonFunctionality.setLogHander(completion: { logDetails in
+                        expect(logDetails["logLevel"]) == level.description.uppercased()
```
NachoSoto, line 371: I don't think you need `uppercased` here since it's already uppercase.

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -90,6 +90,21 @@ import RevenueCat
...
 
+    /**
+     * Sets a log handler and forwards all logs to completion function.
+     *
+     * - Parameter completion: Gets a map with two keys, a `logLevel` which  is one of the ``LogLevel``  name lowercased, and a `message`, with the log message.
+     */
+    @objc public static func setLogHander(completion: @escaping ([String: String]) -> Void) {
+        Purchases.logHandler = { logLevel, message in
+            let logDetails = [
+                "logLevel": logLevel.description.uppercased(),
```
NachoSoto, line 101: I don't think you need `uppercased` here.

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
vegaro: oh yes! I didn't notice

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
vegaro: I will remove it, I don't think it's used

ios/PurchasesHybridCommon/PurchasesHybridCommonTests/PurchasesHybridCommonTests.swift
vegaro: actually, I think it would be good to leave it just in case the description changes in the future to lowercase. I don't think it will but it would break the dart implementation if it does :)

NachoSoto: Ok that makes sense

PR #316, opened 2023-01-31
Fix for dependabot issue parsing Gemfile

vegaro:
This lines in the Gemfile were tripping dependabot. I changed it to be just the path of the Pluginfile.

Screenshot 2023-01-31 at 10 50 57 AM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,7 @@ source "https://rubygems.org"
 
 gem "fastlane"
 
-plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
-eval_gemfile(plugins_path) if File.exist?(plugins_path)
+eval_gemfile("fastlane/Pluginfile")
 
 gem "cocoapods", "~> 1.11"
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -2,8 +2,7 @@ source "https://rubygems.org"
 
 gem "fastlane"
 
-plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
-eval_gemfile(plugins_path) if File.exist?(plugins_path)
```
tonidero, line 6: Hmm this was the recommended way to add the Pluginfile by fastlane. I guess by hardcoding the path, it might not work from a different folder... But seems like dependabot doesn't work with this, so I think we can go with this for now and see if there are any issues.

tonidero (approved):

Gemfile
vegaro: I saw they did it in https://github.com/getsentry/sentry-cocoa/pull/982/files too which gave me confidence lol

PR #328, opened 2023-02-08
Add `setLogHandlerWithOnResult`

vegaro:
I am having issues when calling the lambda version of `setLogHandler` from Unity because it can't find Function1. I think it's related to purchases-unity not being configured for Kotlin. In order to be able to call `setLogHandler` I created a new version of it that uses the current `OnResult` function, so it can be called like this:

```
         CommonKt.setLogHandlerWithOnResult(new OnResult() {
            @Override
            public void onReceived(Map<String, ?> map) {
                sendJSONObject(MappersHelpersKt.convertToJson(map), HANDLE_LOG);
            }

            @Override
            public void onError(ErrorContainer errorContainer) {
                // intentionally left blank since it will never be called
            }
        });
```        

I couldn't make it an override of `setLogHandler` because I kept getting the same errors of not finding Function1, I don't know why. Possibly an issue with the Kotlin compiler:

Screenshot 2023-02-07 at 10 04 30 PM
Screenshot 2023-02-07 at 10 04 10 PM

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -5,7 +5,6 @@ import android.content.Context
 import com.revenuecat.purchases.BillingFeature
 import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.DangerousSettings
-import com.revenuecat.purchases.LogHandler
 import com.revenuecat.purchases.LogLevel
 import com.revenuecat.purchases.Purchases
 import com.revenuecat.purchases.PurchasesConfiguration
@@ -234,6 +233,21 @@ fun setLogLevel(level: String) {
  */
 fun setLogHandler(callback: (logDetails: Map<String, String>) -> Unit) {
     Purchases.logHandler = LogHandlerWithMapping(callback)
+}
+
+/**
+ * Sets a log handler and forwards all logs to completion function. Accepts an OnResult so it can be used from
+ * SDKs that don't have Kotlin configured, since they would error because Kotlin's Function1 would not be found.
+ * Function1 is what Kotlin lambdas are converted to. It has a different name than setLogHandler because naming it
+ * the same also gives errors due of missing Function1.
+ *
+ * @param onResult Gets a map with two keys, a `logLevel` which  is one of the ``LogLevel`` name uppercased,
+ * and a `message`, with the log message. The onError of OnResult will never be called.
+ */
+fun setLogHandlerWithOnResult(onResult: OnResult) {
+    setLogHandler { logDetails ->
+        onResult.onReceived(logDetails)
+    }
 }
 
 fun setProxyURLString(proxyURLString: String?) {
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -236,6 +235,21 @@ fun setLogHandler(callback: (logDetails: Map<String, String>) -> Unit) {
...
+/**
+ * Sets a log handler and forwards all logs to completion function. Accepts an OnResult so it can be used from
+ * SDKs that don't have Kotlin configured, since they would error because Kotlin's Function1 would not be found.
+ * Function1 is what Kotlin lambdas are converted to. It has a different name than setLogHandler because naming it
+ * the same also gives errors due of missing Function1.
+ *
+ * @param onResult Gets a map with two keys, a `logLevel` which  is one of the ``LogLevel`` name uppercased,
+ * and a `message`, with the log message. The onError of OnResult will never be called.
+ */
+fun setLogHandlerWithOnResult(onResult: OnResult) {
```
vegaro, line 247: Open to change the name of the function and also open to create a new functional interface instead of using `OnResult`, although I think since this is internal we can just reuse `OnResult`

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -236,6 +235,21 @@ fun setLogHandler(callback: (logDetails: Map<String, String>) -> Unit) {
     Purchases.logHandler = LogHandlerWithMapping(callback)
 }
 
+/**
+ * Sets a log handler and forwards all logs to completion function. Accepts an OnResult so it can be used from
+ * SDKs that don't have Kotlin configured, since they would error because Kotlin's Function1 would not be found.
```
tonidero, line 240: I guess enabling kotlin in Unity will be difficult right? 

So moving forward we should always try to use this Java interface in PHC instead of using lambdas correct? To avoid having duplicate methods.

tonidero (approved):
Just some questions but looks good

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
vegaro: Yes. I am actually not even sure how to enable it since we don't really have a build.gradle, I don't think it's straightforward. This was my bad I shouldn't have created a lambda parameter first place but didn't want to break the other hybrids. I guess lesson learned is do not make a lambda parameter in phc until we migrate everything to Kotlin

vegaro: Actually enabling Kotlin would mean adding a Kotlin file instead or Java. Maybe we can create a dummy Kotlin file and that would take care of it? I'll try that because I would prefer to use the regular lambda version

vegaro:
@tonidero I tried adding Kotlin to the Unity plugin, can you check this comment https://github.com/RevenueCat/purchases-unity/pull/237/files#r1100835371

android/src/test/java/apitests/kotlin/com/revenuecat/purchases/hybridcommon/CommonApiTests.kt
```diff
@@ -86,6 +85,13 @@ private class CommonApiTests {
         setLogHandler(callback)
     }
 
+    fun checkSetLogHandlerWithOnResult() {
+        setLogHandlerWithOnResult(object : OnResult {
+            override fun onReceived(map: Map<String?, *>?) {}
```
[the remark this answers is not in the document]
vegaro: I don't think we can annotate the type of the key 😕

PR #338, opened 2023-02-22
Passes error instead of two nils

vegaro:
While working on https://github.com/RevenueCat/purchases-unity/pull/245 I noticed how we were passing a `nil` error in case of incompatibilities when getting promotional offers. I changed that to return an error for versions of iOS that are not compatible with discounts.


joshdholtz (approved):
Ohhh, good catch! 💪

tonidero (approved):
👍

PR #345, opened 2023-03-01
`EntitlementInfo`: fixed `unsubscribeDetectedAt` typo

### LINKED ISSUE #340: [Bug] Incorrect Attribute Name mapping for iOS
## Common project specific issues

There are certain project specific issues that are commonly misinterpreted as bugs.

- [Offerings, products, or available packages are empty](https://support.revenuecat.com/hc/en-us/articles/360041793174)
- [Invalid Play Store credentials errors](https://support.revenuecat.com/hc/en-us/articles/360046398913)
- [Unable to connect to the App Store (STORE_PROBLEM) errors](https://support.revenuecat.com/hc/en-us/articles/360046399333)

For support I'd recommend our [online community](https://spectrum.chat/revenuecat), [StackOverflow](https://stackoverflow.com/tags/revenuecat/) and/or [Help Center](https://support.revenuecat.com/hc/en-us) 👍

If you have a clearly defined bug (with a [Minimal, Complete, and Reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)) that is not specific to your project, follow the steps below to file it with RevenueCat. For SDK-related bugs, make sure they can be reproduced on a physical device, not a simulator (there are simulator-specific problems that prevent purchases from working).

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue.

1. Environment
    1. Platform: **iOS**
    2. SDK version: **latest and previous**
    3. OS version: **all**
    4. XCode/Android Studio version: **n/a**
    5. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue **n/a**
3. Steps to reproduce, with a description of expected vs. actual behavior

In the below location of the code base, there seems to be an inadvertent attribute name mapping error for iOS:

https://github.com/RevenueCat/purchases-hybrid-common/blob/main/ios/PurchasesHybridCommon/PurchasesHybridCommon/EntitlementInfo+HybridAdditions.swift#L29-L30

_Expected Behaviour_

Attribute names are: `unsubscribeDetectedAt` and `unsubscribeDetectedAtMillis`

_Actual Behavior_

The attribute names mapped to us past tense, i.e. `unsubscribed`, therefore they are mapped to `unsubscribedDetectedAt` and `unsubscribedDetectedAtMillis` which deviates from the published API contract (and the Android equivalent).

5. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

NachoSoto:
Fixes #340

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/EntitlementInfo+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/EntitlementInfo+HybridAdditions.swift
@@ -26,8 +26,8 @@ import RevenueCat
             "store": storeString,
             "productIdentifier": productIdentifier,
             "isSandbox": isSandbox,
-            "unsubscribedDetectedAt": unsubscribeDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
-            "unsubscribedDetectedAtMillis": unsubscribeDetectedAt?.rc_millisecondsSince1970AsDouble() ?? NSNull(),
+            "unsubscribeDetectedAt": unsubscribeDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
+            "unsubscribeDetectedAtMillis": unsubscribeDetectedAt?.rc_millisecondsSince1970AsDouble() ?? NSNull(),
             "billingIssueDetectedAt": billingIssueDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
             "billingIssueDetectedAtMillis": billingIssueDetectedAt?.rc_millisecondsSince1970AsDouble() ?? NSNull(),
             "ownershipType": ownershipTypeString
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/__Snapshots__/StoreKitIntegrationTests/SK1-testCanPurchasePackage.1.json
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/__Snapshots__/StoreKitIntegrationTests/SK1-testCanPurchasePackage.1.json
@@ -18,8 +18,8 @@
           "periodType" : "INTRO",
           "productIdentifier" : "com.revenuecat.purchases_hybrid_common.monthly_19.99_.1_week_intro",
           "store" : "APP_STORE",
-          "unsubscribedDetectedAt" : null,
-          "unsubscribedDetectedAtMillis" : null,
+          "unsubscribeDetectedAt" : null,
+          "unsubscribeDetectedAtMillis" : null,
           "willRenew" : true
         }
       },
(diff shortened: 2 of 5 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/EntitlementInfo+HybridAdditions.swift
```diff
@@ -26,8 +26,8 @@ import RevenueCat
             "store": storeString,
             "productIdentifier": productIdentifier,
             "isSandbox": isSandbox,
-            "unsubscribedDetectedAt": unsubscribeDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
-            "unsubscribedDetectedAtMillis": unsubscribeDetectedAt?.rc_millisecondsSince1970AsDouble() ?? NSNull(),
+            "unsubscribeDetectedAt": unsubscribeDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
+            "unsubscribeDetectedAtMillis": unsubscribeDetectedAt?.rc_millisecondsSince1970AsDouble() ?? NSNull(),
```
vegaro, line 30: This is a breaking change for React Native and Cordova, since these get transformed directly into Javascript objects. We have typescript definitions, but using typescript is not mandatory. Should we maybe return both this wrong version and the correct one?

What do you think?

tonidero: Oh so it wasn't working at all before?

tonidero (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/EntitlementInfo+HybridAdditions.swift
NachoSoto: Nope. That's what I meant by the need for end-to-end tests so we can catch this stuff.

vegaro: Our React Native and Cordova plugins just provide Typescript interfaces, we don't do any parsing. This will break anyone that's not using typescript and instead looked at the Javascript object directly, saw we were sending 'unsubscribed' and is accessing that key in the JavaScript object we send down.

It is not very probable anyone is doing that and probably safe to change, but wanted to point out this breaks whoever is accessing that key directly.

In the past, I've kept both versions for a while just in case

NachoSoto: > It is not very probable anyone is doing that and probably safe to change, but wanted to point out this breaks whoever is accessing that key directly.

Oooh I see what you're saying.
We could offer both, but given that those users are using a key that wasn't documented, maybe it's fine. What do you think?

vegaro: Sure, makes sense. Let's ship it!

NachoSoto: Thanks!

PR #351, opened 2023-03-06
[AUTOMATIC] Android 5.7.1 => 5.8.0

RCGitBot:
[AUTOMATIC] Android 5.7.1 => 5.8.0


NachoSoto:
@tonidero do you want to ship this? Should we add a flag to be able to enable diagnostics for Android first?

tonidero:
I think we can merge this but we shouldn't ship a release only with this, since the only change is diagnostics and can't be used for hybrids until we add support to enable diagnostics in those.

tonidero (approved):

PR #355, opened 2023-03-13
[BC5] Adds PREPAID_SUBSCRIPTION to mapProductType

joshdholtz:
## Motivation

[CF-1265](https://revenuecats.atlassian.net/browse/CF-1265)

## Description

`StoreProduct.mapProductType` only had a value of `AUTO_RENEWABLE_SUBSCRIPTION` which does not work if its a prepaid subscription. Even though we don't _really_ say we support prepaid subscriptions yet, it just made sense to add a support for this right away since minimal effort.

So... `PREPAID_SUBSCRIPTION` is now a value if its a `ProductType.SUBS` and `RecurrenceMode.NON_RECURRING`

I also fixed a bunch of tests I had commented out that I forgot to fix in the main `bc5-support` PR 🤷‍♂️ 

[CF-1265]: https://revenuecats.atlassian.net/browse/CF-1265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
@@ -3,6 +3,7 @@ package com.revenuecat.purchases.hybridcommon.mappers
 import com.revenuecat.purchases.ProductType
 import com.revenuecat.purchases.models.Period
 import com.revenuecat.purchases.models.PricingPhase
+import com.revenuecat.purchases.models.RecurrenceMode
 import com.revenuecat.purchases.models.StoreProduct
 
 val StoreProduct.priceAmountMicros: Long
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -53,7 +54,13 @@ internal fun StoreProduct.mapProductCategory(): String {
 internal fun StoreProduct.mapProductType(): String {
     return when (type) {
         ProductType.INAPP -> "CONSUMABLE"
-        ProductType.SUBS -> "AUTO_RENEWABLE_SUBSCRIPTION" // TODO: Add a new string here prepaid (check recurrence mode)
+        ProductType.SUBS -> {
+            if (defaultOption?.fullPricePhase?.recurrenceMode == RecurrenceMode.NON_RECURRING) {
```
beylmk, line 58: i'm not remembering, do we consume this product type anywhere? just wanna be sure we remember to check for prepaid wherever we consume this too...

beylmk (approved):
🌴

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
joshdholtz: It doesn't _look_ like it. I doesn't even look like its used in Flutter 🤷‍♂️

PR #356, opened 2023-03-13
[BC5] Add isPersonalizedPrice to purchaseProduct and purchasePackage

joshdholtz:
## Motivation

[CF-1266](https://revenuecats.atlassian.net/browse/CF-1266)

## Description

Adds `isPersonalizedPrice: Boolean?` argument onto `purchaseProduct` and `purchasePackage`

[CF-1266]: https://revenuecats.atlassian.net/browse/CF-1266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -58,6 +58,7 @@ fun purchaseProduct(
     oldSku: String?,
     prorationMode: GoogleProrationMode?,
     type: String,
+    isPersonalizedPrice: Boolean?,
     onResult: OnResult
 ) {
     if (activity != null) {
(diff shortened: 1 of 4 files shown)


NachoSoto (approved):
Looks good.

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -58,6 +58,7 @@ fun purchaseProduct(
     oldSku: String?,
     prorationMode: GoogleProrationMode?,
     type: String,
+    isPersonalizedPrice: Boolean?,
```
beylmk, line 61: i've toyed in my head with whether isPlayPricePersonalized or something that indicates this is google-specific would be better naming... and it feels even more important maybe in the hybrids. thoughts?

beylmk: might end up wanting to do oldGoogleProductId and googldProrationMode too... 

also, speaking of... looks like oldsku can be changed to oldProductId. ignore if you're doing that in the future..

beylmk (approved):
just looking for feedback on the naming, nothing blocking

PR #357, opened 2023-03-13
[BC5] Temporarily remove Unity IAP from tests and release until its supported

joshdholtz:
## Motivation

[CF-1274](https://revenuecats.atlassian.net/browse/CF-1274)

CircleCI failing trying to install `purchases-unityiap:6.0.0-beta.2`

## Description

- Removes `unityIAP` flavor from the `build.gradle`
- Remove kicking off a unity bump on release

[CF-1274]: https://revenuecats.atlassian.net/browse/CF-1274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -276,10 +276,6 @@ jobs:
           name: Kick off Cordova automatic dependency update
           command: bundle exec fastlane bump_hybrid_dependencies repo_name:cordova-plugin-purchases
           when: always
-      - run:
-          name: Kick off Unity automatic dependency update
-          command: bundle exec fastlane bump_hybrid_dependencies repo_name:purchases-unity
-          when: always
 
 workflows:
   version: 2
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,7 +9,6 @@ buildscript {
     ext.mockk_version = '1.12.3'
     ext.gradle_maven_publish = '0.22.0'
     ext.purchases_version = '6.0.0-beta.2'
-    ext.billing_unityiap_purchases_version = '3.0.3'
 
     repositories {
         google()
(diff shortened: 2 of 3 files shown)


android/build.gradle
```diff
@@ -69,9 +65,6 @@ android {
     }
     flavorDimensions "dependencyVersions"
     productFlavors {
-        unityIAP {
-            dimension "dependencyVersions"
-        }
         latestDependencies {
             dimension "dependencyVersions"
         }
```
tonidero, line 70: I guess since we only have one flavor with this, we could also remove this, but since we plan to bring it back, I think it's probably ok to leave it.

tonidero (commented):
This looks good but one change missing is removing deploying the unityIAP artifact from the fastfile.

.circleci/config.yml
```diff
@@ -276,10 +276,6 @@ jobs:
           name: Kick off Cordova automatic dependency update
           command: bundle exec fastlane bump_hybrid_dependencies repo_name:cordova-plugin-purchases
           when: always
-      - run:
-          name: Kick off Unity automatic dependency update
-          command: bundle exec fastlane bump_hybrid_dependencies repo_name:purchases-unity
-          when: always
```
tonidero, line 282: Within Unity, we release 2 versions of the SDK, one with the billing client and one without. We could potentially release the version with the billing client (the one different is the one without it, since it's meant for devs using Unity's billing client SDK instead of our own). But just not bumping unity is probably safer, until we fully support Unity 👍

android/build.gradle
```diff
@@ -9,7 +9,6 @@ buildscript {
     ext.mockk_version = '1.12.3'
     ext.gradle_maven_publish = '0.22.0'
     ext.purchases_version = '6.0.0-beta.2'
-    ext.billing_unityiap_purchases_version = '3.0.3'
```
tonidero, line 12: Hmm we were using BC3 for tests in PHC... It probably works since we don't really run any test that requires it, we just need it to make it compile. Once we bring this back, we probably should update this version though

tonidero (approved):
LGTM!

PR #359, opened 2023-03-14
[BC5] Add subscription options support

joshdholtz:
## Motivation

[CF-1221](https://revenuecats.atlassian.net/browse/CF-1221)
[CF-1227](https://revenuecats.atlassian.net/browse/CF-1267)

Hybrids should have the ability to view and purchase subscription options.

## Description

- Added `defaultOption` and `subscriptionOptions` to `StoreProduct` mapper
- Add new `purchaseSubscriptionOption()` function
  - Uses `productIdentifier` (sub) and `optionIdentifier` (base plan + optional offer) to find the `SubscriptionOption`

See https://github.com/RevenueCat/purchases-flutter/pull/641 on how this is used in Flutter

[CF-1221]: https://revenuecats.atlassian.net/browse/CF-1221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CF-1227]: https://revenuecats.atlassian.net/browse/CF-1227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -134,7 +134,6 @@ fun purchaseProduct(
     }
 }
 
-
 fun purchasePackage(
     activity: Activity?,
     packageIdentifier: String,
@@ -183,6 +182,73 @@ fun purchasePackage(
                     )
                 }
             }
+        )
+    } else {
+        onResult.onError(
+            PurchasesError(
+                PurchasesErrorCode.PurchaseInvalidError,
+                "There is no current Activity"
+            ).map()
+        )
+    }
+}
+
+fun purchaseSubscriptionOption(
+    activity: Activity?,
+    productIdentifier: String,
+    optionIdentifier: String,
+    googleOldProductId: String?,
+    googleProrationMode: GoogleProrationMode?,
+    googleIsPersonalizedPrice: Boolean?,
+    onResult: OnResult
+) {
+    if (activity != null) {
+        val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
+            val optionToBuy = storeProducts.mapNotNull { storeProduct ->
+                storeProduct.subscriptionOptions?.map { Pair(storeProduct, it) }
+            }.flatten().firstOrNull { (storeProduct, subscriptionOption) ->
+                // TODO: Verify this works because "subId:basePlanId" (it should but strings are silly)
+                storeProduct.purchasingData.productId == productIdentifier && subscriptionOption.id == optionIdentifier
+            }?.second
+
+            if (optionToBuy != null) {
+                val purchaseParams = PurchaseParams.Builder(optionToBuy, activity)
+
+                // Product upgrade
+                if (googleOldProductId != null && googleOldProductId.isNotBlank()) {
+                    purchaseParams.oldProductId(googleOldProductId)
+                    if (googleProrationMode != null) {
+                        purchaseParams.googleProrationMode(googleProrationMode)
+                    }
+                }
+
+                // Personalized price
+                googleIsPersonalizedPrice?.let {
+                    purchaseParams.isPersonalizedPrice(googleIsPersonalizedPrice)
+                }
+
+                // Perform purchase
+                Purchases.sharedInstance.purchaseWith(
+                    purchaseParams.build(),
+                    onError = getPurchaseErrorFunction(onResult),
+                    onSuccess = getPurchaseCompletedFunction(onResult)
+                )
+            } else {
+                onResult.onError(
+                    PurchasesError(
+                        PurchasesErrorCode.ProductNotAvailableForPurchaseError,
+                        "Couldn't find product."
+                    ).map()
+                )
+            }
+
+        }
+
+        Purchases.sharedInstance.getProductsWith(
+            listOf(productIdentifier),
+            ProductType.SUBS,
+            { onResult.onError(it.map()) },
+            onReceiveStoreProducts
         )
     } else {
         onResult.onError(
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -78,7 +78,7 @@ fun purchaseProduct(
                 it.id == productIdentifier && it.type.name.equals(type, ignoreCase = true)
             }
             if (productToBuy != null) {
-                val purchaseParams = PurchaseParams.Builder(productToBuy, activity)
+                val purchaseParams = PurchaseParams.Builder(activity, productToBuy)
 
                 // Product upgrade
                 if (googleOldProductId != null && googleOldProductId.isNotBlank()) {
@@ -103,7 +103,7 @@ fun purchaseProduct(
                 onResult.onError(
                     PurchasesError(
                         PurchasesErrorCode.ProductNotAvailableForPurchaseError,
-                        "Couldn't find product."
+                        "Couldn't find product $productIdentifier"
                     ).map()
                 )
             }
@@ -152,7 +152,7 @@ fun purchasePackage(
                         it.identifier.equals(packageIdentifier, ignoreCase = true)
                     }
                 if (packageToBuy != null) {
-                    val purchaseParams = PurchaseParams.Builder(packageToBuy, activity)
+                    val purchaseParams = PurchaseParams.Builder(activity, packageToBuy)
 
                     // Product upgrade
                     if (googleOldProductId != null && googleOldProductId.isNotBlank()) {
@@ -177,7 +177,7 @@ fun purchasePackage(
                     onResult.onError(
                         PurchasesError(
                             PurchasesErrorCode.ProductNotAvailableForPurchaseError,
-                            "Couldn't find product."
+                            "Couldn't find product for package $packageIdentifier"
                         ).map()
                     )
                 }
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -36,12 +38,13 @@ fun StoreProduct.map(): Map<String, Any?> =
...
         "currencyCode" to priceCurrencyCode,
-        // TOOD: BC5 allows up to 2 intro phases (free and then discount), should this only show the first?
         "introPrice" to mapIntroPrice(),
         "discounts" to null,
         "productCategory" to mapProductCategory(),
         "productType" to mapProductType(),
-        "subscriptionPeriod" to period?.iso8601
+        "subscriptionPeriod" to period?.iso8601,
+        "defaultOption" to defaultOption?.mapSubscriptionOption(this),
+        "subscriptionOptions" to subscriptionOptions?.map { it.mapSubscriptionOption(this) },
```
joshdholtz, line 47: These new fields were needed to expose subscription options that can be purchased

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
...
+    googleIsPersonalizedPrice: Boolean?,
+    onResult: OnResult
+) {
+    if (activity != null) {
+        val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
+            val optionToBuy = storeProducts.mapNotNull { storeProduct ->
+                storeProduct.subscriptionOptions?.map { Pair(storeProduct, it) }
+            }.flatten().firstOrNull { (storeProduct, subscriptionOption) ->
+                storeProduct.purchasingData.productId == productIdentifier && subscriptionOption.id == optionIdentifier
+            }?.second
```
beylmk, line 201: 🥴 could we add some intermediate vals for readability?

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
     }
 }
 
+fun purchaseSubscriptionOption(
+    activity: Activity?,
+    productIdentifier: String,
+    optionIdentifier: String,
```
beylmk, line 189: so will the hybrids pass through just the identifier, not the subscriptionoption itself?

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -123,4 +126,36 @@ private fun Period.mapPeriod(): Map<String, Any?>? {
             "periodNumberOfUnits" to 0
         )
     }
+}
+
+private fun SubscriptionOption.mapSubscriptionOption(storeProduct: StoreProduct): Map<String, Any?> {
+    return mapOf(
+        "id" to id,
+        "storeProductId" to storeProduct.id,
+        "productId" to storeProduct.purchasingData.productId,
```
beylmk, line 135: what's the difference bt storeProductId and productId?

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -630,6 +630,63 @@ internal class CommonKtTests {
         assertEquals(expectedProductIdentifier, receivedResponse?.get("productIdentifier"))
     }
 
+    @Test
```
beylmk, line 633: is it possible to add a test for that code that found the subscriptionOption by its id?

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
...
+                storeProduct.subscriptionOptions?.map { Pair(storeProduct, it) }
+            }.flatten().firstOrNull { (storeProduct, subscriptionOption) ->
+                storeProduct.purchasingData.productId == productIdentifier && subscriptionOption.id == optionIdentifier
+            }?.second
+
+            if (optionToBuy != null) {
+                val purchaseParams = PurchaseParams.Builder(optionToBuy, activity)
+
+                // Product upgrade
+                if (googleOldProductId != null && googleOldProductId.isNotBlank()) {
```
vegaro, line 207: Another way of doing this is:

```
                    googleOldProductId.takeUnless { it.isNullOrBlank() }?.let {
```

joshdholtz: Kotlin is so beautiful 😍

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
...
+    if (activity != null) {
+        val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
+            val optionToBuy = storeProducts.mapNotNull { storeProduct ->
+                storeProduct.subscriptionOptions?.map { Pair(storeProduct, it) }
+            }.flatten().firstOrNull { (storeProduct, subscriptionOption) ->
+                storeProduct.purchasingData.productId == productIdentifier && subscriptionOption.id == optionIdentifier
+            }?.second
+
+            if (optionToBuy != null) {
+                val purchaseParams = PurchaseParams.Builder(optionToBuy, activity)
```
tonidero, line 204: Hmm this is a bit of a nitpick and actually belongs in the android sdk, but I would have preferred activity to be the first parameter here...

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
...
+                Purchases.sharedInstance.purchaseWith(
+                    purchaseParams.build(),
+                    onError = getPurchaseErrorFunction(onResult),
+                    onSuccess = getPurchaseCompletedFunction(onResult)
+                )
+            } else {
+                onResult.onError(
+                    PurchasesError(
+                        PurchasesErrorCode.ProductNotAvailableForPurchaseError,
+                        "Couldn't find product."
```
tonidero, line 229: Should we add the missing productId + optionId missing to the error message?

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
     }
 }
 
+fun purchaseSubscriptionOption(
+    activity: Activity?,
```
tonidero, line 187: Hmm I know this was like this from the other purchase methods, but I wonder why make this nullable... Any context @vegaro ?

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
joshdholtz: This is... confusing a little bit 🤦‍♂️ And its documented in the Flutter model but it should also be documented here

- `id` is `{basePlanId}:{offerId}` which is used for matching `SubscriptionOption` on purchase
- `storeProductId` is `{subId}:{basePlanId}` is **NOT** used for any purchasing marching because it is done with `id` and `productId`
- `productId` is `{subId}` which is used for querying `getProductsWith(productId)` for only the product we need to find

It's not very clear so I should comment the use case of them here

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -184,6 +183,72 @@ fun purchasePackage(
...
+    productIdentifier: String,
+    optionIdentifier: String,
+    googleOldProductId: String?,
+    googleProrationMode: GoogleProrationMode?,
+    googleIsPersonalizedPrice: Boolean?,
+    onResult: OnResult
+) {
+    if (activity != null) {
+        val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
+            val optionToBuy = storeProducts.mapNotNull { storeProduct ->
```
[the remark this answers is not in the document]
joshdholtz: Oh, it needs to be `mapNotNull` because inside of this map it could return a null with `storeProduct.subscriptionOptions?`

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
joshdholtz: Yup yup! This is how they worked for `purchasePackage` and `purchaseProduct` 🤷‍♂️  I assume its so that each individual just needs to take the strings from their objects and pass them to PHC

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -194,6 +193,74 @@ fun purchasePackage(
...
+    googleOldProductId: String?,
+    googleProrationMode: GoogleProrationMode?,
+    googleIsPersonalizedPrice: Boolean?,
+    onResult: OnResult
+) {
+    if (activity != null) {
+        val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
+            val allSubscriptionOptions = storeProducts.mapNotNull { storeProduct ->
+                storeProduct.subscriptionOptions?.map { Pair(storeProduct, it) }
+            }.flatten()
```
beylmk, line 209: maybe some intermediate vars to make this more readable would be good

but i'm especially bad at following these chained calls so 🤷

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
beylmk: we'll need to add presentedOfferingId once my changes are merged

beylmk (changes requested):
couple minor things...

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
joshdholtz: Ah, sounds good! I'll do that in a separate PR since that requires a new Android version to be released but I'll create a JIRA task for that 💪

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
tonidero: We are also exposed to users passing a mismatching productId/optionId I guess... But we do log an error, which i think it's the best we can do, considering we are losing the original objects.

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -630,6 +625,63 @@ internal class CommonKtTests {
...
+    fun `purchaseSubsscriptionOption passes correct productIdentifier after a successful purchase`() {
```
tonidero, line 629:
```suggestion
    fun `purchaseSubscriptionOption passes correct productIdentifier after a successful purchase`() {
```

tonidero (approved):
After addressing the remaining concerns, this looks good to me. We are keeping this on a branch to make a beta first, correct?

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
joshdholtz: I need a spellchecker so bad

PR #365, opened 2023-04-03
[BC5] Rename `Period` fields to `unit` and `value`

joshdholtz:
## Motivation

The fields `periodUnit` and `periodNumberOfUnits` were too smurfy and made mapping more difficult to `unit` and `value` in hybrids

## Description

- Rename to `unit` and `value` when inside of a `SubscriptionOption`
- Kept `periodUnit` and `periodNumberOfUnits` on `StoreProduct` for backwards compatibility

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
@@ -77,7 +77,7 @@ internal fun StoreProduct.mapIntroPrice(): Map<String, Any?>? {
             // Check freeTrialPeriod first to give priority to trials
             // Format using device locale. iOS will format using App Store locale, but there's no way
             // to figure out how the price in the SKUDetails is being formatted.
-            freeTrialPeriod?.mapPeriod()?.let { periodFields ->
+            freeTrialPeriod?.mapPeriod(false)?.let { periodFields ->
                 mapOf(
                     "price" to 0,
                     "priceString" to formatUsingDeviceLocale(priceCurrencyCode, 0),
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -102,28 +102,31 @@ internal fun StoreProduct.mapIntroPrice(): Map<String, Any?>? {
     }
 }
 
-private fun Period.mapPeriod(): Map<String, Any?>? {
+private fun Period.mapPeriod(forSubscriptionOption: Boolean): Map<String, Any?>? {
+    val unitKey = if (forSubscriptionOption) "unit" else "periodUnit";
+    val valueKey = if (forSubscriptionOption) "value" else "periodNumberOfUnits"
```
tonidero, line 107: Hmm I think changing the mapping for the same object can be confusing... 

If we think having `unit` and `value` is enough info and considering when we release this, it will likely be a major, I think we should just rename it everywhere. If this is not released as a major, we could deprecate the old fields and keep them around. Thoughts?

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -102,28 +102,31 @@ internal fun StoreProduct.mapIntroPrice(): Map<String, Any?>? {
     }
 }
 
-private fun Period.mapPeriod(): Map<String, Any?>? {
+private fun Period.mapPeriod(forSubscriptionOption: Boolean): Map<String, Any?>? {
```
tonidero, line 105: Hmm I think it might be better to just have 2 mapping functions. I think it's different enough to warrant that?

tonidero (approved):
After talking with Josh, I understand the plan better about how the launch for the hybrids will go. Just left a suggestion but makes sense

PR #366, opened 2023-04-03
[BC5] Fix purchaseProduct to work with productIdentifiers with base plans

joshdholtz:
## Motivation

`purchaseProduct(productIdentifer)` was failing when product id of `subId:basePlanId` was being passed.

It was failing to fetch products since products are only fetched by `subId`

### The quick and backwards capability fix

- `productIdentifiers` is split by ":" when fetching products
- Finding the matching `StoreProduct` will compare `productIdentifier` to to `storeProduct.id`

### The better fix going forward

- There is a new optional `googleBasePlanId` parameter on `purchaseProducts`
- Finding the matching `StoreProduct` will look at:
  - `storeProduct.purchasingData.productId` for the `productIdentifier`
  - `storeProduct.googleProduct.basePlanId` for the `googleBasePlanId`


android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -74,8 +76,14 @@ fun purchaseProduct(
     if (activity != null) {
         val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
             val productToBuy = storeProducts.firstOrNull {
-                // TODO: Verify this works because "subId:basePlanId" (it should but strings are silly)
-                it.id == productIdentifier && it.type.name.equals(type, ignoreCase = true)
+                val isProduct = (it.id == productIdentifier && it.type.name.equals(type, ignoreCase = true))
+                val isBasePlan = (
+                    it.purchasingData.productId == productIdentifier
+                        && it.googleProduct?.basePlanId == googleBasePlanId
+                        && it.type.name.equals(type, ignoreCase = true)
```
beylmk, line 86: do we need to check the type again?

beylmk: also, is checking the type just for safety? or can you set up an OTP and a sub with the same product id?

joshdholtz: Technically... probably not 🤷‍♂️ Non-subs won't have this but also only subs will have

Can probably remove if you think we should!

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -74,8 +76,18 @@ fun purchaseProduct(
     if (activity != null) {
         val onReceiveStoreProducts: (List<StoreProduct>) -> Unit = { storeProducts ->
             val productToBuy = storeProducts.firstOrNull {
-                // TODO: Verify this works because "subId:basePlanId" (it should but strings are silly)
-                it.id == productIdentifier && it.type.name.equals(type, ignoreCase = true)
+                // Comparison for when productIdentifier is "subId:basePlanId"
+                val foundByProductIdContainingBasePlan =
+                    (it.id == productIdentifier && it.type.name.equals(type, ignoreCase = true))
+
+                // Comparison for when productIdentifier is "subId" and googleBasePlanId is "basePlanId"
```
tonidero, line 83: Hmm can you explain in what situations we would pass a "subId:basePlanId" from the hybrids, and when a "subId" and a "basePlanId" separately?

joshdholtz: The `purchaseStoreProduct(StoreProduct)` is not required but it seems like it would be nice to offer up this option when making this change 🤷‍♂️

tonidero: Hmm interesting... Ok, I would mention this as a comment here then? So there is more context why both comparisons are needed.

tonidero: And yeah, I think adding a `purchasesStoreProduct` would be great (sad that javascript doesn't support overloading :( ), and maybe even deprecating the one that receives a string? Not sure if we need that for a specific reason... CC @vegaro

tonidero (approved):
LGTM! I think just adding that explanation for context

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -62,10 +63,14 @@ fun getProductInfo(
     }
 }
 
+/**
+ *
```
beylmk, line 67: 👋

beylmk (approved):
looks good! thanks for the commenting and explaining

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
vegaro: the one that receives a string is from the old times when we didn't have typescript and we didn't have a Product interface.

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
vegaro: you actually can, I don't think anyone should though haha

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
joshdholtz: @vegaro Aye aye! Will add those to RN when I get back there 😊

vegaro: `purchaseStoreProduct(StoreProduct)` is going to remove the need to pass the type right? I am so excited about that change. There are so many support tickets of people that forget to pass the type...

PR #368, opened 2023-04-04
[BC5] Use Int for Google proration mode to make mapping logic to GoogleProrationMode reusable

joshdholtz:


android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -443,6 +476,21 @@ fun getPromotionalOffer() : ErrorContainer {
...
+        ?.takeIf { oldProductId != null }
+        ?.let { index ->
+            GoogleProrationMode.values().find {
+                it.playBillingClientMode == prorationMode
+            } ?: run {
+                throw InvalidProrationMode()
+            }
+        }
+}
+
```
joshdholtz, line 493: I wrote this so many different ways with lambdas, `Pair`s, copy/pasted implementation but this one seemed to be the least worst 🤷‍♂️ 

- Allows for returning of a nullable `GoogleProrationMode` without it being an error
- Testable code
- Clear error when not a valid value

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -443,6 +476,21 @@ fun getPromotionalOffer() : ErrorContainer {
 
 // region private functions
 
+internal class InvalidProrationMode(): Exception()
+
+@Throws(InvalidProrationMode::class)
+internal inline fun getGoogleProrationMode(oldProductId: String?, prorationMode: Int?) : GoogleProrationMode?  {
+    return prorationMode
+        ?.takeIf { oldProductId != null }
```
tonidero, line 484: Hmm why do we need this? If from the hybrids we pass an empty `oldProductId` but a proration mode, seems like that would be an issue in the hybrids. I think phc just needs to map the `prorationMode` int to the enum?

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -443,6 +476,21 @@ fun getPromotionalOffer() : ErrorContainer {
 
 // region private functions
 
+internal class InvalidProrationMode(): Exception()
```
tonidero, line 479: Maybe rename to `InvalidProrationModeException`? Not sure if that's the Java dev in me 😅

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
tonidero: Hmm we do have the [Result](https://github.com/RevenueCat/purchases-android/blob/main/utils/src/main/java/com/revenuecat/purchases/utils/Result.kt) class in android that could fit here... We can't use it directly here (since it's internal) so we would have to duplicate it and might remove some duplication.

But I think this is probably ok.

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -881,6 +882,72 @@ internal class CommonKtTests {
...
+        val prorationModeInt: Int? = -1
+
+        var catchWasCalled = false
+        try {
+            val mode = getGoogleProrationMode(oldProductId, prorationModeInt)
+        } catch (e: InvalidProrationMode) {
+            catchWasCalled = true
+        }
+
+        assert(catchWasCalled)
```
tonidero, line 933: Interesting we don't use AssertJ in PHC... Also I think there is an `assertTrue` we can use here?

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -881,6 +882,72 @@ internal class CommonKtTests {
...
+        val prorationModeInt: Int? = 1000
+
+        var catchWasCalled = false
+        try {
+            val mode = getGoogleProrationMode(oldProductId, prorationModeInt)
+        } catch (e: InvalidProrationMode) {
+            catchWasCalled = true
+        }
+
+        assert(catchWasCalled)
```
tonidero, line 948: Same here

tonidero (approved):
LGTM!

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
joshdholtz: It was just for an additional safety check 🤷‍♂️ Didn't want to possibly throw an exception if a bad proration mode was entered for some reason. Want to limit the check _only_ if we really needed to (which is when an `oldProductId` is also set.

But probably overkill 😅  I'm fine removing this if you think that's better! 

Thoughts?

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
joshdholtz: I did look into using Result as well 😅 I pretty much hated how all of the solutions to this looked 😛

PR #369, opened 2023-04-04
[BC5] Add `iso8601` to `Period` for subscription option pricing phases

joshdholtz:
## Motivation

Adds `iso8601` to `Period` for hybrids to use because why not if we have the data 🤷‍♂️ 

And also @vegaro requested it - https://github.com/RevenueCat/purchases-flutter/pull/641#discussion_r1156145971 😇 

## Description

Adds `iso8601` field to the `Period` mapping 👇 

```json
{
  "unit": "MONTH",
  "value": 1,
  "iso8601": "P1M"
}
```

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
@@ -23,7 +23,7 @@ private val StoreProduct.introductoryPhase: PricingPhase?
     get() = this.subscriptionOptions?.introOffer?.introPhase
 val StoreProduct.introductoryPrice: String?
     get() = this.introductoryPhase?.price?.formatted
-val StoreProduct.introductoryPricePeriodNEW: Period?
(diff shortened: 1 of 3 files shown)


tonidero (approved):

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -151,7 +151,7 @@ private fun Period.mapPeriod(): Map<String, Any?>? {
             "unit" to "DAY",
             "value" to 0
         )
-    }
+    } + mapOf("iso8601" to this.iso8601)
```
tonidero, line 154: Should we add it to the other mapping (for store product) as well? It shouldn't really break anything and that way we have the extra data, if we need it.

If we do, we might want to extract the key to a constant.

joshdholtz: The iSO 8601 version is already there in the store product version 😇

PR #371, opened 2023-04-06
Add  platform check,`OfferPaymentMode`, and `presentedOfferingIdentifier`

joshdholtz:
## Motivation

[CF-1329](https://revenuecats.atlassian.net/browse/CF-1329)

- Updated to Android `6.1.1`
- New `Purchases.sharedInstance.store` to limit `purchaseSubscriptionOption()` only for `PLAY_STORE`
- New `OfferPaymentMode` on `PricingPhase` and put in mapper
- Need to show `presentedOfferingIdentifier` in mapper and pass into purchase methods

## Description

- Bumped native Android version to `6.1.1`
- Added `offerPaymentMode` to the `PricingPhase` mapper
- Check `Purchases.sharedInstance.store` in `purchaseSubscriptionOption()` only for `PLAY_STORE`
- Added `presentedOfferingIdentifier` to `StoreProduct` and `SubscriptionOption` mapper
- `purchaseProduct()` and `purchaseSubscriptionOption()` have a new `presentedOfferingIdentifier: String?` parameter
- Fixed `purchaseSubscriptionOption()` tests that were actually called `purchaseProduct()` 🤦‍♂️ 

### Reapplying the `presentedOfferingIdentifier`

In the native Android SDK, the `presentedOfferingIdentifier` is a property of `StoreProduct` and `SubscriptionOption`. When the users called `purchase()` with one of these objects, that offering identifier is used without the developer having to do anything.

Because the way hybrids works, the `StoreProduct`s and `SubscriptionOption`s instances that get fetched and are aren't up to the hybrid level, aren't the same instances that are purchased going back down to the native layer. Because of this, the `presentedOfferingIdentifier` needs to get passed to the hybrid common `purchaseProduct()` and `purchaseSubscriptionOption()` methods and reapplied.

[CF-1329]: https://revenuecats.atlassian.net/browse/CF-1329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -23,6 +23,7 @@ import com.revenuecat.purchases.logInWith
 import com.revenuecat.purchases.logOutWith
 import com.revenuecat.purchases.models.BillingFeature
 import com.revenuecat.purchases.models.GoogleProrationMode
+import com.revenuecat.purchases.models.GoogleSubscriptionOption
 import com.revenuecat.purchases.models.StoreProduct
 import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.models.googleProduct
@@ -71,6 +72,7 @@ fun purchaseProduct(
     googleOldProductId: String?,
     googleProrationMode: Int?,
     googleIsPersonalizedPrice: Boolean?,
+    presentedOfferingIdentifier: String?,
     onResult: OnResult
 ) {
     val googleProrationMode = try {
@@ -102,7 +104,8 @@ fun purchaseProduct(
                 // 1) When productIdentifier is "subId:basePlanId" format (for backwards compatibility with hybrids)
                 // 2) When productIdentifier is "subId" and googleBasePlanId is "basePlanId"
                 foundByProductIdContainingBasePlan || foundByProductIdAndGoogleBasePlanId
-            }
+            }?.applyOfferingIdentifier(presentedOfferingIdentifier)
+
             if (productToBuy != null) {
                 val purchaseParams = PurchaseParams.Builder(activity, productToBuy)
 
@@ -240,6 +243,7 @@ fun purchaseSubscriptionOption(
     googleOldProductId: String?,
     googleProrationMode: Int?,
     googleIsPersonalizedPrice: Boolean?,
+    presentedOfferingIdentifier: String?,
     onResult: OnResult
 ) {
     val googleProrationMode = try {
@@ -258,10 +262,13 @@ fun purchaseSubscriptionOption(
             // Iterates over StoreProducts and SubscriptionOptions to find
             // the first matching product id and subscription option id
             val optionToPurchase = storeProducts.firstNotNullOfOrNull { storeProduct ->
-                storeProduct.subscriptionOptions?.firstOrNull { subscriptionOption ->
-                    storeProduct.purchasingData.productId == productIdentifier &&
+                // Create StoreProduct copy with presentedOfferingIdentifier if exists
+                // This will give all SubscriptionOption the presentedOfferingIdentifier
+                storeProduct.applyOfferingIdentifier(presentedOfferingIdentifier)
+                    .subscriptionOptions?.firstOrNull { subscriptionOption ->
+                        storeProduct.purchasingData.productId == productIdentifier &&
                             subscriptionOption.id == optionIdentifier
-                }
+                    }
             }
 
             if (optionToPurchase != null) {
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -258,10 +262,13 @@ fun purchaseSubscriptionOption(
...
             val optionToPurchase = storeProducts.firstNotNullOfOrNull { storeProduct ->
-                storeProduct.subscriptionOptions?.firstOrNull { subscriptionOption ->
-                    storeProduct.purchasingData.productId == productIdentifier &&
+                // Create StoreProduct copy with presentedOfferingIdentifier if exists
+                // This will give all SubscriptionOption the presentedOfferingIdentifier
+                storeProduct.applyOfferingIdentifier(presentedOfferingIdentifier)
+                    .subscriptionOptions?.firstOrNull { subscriptionOption ->
+                        storeProduct.purchasingData.productId == productIdentifier &&
                             subscriptionOption.id == optionIdentifier
-                }
```
joshdholtz, line 264: Applying the offering identifier to a `GoogleStoreProduct` will also apply it to all the subscription options.

This is kind of an inefficient way to do this but we don't expose a nice way to applying the offering identifier on a `SubscriptionOption` without having to cast to a `GoogleSubscriptionOption` which feels bad.

The cleanest way to do it is to just apply it a `StoreProduct` level which a `GoogleStoreProduct` will then apply it to all of its `SubscriptionOptions`.

This could be something to improve later on.

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -490,6 +497,12 @@ internal fun getGoogleProrationMode(prorationMode: Int?) : GoogleProrationMode?
         }
 }
 
+private fun StoreProduct.applyOfferingIdentifier(presentedOfferingIdentifier: String?) : StoreProduct {
+    return presentedOfferingIdentifier?.let {
+        this.copyWithOfferingId(it)
+    } ?: this
+}
```
joshdholtz, line 504: This is just a little helper to make the above code cleaner looking 🤷‍♂️

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -952,6 +1102,12 @@ internal class CommonKtTests {
 
 const val MICROS_MULTIPLIER = 1_000_000
 
+fun getPresentedOfferingId(purchaseParams: PurchaseParams) : String? {
+    // Uses reflection to test presentedOfferingIdentifier exists in purchasing params
+    val prop = PurchaseParams::class.members.firstOrNull { member -> member.name == "presentedOfferingIdentifier" }
+    return prop!!.call(purchaseParams) as? String?
+}
```
joshdholtz, line 1109: We don't make `presentedOfferingIdentifier` public on `PurchaseParams` so had to use reflection to test 😬 

Not sure if there is a better way to do this but it worked 🤷‍♂️

mcastany (approved):
Not sure if I'm the best to review it, but looks good to me

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -187,6 +190,7 @@ private fun PricingPhase.mapPricingPhase(): Map<String, Any?> {
         "recurrenceMode" to recurrenceMode.identifier,
         "billingCycleCount" to billingCycleCount,
         "price" to price.mapPrice(),
+        "offerPaymentMode" to offerPaymentMode?.toString()
```
tonidero, line 193: So this will be a string instead of an int right? For enums, we've been traditionally using ints, but I think this works too... We will map it back to an enum in the hybrids I guess?

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
tonidero: Well, the other way I was thinking would be mockking the `PurchaseParams` constructor (mockk has a `mockkConstructor` function) to return a `mockk` and we can validate that the methods we expect are called in that mockk. I would prefer that rather than reflection? I try to avoid reflection as much as possible 😬.

tonidero (approved):
I would try to avoid reflection if possible. Other than that, looks good

android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
joshdholtz: Yup yup , will map it back to an enum in the hybrids! I can turn it into an int though 😊 That _probably_ makes more sense

android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
tonidero: Just talked with @joshdholtz and it might be a bit tricky to use `mockkConstructor` here. For the sake of shipping and considering this is only a test and we are planning to refactor it in the future. We can go ahead with this.

joshdholtz: Talked with @tonidero about how to do this because I'm not a `mockk` expert and decided that we are going to keep this reflection test **FOR NOW**.

Will create a new task to expose `presentedOfferingIdentifier` as a getting on `PurchaseParams` which will allow us to remove this reflection test in the future.

PR #374, opened 2023-04-10
`CommonFunctionality.beginRefundRequest`: available on Catalyst

NachoSoto:
Fixes https://github.com/RevenueCat/react-native-purchases/pull/591.

These methods are actually available on Catalyst.


tonidero (approved):
Ahh right. Makes sense 👍

PR #376, opened 2023-04-12
[BC5] Rename `productCategory` to `productType` and `productType` to `productSubtype`

joshdholtz:
## Motivation

Rename... 
- `productCategory` to `productType`
- `productType` to `productSubtype`

This was confusing and making mapping of store mapper JSON to hybrid models kind of weird 🤷‍♂️ 

## Description

Rename... 
- `productCategory` to `productType`
- `productType` to `productSubtype`

None of these were being used before the BC5 versions and not being used on iOS so this is pretty safe to change right now 😇

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
@@ -1,7 +1,6 @@
 package com.revenuecat.purchases.hybridcommon.mappers
 
 import com.revenuecat.purchases.ProductType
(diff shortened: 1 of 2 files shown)


MarkVillacampa (approved):

vegaro:
`productCategory` exists in iOS and it has values of `subscription` and `nonSubscription`

https://github.com/RevenueCat/purchases-hybrid-common/blob/5a6ccf2cd7837889546a096967d0693b67a886c7/ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct%2BHybridAdditions.swift#L24

I am not sure we should be renaming this to be honest

PR #382, opened 2023-04-12
`CircleCI`: don't allow deploy until tests have finished

NachoSoto:
Right now the "hold" job can be approved and the release started before tests have passed. This prevents that.


NachoSoto:
Never mind, can't do this because those are 2 separate workflows.

PR #384, opened 2023-04-12
[BC5] Use new mapped product type values for `purchaseProduct()` and `getProductInfo()`

joshdholtz:
## Motivation

There were inconsistencies between the product type being sent to the hybrid on `StoreProduct` and the purchase type being used `purchaseProduct()` and `getProductInfo()`

## Description

This version of hybrid will now use `SUBSCRIPTION` and `NON_SUBSCRIPTION` as its `ProductType`.
This `ProductType` is used on `StoreProduct` and on the method calls for `purchaseProduct()` and `getProductInfo()`

Created a new hybrid common internal `MappedProductType` enum that is used for generating the mapped `StoreProduct` JSON and mapping the product type string on `purchaseProduct` and `getProductInfo` into native Android SDK `ProductType`s.

This still supports the legacy "subs" and "inapp" that come from deprecated hybrid methods that may still be used.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -19,6 +19,7 @@ import com.revenuecat.purchases.getCustomerInfoWith
 import com.revenuecat.purchases.getOfferingsWith
 import com.revenuecat.purchases.getProductsWith
 import com.revenuecat.purchases.hybridcommon.mappers.LogHandlerWithMapping
+import com.revenuecat.purchases.hybridcommon.mappers.MappedProductType
 import com.revenuecat.purchases.hybridcommon.mappers.map
 import com.revenuecat.purchases.logInWith
 import com.revenuecat.purchases.logOutWith
@@ -58,7 +59,7 @@ fun getProductInfo(
     val onError: (PurchasesError) -> Unit = { onResult.onError(it.map()) }
     val onReceived: (List<StoreProduct>) -> Unit = { onResult.onReceived(it.map()) }
 
-    if (type.equals("subs", ignoreCase = true)) {
+    if (mapStringTypeToProductType(type) == ProductType.SUBS) {
         Purchases.sharedInstance.getProductsWith(productIDs, ProductType.SUBS, onError, onReceived)
     } else {
         Purchases.sharedInstance.getProductsWith(productIDs, ProductType.INAPP, onError, onReceived)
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -493,6 +496,27 @@ fun getPromotionalOffer() : ErrorContainer {
...
 // region private functions
 
+internal fun mapStringTypeToProductType(type: String) : ProductType {
+    MappedProductType.values().firstOrNull { it.value.equals(type, ignoreCase = true) }?.let {
+        return when(it) {
+            MappedProductType.NON_SUBSCRIPTION -> ProductType.INAPP
+            MappedProductType.SUBSCRIPTION -> ProductType.SUBS
+            MappedProductType.UNKNOWN -> ProductType.UNKNOWN
+        }
+    }
```
swehner, line 506: I don't understand this block, wouldn't it be easier to do something similar as below?

```
return when(type.uppercase()) {
  MappedProductType.NON_SUBSCRIPTION.name -> ProductType.INAPP
  MappedProductType.SUBSCRIPTION.name -> ProductType.SUBS
  MappedProductType.UNKNOWN.name -> ProductType.UNKNOWN
```
I don't see how iterating over the values changes anything here, but maybe I'm missing something.

swehner (commented):
Looks good - minor comment on one of the type conversions

android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
joshdholtz: @swehner Then I have to make a map 😊 The string value of `MappedProductType` is already being sent over to the hybrid layer so this the suggested way to find its match when it comes back to hybrid common 🤷‍♂️ 

We _could_ make a map but then we'd have to remember to update the map if a new product type was added where this won't have to change if/when a new product type is added.

swehner: But it's the same thing, right? If you add a new value it won't match anything in the when clause and you need to update it anyways... 😕 i'm confusulated

joshdholtz: Oh, I see what you are saying! `when`s are exhaustive if we add a new type it will show a compile error like this 👇 

Screenshot 2023-04-12 at 12 06 03 PM

swehner: But it's also good, right? That's ideally what we wanto - the compiler to complain when a type will go unmapped...

joshdholtz: This wouldn't complain with a compile error because strings can't be exhausting 👇 

```kotlin
return when(type.uppercase()) {
  MappedProductType.NON_SUBSCRIPTION.name -> ProductType.INAPP
  MappedProductType.SUBSCRIPTION.name -> ProductType.SUBS
  MappedProductType.UNKNOWN.name -> ProductType.UNKNOWN
```

only a `when` with an enum type in it would yell at us for not being exhaustive like 👇 

```kotlin
when(type) {
  MappedProductType.NON_SUBSCRIPTION -> ProductType.INAPP
  MappedProductType.SUBSCRIPTION -> ProductType.SUBS
  MappedProductType.UNKNOWN -> ProductType.UNKNOWN
}
```

But... to make this cleaner I could add this property on `MappedProductType` so that mapping isn't in the main logic 🤷‍♂️ 

```kotlin
internal enum class MappedProductType(val value: String) {
    SUBSCRIPTION("SUBSCRIPTION"),
    NON_SUBSCRIPTION("NON_SUBSCRIPTION"),
    UNKNOWN("UNKNOWN");

    val toProductType: ProductType
        get() = when(this) {
            MappedProductType.NON_SUBSCRIPTION -> ProductType.INAPP
            MappedProductType.SUBSCRIPTION -> ProductType.SUBS
            MappedProductType.UNKNOWN -> ProductType.UNKNOWN
        }
}
```

swehner: Looks good

PR #385, opened 2023-04-12
Release/5.0.0-beta.5

joshdholtz:
### Breaking Changes
* [BC5] iOS rename `productCategory` to `product type` and `productType` to `productSubtype` (#377) via Josh Holtz (@joshdholtz)
* [BC5] Android rename `productCategory` to `productType` and `productType` to `productSubtype` (#376) via Josh Holtz (@joshdholtz)
### Other Changes
* [BC5] Use new mapped product type values for `purchaseProduct()` and `getProductInfo()` (#384) via Josh Holtz (@joshdholtz)


MarkVillacampa (approved):

aboedo:
this effectively _changes_ the field `productType`, right? Maybe that's worth calling out more explicitly - code will continue to look correct in that the field exists, but its type and values are different

aboedo:
Why was the change needed? Asking because every single breaking change hurts, we need to make sure there's absolutely no way around it

joshdholtz:
@aboedo 

## The History

_(You probably know all of this but calling this out for clarity)_

Android usage in PHC 'purchaseProduct(productIdentifer, type)` requires a product type as one of the arguments. This, in the current live version, was a string of "subs" or "inapp". If developers used this method, they would have to know or hardcode the this type themselves because our `StoreProduct` models **never had** the product category/type on them. 

Flutter never had this product type or category in the class. React Native never had this in the interface.

The field that changed was only ever in that JSON and was never directly used by our SDKS in and public property

## This Change

I wanted to provide a new purchase `purchaseStoreProduct()` API where the product category/type in the `StoreProduct` could be passed directly into the method to purchase to make a much easier API for the developers. This would require a different type because the `PurchaseType` that is currently used on purchase has different values. 

In think I got stuck in the weeds with dinner if these thoughts because now that I think about it, even though this name change is probably not 100% necessary, we can probably get by without these changes and still make it work.

## In Summary

I'm going to make revert this and make it work with what we had by having using the `ProductCategory` terminology.

PR #387, opened 2023-04-13
[BC5] Add product category  support in `getProductInfo()` and `purchaseProduct()`

joshdholtz:
## Motivation

Currently, the `purchaseProduct(identifier, type)` method requires the developer to manually pass the type (either "subs" or "inapp") of the product they are purchasing. Even if they are basing this off of a fetched `StoreProduct`, the developer would need to determine what type of product it is because its currently not exposed in any of the hybrid SDKs.

New hybrids versions will have a new `productCategory` property on `StoreProduct`. This value has always been sent from the Hybrid Common layer but was just never exposed on the models. It has values of "SUBSCRIPTION" and "NON_SUBSCRIPTION".

These values should be able to be passed into the type of `purchaseProduct()` without any developer modification to make a purchase

Example: 

```dart
Purchases.purchaseProduct(storeProduct.identifier, storeProduct.productCategory)
``` 

Goal of this is to **not** be a breaking API change but for Hybrid Common to handle these different values of legacy values ("subs" and "inapp") and the product categories of "SUBSCRIPTION" and "NON_SUBSCRIPTION" going forward.

## Description

Mostly taken from #384 (which was reverted in the great unmigration of product type)

- Moved hardcoded strings into `MappedProductCategory` enum with values of:
  - `SUBSCRIPTION`
  - `NON_SUBSCRIPTION`
  - `UNKNOWN`
- New `mapStringToProductType()` helper method for converting from `String` to `ProductType`
  - First checks if string is a `MappedProductCategory`
  - Then sees if legacy value of "subs" or "inapp"
- `mapStringToProductType()` is used in `getProductInfo()` and `purchaseProduct()

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -19,6 +19,7 @@ import com.revenuecat.purchases.getCustomerInfoWith
 import com.revenuecat.purchases.getOfferingsWith
 import com.revenuecat.purchases.getProductsWith
 import com.revenuecat.purchases.hybridcommon.mappers.LogHandlerWithMapping
+import com.revenuecat.purchases.hybridcommon.mappers.MappedProductCategory
 import com.revenuecat.purchases.hybridcommon.mappers.map
 import com.revenuecat.purchases.logInWith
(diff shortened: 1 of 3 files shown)


vegaro (approved):
Perfect I love this will add support for both the deprecated and the new product category. Just in case someone passes that 😄

PR #411, opened 2023-05-30
Remove `watchOS` as a supported platform

NachoSoto:
None of the hybrids support `watchOS`, and it's unlikely that we'l need this in the future. This allows us to remove the requirement for Xcode 13 as well.

PR #417, opened 2023-05-31
Makes pushing to Cocoapods its own job

vegaro:
Pushing to cocoapods keeps failing until they fix https://github.com/CocoaPods/CocoaPods/issues/11621

We should make pushing to pods its own job so it doesn't break the release process

This PR also cleans up the config.yaml a bit by updating to the latest version of the orb

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,9 +303,6 @@ workflows:
             - deploy-android
       - push-pods:
           <<: *release-tags
-          # Xcode 14 not supported until https://github.com/CocoaPods/CocoaPods/issues/11558 is fixed.
-          # This runs as its own job until https://github.com/CocoaPods/CocoaPods/issues/11621 is fixed.
-          xcode_version: '13.4.1'
           requires:
             - make-github-release
       - trigger-dependent-updates:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -296,6 +305,13 @@ workflows:
           requires:
             - deploy-ios
             - deploy-android
+      - push-pods:
+          <<: *release-tags
+          # Xcode 14 not supported until https://github.com/CocoaPods/CocoaPods/issues/11558 is fixed.
```
NachoSoto, line 310: Actually we should be able to use `14.3` since I did #411. This issue was about `watchOS` but that's why I removed it so we could finally move everything here to Xcode 14 :)

NachoSoto (commented):
Thank you 🙏🏻

NachoSoto (approved):

.circleci/config.yml
```diff
@@ -154,6 +141,18 @@ jobs:
           command: bundle exec fastlane ios release
```
NachoSoto, line 141: Oh we needed to remove this.

PR #418, opened 2023-06-02
Fix tests by removing mocking of currency formatters which weren't ne…

joshdholtz:
## Motivation

Fix failing Android tests

## Description

- Remove mocking of currency formatters (which weren't being used anymroe) which were causing some Java reflection issues
- Fixed tests for `isPersonalizePrice` there were failed due to Android `6.4.0` upgrade

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -75,7 +75,7 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'androidx.annotation:annotation:1.6.0'
+    implementation 'androidx.annotation:annotation:1.3.0'
 
     latestDependenciesApi "com.revenuecat.purchases:purchases:$purchases_version"
     latestDependenciesApi "com.revenuecat.purchases:purchases-core-common:$purchases_version"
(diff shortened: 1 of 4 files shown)


android/src/test/java/com/revenuecat/purchases/hybridcommon/CommonKtTests.kt
```diff
@@ -423,7 +423,7 @@ internal class CommonKtTests {
             mockPurchases.purchase(any<PurchaseParams>(), capture(capturedPurchaseCallback))
         } answers {
             val params = it.invocation.args.first() as PurchaseParams
-            assertEquals(false, params.isPersonalizedPrice)
+            assertEquals(null, params.isPersonalizedPrice)
```
NachoSoto, line 426: Should this be `assertNull` now?
But I'm confused why this boolean is nullable now?

joshdholtz: Will change to `assertNull` 💪 

And it was always nullable (see https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt#L11) but there was a bug where it was being default to `false` (so it was never `null`)

This was fixed between `6.1.1` and `6.4.0` which is why these tests are changing now 😊

android/build.gradle
```diff
@@ -75,7 +75,7 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation 'androidx.annotation:annotation:1.6.0'
+    implementation 'androidx.annotation:annotation:1.3.0'
```
MarkVillacampa, line 78: Helpful comment for future us maybe?

```suggestion
    // TODO: Upgrading to 1.6.0 depends on kotlin stdlib 1.8 which breaks the build. Upgrade once we're on Kotlin 1.8.
    implementation 'androidx.annotation:annotation:1.3.0'
```

MarkVillacampa (approved):

PR #422, opened 2023-06-05
`MockStoreProduct`: fixed `Sendable` warnings

NachoSoto:
Fixes
> Stored property 'productCategory' of 'Sendable'-conforming class 'MockStoreProduct' is mutable

We can make it a `struct` since it's a `protocol`, solving the mutability issue by making it a value type.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Offering+HybridAdditionsTests.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Offering+HybridAdditionsTests.swift
@@ -68,11 +68,15 @@ class OfferingInfoHybridAdditionsTests: QuickSpec {
     }
 }
 
-private final class MockStoreProduct: StoreProductType, Sendable {
+// These are `enum`s, but for some reason Swift doesn't make them implicitly Sendable.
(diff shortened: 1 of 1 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Offering+HybridAdditionsTests.swift
```diff
@@ -68,11 +68,15 @@ class OfferingInfoHybridAdditionsTests: QuickSpec {
     }
 }
 
-private final class MockStoreProduct: StoreProductType, Sendable {
+// These are `enum`s, but for some reason Swift doesn't make them implicitly Sendable.
```
tonidero, line 71: Hmm I'm concerned of doing this without understanding the reason... As in, are we sure these are sendable in the public API?

NachoSoto: `enum`s without associated types are implicitly `Sendable`, just like `struct`s, because they don't have reference semantics. We could mark them as such in `purchases-ios`, but it's unnecessary. Since this is just a test type, I think it's easier to do this instead.

tonidero (approved):
Makes sense, thanks for explaining!

PR #448, opened 2023-06-27
CI: wait until pods have been pushed to trigger dependent updates

MarkVillacampa:
We were doing it in parallel which could cause dependent's bump PRs to fail if pods take longer to be pushed.

Screenshot 2023-06-27 at 21 10 33


NachoSoto (approved):
❤️  Thank you!

PR #450, opened 2023-06-27
`.composite-enable`: fixed file reference

NachoSoto:
We were using 2 different ways of referencing this file: `file()` and `new File`.

When using an absolute path, the first `file().exists()` was returning `true`, but `new File` was failing to find it.


tonidero (approved):

tonidero:
I guess we wouldn't be using absolute paths normally, but making it consistent makes sense 👍

PR #455, opened 2023-07-04
Move typescript interfaces to PHC

tonidero:
This PR moves most of the typescript interfaces from the react native SDK to PHC. The interfaces are mostly a copy-paste of those in react-native-purchases. I haven't copied the `Purchases` interface yet. Will do that separately since that will be a bit trickier considering that in react native we are using static functions.

Counterpart react native PR: https://github.com/RevenueCat/react-native-purchases/pull/684

##### TODO
- [x] Make sure they will be usable by both react native and capacitor. This includes using them within the hybrids and making sure they are exposed to consumers. Confirmed these interfaces are visible in the react native PR
- [x] Make sure the documentation is generated correctly: While typedoc separates the documentation into an "external" visibility, we can make it show those by default, so it should still be ok.
- [x] Add typescript linter
- [x] Add typescript API tests (Added in #459)
- [ ] Move rest of interfaces to PHC (Missing purchases interface, will do that one separately)
- [x] Publish to NPM


typescript/tsconfig.json
tonidero, line 1: This is also a copy-paste of react-native with some slight differences. I was reading the documentation for each of these and most of them made sense to keep.

typescript/apitesters/customerInfo.ts
```diff
@@ -0,0 +1,54 @@
+import {
+  LogInResult,
+  CustomerInfo,
+  PurchasesEntitlementInfo,
+  PurchasesEntitlementInfos,
+  PurchasesStoreTransaction
+} from "../src";
+
+function checkLoginResult(result: LogInResult) {
```
tonidero, line 9: About the API tests. I was thinking that, while we can have them here, we should also keep them in the hybrids just in case.

tonidero:
Once I get the approval, I will deploy this package to npm

.circleci/config.yml
```diff
@@ -202,6 +209,19 @@ jobs:
           name: Deployment
           command: bundle exec fastlane android deploy
 
+  deploy-typescript:
+    executor: node/default
+    working_directory: ~/purchases-hybrid-common/typescript
+    steps:
+      - checkout:
+          path: ~/purchases-hybrid-common
```
vegaro, line 217: Is this needed?

.circleci/config.yml
```diff
@@ -280,6 +300,17 @@ jobs:
           command: bundle exec fastlane bump_hybrid_dependencies repo_name:cordova-plugin-purchases
           when: always
 
+  typescript-lint:
+    executor: node/default
+    working_directory: ~/purchases-hybrid-common/typescript
+    steps:
+      - checkout:
+          path: ~/purchases-hybrid-common
```
vegaro, line 308: same question here

typescript/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@revenuecat/purchases-hybrid-typescript-interfaces",
```
vegaro, line 2: what about just?

```
@revenuecat/purchases-typescript
```

In case we want to add something that are not interfaces. The `hybrid` part I think is too generic and I wish I hadn't added it to `purchases-hybrid-common` lol

.circleci/config.yml
tonidero: Not really... I think I just need to update the `working_directory` to be a relative path instead. I just copied this from somewhere, but that's probably cleaner 👍

typescript/package.json
tonidero: Hmm I do wonder if it's too generic and might lead some people to confuse it and use it (note that this is meant to be internal... I guess we could call it `@revenuecat/purchases-typescript-internal` maybe? wdyt?

.circleci/config.yml
tonidero: So after testing it out, I actually think it's needed... If I change the working_directory but not this, it causes errors when creating the folders. I could cd to the folder before the command... But feels that this is better. Lmk if you think otherwise!

typescript/package.json
vegaro: I like `@revenuecat/purchases-typescript-internal` yes!

vegaro (approved):

PR #459, opened 2023-07-13
Add API tests to typescript interfaces

tonidero:
This PR adds API tests to the typescript interfaces package added in #455 

These are basically copied from the react native repo.


.circleci/config.yml
```diff
@@ -311,6 +311,17 @@ jobs:
       - run:
           command: yarn && yarn eslint
 
+  typescript-apitests:
+    executor: node/default
+    working_directory: ~/purchases-hybrid-common/typescript
+    steps:
+      - checkout:
+          path: ~/purchases-hybrid-common
```
vegaro, line 319: Is this needed?

vegaro (approved):

.circleci/config.yml
tonidero: As per https://github.com/RevenueCat/purchases-hybrid-common/pull/455#discussion_r1279118549, we are keeping it like this. We could move the `working_directory` to the steps, but this works for now.

PR #490, opened 2023-09-04
Use `esnext` module typescript compiler option instead of `commonjs`

tonidero:
This changes the `module` typescript compiler option from `commonjs` to `esnext`. `@revenuecat/purchases-capacitor` already uses `esnext`. As for RNP, it still uses `commonjs` but I haven't seen any issues after doing this change and testing locally.
This is meant to deal with a problem with using enums from `purchases-typescript-internal` in `@revenuecat/purchases-capacitor`. Additionally, it should solve https://github.com/RevenueCat/purchases-capacitor/issues/98

### DIFF
--- a/typescript/tsconfig.json
+++ b/typescript/tsconfig.json
@@ -16,7 +16,7 @@
     "noUnusedParameters": true,
     "skipLibCheck": true,
     "target": "es5",
-    "module": "commonjs",
+    "module": "esnext",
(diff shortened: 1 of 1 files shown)


tonidero:
Moving back to draft since we are going to try another alternative.

tonidero:
Closing this since we decided to just deploy a different package with the esm module.

PR #494, opened 2023-09-06
Support deploying separate ESNext typescript package

tonidero:
Our `purchases-capacitor` plugin uses `esnext` instead of `commonjs` to compile the typescript code. This was causing some issues where some code wasn't found, and generating some warnings: https://github.com/RevenueCat/purchases-capacitor/issues/98

This PR adds another package, `@revenuecat/purchases-typescript-internal-esm` for our typescript shared code. This package will be compiled using `esnext` and will be used by `purchases-capacitor` to fix those issues. 

We did try to bundle both in a single package but that was causing other issues. This should fix the issue and avoid those complications

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -364,24 +364,46 @@ end
 platform :typescript do
   desc "Deploy npm package to npm.js"
   lane :deploy do |options|
-    version_number = current_version_number
-    is_prerelease = Gem::Version.new(version_number).prerelease?
+    build_and_publish_typescript_package
+  end
 
-    build_args = [
-      'yarn && yarn build'
-    ]
+  lane :deploy_esm do |options|
+    make_typescript_package_esm
+    build_and_publish_typescript_package
+  end
+end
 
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -364,24 +364,46 @@ end
...
+  replace_text_in_files(
+    previous_text: "@revenuecat/purchases-typescript-internal",
+    new_text: "@revenuecat/purchases-typescript-internal-esm",
+    paths_of_files_to_update: ["typescript/package.json"]
+  )
+  replace_text_in_files(
+    previous_text: "commonjs",
+    new_text: "esnext",
+    paths_of_files_to_update: ["typescript/tsconfig.json"]
+  )
```
tonidero, line 386: I thought about having separate `package.json` and `tsconfig.json` files instead of just replacing these values... But that would cause some duplication. This seemed the easiest option for now.

vegaro (approved):
LGTM!

PR #497, opened 2023-09-15
Trigger automatic PHC updates in Capacitor plugin

tonidero:
This PR makes sure we trigger a PHC update in the Capacitor repo when a new PHC release is made


NachoSoto (approved):
👏🏻

PR #507, opened 2023-09-28
Add product type to typescript types

vegaro:
We are sending `productType` in the Kotlin and Swift side, but the property is missing from the types

### DIFF
--- a/typescript/apitesters/offerings.ts
+++ b/typescript/apitesters/offerings.ts
@@ -24,7 +24,7 @@ function checkProduct(product: PurchasesStoreProduct) {
   const productCategory: PRODUCT_CATEGORY | null = product.productCategory;
   const productType: PRODUCT_TYPE = product.productType;
   const defaultOption: SubscriptionOption | null = product.defaultOption;
-  const subscriptionOptions: SubscriptionOption[] | null =  product.subscriptionOptions;
+  const subscriptionOptions: SubscriptionOption[] | null = product.subscriptionOptions;
   const presentedOfferingIdentifier: string | null = product.presentedOfferingIdentifier;
 }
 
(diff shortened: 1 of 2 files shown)


typescript/apitesters/offerings.ts
```diff
@@ -22,6 +22,7 @@ function checkProduct(product: PurchasesStoreProduct) {
   const discounts: PurchasesStoreProductDiscount[] | null = product.discounts;
   const subscriptionPeriod: string | null = product.subscriptionPeriod;
   const productCategory: PRODUCT_CATEGORY | null = product.productCategory;
```
vegaro, line 24: I am not sure why this is nullable 🤔

typescript/apitesters/offerings.ts
```diff
@@ -126,43 +127,56 @@ function checkPrice(price: Price) {
...
+function checkOfferProductType(productType: PRODUCT_TYPE) {
+  switch (productType) {
+    case PRODUCT_TYPE.CONSUMABLE:
+    case PRODUCT_TYPE.NON_CONSUMABLE:
+    case PRODUCT_TYPE.NON_RENEWABLE_SUBSCRIPTION:
+    case PRODUCT_TYPE.AUTO_RENEWABLE_SUBSCRIPTION:
+    case PRODUCT_TYPE.PREPAID_SUBSCRIPTION:
+    case PRODUCT_TYPE.UNKNOWN:
+      break;
+
```
tonidero, line 180: Extra line break
```suggestion
```

typescript/src/offerings.ts
```diff
@@ -107,6 +107,10 @@ export interface PurchasesStoreProduct {
      * Product category.
      */
     readonly productCategory: PRODUCT_CATEGORY | null;
+    /**
+     * Product type.
```
tonidero, line 111: Hmm should we improve the documentation while we're at it?

typescript/src/offerings.ts
```diff
@@ -135,12 +139,44 @@ export enum PRODUCT_CATEGORY {
...
+     */
+    NON_RENEWABLE_SUBSCRIPTION = "NON_RENEWABLE_SUBSCRIPTION",
+
+    /**
+     * An auto-renewable subscription.
+     */
+    AUTO_RENEWABLE_SUBSCRIPTION = "AUTO_RENEWABLE_SUBSCRIPTION",
+
+    /**
+     * A subscription that is pre-paid.
```
tonidero, line 176: Correct me if I'm wrong but are some of these iOS/Android only? If so, should we add it to the documentation?

tonidero (approved):

typescript/src/offerings.ts
vegaro: That's correct, I will add it to the docs

tonidero (approved):

typescript/src/offerings.ts
```diff
@@ -107,6 +107,10 @@ export interface PurchasesStoreProduct {
      * Product category.
      */
     readonly productCategory: PRODUCT_CATEGORY | null;
+    /**
+     * The specific type of subscription or one time purchase this product represents
```
NachoSoto, line 111: Can you add a warning to indicate that this isn't available for StoreKit 1 products? https://github.com/RevenueCat/purchases-ios/blob/4.26.1/Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift#L111

tonidero: Did this in #515

PR #508, opened 2023-09-28
fix proxyURL not crashing if invalid url is passed

aboedo:
Fixes an issue where in iOS 17 onwards, passing in an invalid proxyURLString won't actually fail. 

https://twitter.com/developermaris/status/1701487361069031542?s=61&t=-F_XHhE9Npq5cORxmw3b8w

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -41,11 +41,15 @@ import RevenueCat
         set {
             if let value = newValue {
                 var url: URL?
+                #if swift(>=5.9)
                 if #available(iOS 17.0, macCatalyst 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *) {
                     url = URL(string: value, encodingInvalidCharacters: false)
                 } else {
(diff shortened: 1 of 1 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -40,7 +40,13 @@ import RevenueCat
         get { Purchases.proxyURL?.absoluteString }
         set {
             if let value = newValue {
-                guard let proxyURL = URL(string: value) else {
+                var url: URL?
```
aboedo, line 43: the good news is that we already had a test for this. 
The bad news is that the test didn't fail 😓  I suppose it's because we weren't testing on iOS 17

vegaro: Oh ugh, that's a bummer

vegaro (approved):

tonidero (approved):

aboedo:
oh, I guess older Xcode won't even compile that because it doesn't know which method we're talking about, so I'll have to do the xcode version / swift version workaround. fun

aboedo:
fixed (hopefully)

PR #510, opened 2023-09-29
Add in app messages API support

tonidero:
This adds a new API to be used by the hybrids to support deferring in app messages so they don't show automatically and provides an API to display these messages on convenience.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -219,7 +219,10 @@ import RevenueCat
     @objc(showStoreMessagesForTypes:completion:)
     static func showStoreMessages(forRawValues rawValues: Set<NSNumber>,
                                   completion: @escaping () -> Void) {
-        Self.sharedInstance.showStoreMessages(forRawValues: rawValues, completion: completion)
+        let storeMessageTypes = rawValues.map { number in
+            StoreMessageType(rawValue: number.intValue)
+        }
+        Self.sharedInstance.showStoreMessages(for: Set(storeMessageTypes), completion: completion)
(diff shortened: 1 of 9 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -215,7 +215,10 @@ import RevenueCat
     @objc(showStoreMessagesForTypes:completion:)
     static func showStoreMessages(forRawValues rawValues: Set<NSNumber>,
                                   completion: @escaping () -> Void) {
-        Self.sharedInstance.showStoreMessages(forRawValues: rawValues, completion: completion)
+        let storeMessageTypes = rawValues.map { number in
```
NachoSoto, line 218: I think this should be
```suggestion
        let storeMessageTypes = rawValues.compactMap { number in
```

NachoSoto (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -199,6 +199,41 @@ import RevenueCat
...
+@objc public extension CommonFunctionality {
+
+#if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+    @available(iOS 16.0, *)
+    @available(tvOS, unavailable)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @objc(showStoreMessagesCompletion:)
+    static func showStoreMessages(completion: @escaping () -> Void) {
+        _ = Task<Void, Never> {
```
tonidero, line 212: This is temporary. Planning to use completion block methods in iOS, but didn't expose it in the native SDK.

PR #572, opened 2023-11-20
Fix breaking change in objc in 7.4.0

tonidero:
This supports the old configure method after accidental breaking change introduced in 7.4.0

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
@@ -68,6 +68,29 @@ import RevenueCat
         return purchases
     }
 
+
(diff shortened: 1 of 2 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
```diff
@@ -22,7 +22,7 @@ import RevenueCat
                           usesStoreKit2IfAvailable: Bool = false,
                           dangerousSettings: DangerousSettings?,
                           shouldShowInAppMessagesAutomatically: Bool = true,
-                          verificationMode: String? = nil) -> Purchases {
+                          verificationMode: String?) -> Purchases {
```
tonidero, line 25: The compiler didn't complain 😕  but I think this is needed to disambiguate both configure methods.

vegaro (approved):

PR #574, opened 2023-11-20
Fix entitlement verification name typo

tonidero:
We set the wrong name in the typescript interface. Also adds API tests for purchases configuration to try to catch this more easily in the future.

### DIFF
--- a/typescript/src/purchasesConfiguration.ts
+++ b/typescript/src/purchasesConfiguration.ts
@@ -52,5 +52,5 @@ export interface PurchasesConfiguration {
    * Verification strictness levels for [EntitlementInfo].
    * See https://rev.cat/trusted-entitlements for more info.
    */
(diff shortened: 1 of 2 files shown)


typescript/src/purchasesConfiguration.ts
```diff
@@ -52,5 +52,5 @@ export interface PurchasesConfiguration {
    * Verification strictness levels for [EntitlementInfo].
    * See https://rev.cat/trusted-entitlements for more info.
    */
-  entitlementVerificationInfo?: ENTITLEMENT_VERIFICATION_MODE;
+  entitlementVerificationMode?: ENTITLEMENT_VERIFICATION_MODE;
```
tonidero, line 55: This is technically a breaking change, but I would argue the old name didn't make sense, so it's a fix.

vegaro (approved):

PR #586, opened 2023-12-05
Paywalls: Expose delegate in PaywallProxy to receive events

tonidero:
This allows to set a `PaywallViewControllerDelegate` to the `PaywallProxy` to allow to receive purchase events in the hybrids. This is a first step to fix: https://github.com/RevenueCat/purchases-flutter/issues/886

### TODO
- [x] Expose a delegate method that gets called when the PaywallViewController gets dismissed.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
@@ -61,7 +61,6 @@ extension PaywallProxy: PaywallViewControllerDelegate {
     public func paywallViewController(_ controller: PaywallViewController,
                                       didFinishPurchasingWith customerInfo: CustomerInfo) {
         self.delegate?.paywallViewController?(controller, didFinishPurchasingWith: customerInfo)
-        controller.dismiss(animated: true)
     }
 
     public func paywallViewController(_ controller: PaywallViewController,
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
@@ -73,6 +73,10 @@ extension PaywallProxy: PaywallViewControllerDelegate {
                                       didFinishRestoringWith customerInfo: CustomerInfo) {
         self.delegate?.paywallViewController?(controller, didFinishRestoringWith: customerInfo)
     }
+
+    public func paywallViewControllerWasDismissed(_ controller: PaywallViewController) {
+        self.delegate?.paywallViewControllerWasDismissed?(controller)
+    }
 }
 
 #endif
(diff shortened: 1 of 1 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -57,9 +60,20 @@ extension PaywallProxy: PaywallViewControllerDelegate {
...
                                       didFinishPurchasingWith customerInfo: CustomerInfo) {
+        self.delegate?.paywallViewController?(controller, didFinishPurchasingWith: customerInfo)
         controller.dismiss(animated: true)
     }
 
+    public func paywallViewController(_ controller: PaywallViewController,
+                                      didFinishPurchasingWith customerInfo: CustomerInfo,
+                                      transaction: StoreTransaction?) {
+        self.delegate?.paywallViewController?(controller, didFinishPurchasingWith: customerInfo, transaction: transaction)
+    }
```
vegaro, line 71: Do we need to dismiss the view controller here as well?

vegaro (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
tonidero: Currently in native iOS, both delegate functions are called. So if we do, we would be dismissing it twice

tonidero: Honestly, we probably should only have one of these I think... But that would be part of the native SDK.

vegaro: Makes sense. Can you add a comment? If it were me, I would have added that extra dismiss 🤭

tonidero: Added in https://github.com/RevenueCat/purchases-hybrid-common/pull/586/commits/2b4526062a20e1cec7ef702618c32b6d4eb71f21

tonidero:
I'm moving this back to draft. As I was working on it, I noticed we need to expose a method in the delegate for when the paywall is dismissed/cancelled without purchasing. It's much easier to do this on the native side so will hold this until that's ready

NachoSoto (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -57,9 +60,22 @@ extension PaywallProxy: PaywallViewControllerDelegate {
 
     public func paywallViewController(_ controller: PaywallViewController,
                                       didFinishPurchasingWith customerInfo: CustomerInfo) {
+        self.delegate?.paywallViewController?(controller, didFinishPurchasingWith: customerInfo)
         controller.dismiss(animated: true)
```
NachoSoto, line 64: This is done automatically now so let's remove it: https://github.com/RevenueCat/purchases-ios/pull/3517

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
NachoSoto: See my other comment regarding https://github.com/RevenueCat/purchases-ios/pull/3517
So we can remove the comment too :)

tonidero:
Holding this until https://github.com/RevenueCat/purchases-ios/pull/3493 is released.

NachoSoto:
@tonidero releasing it right now.

vegaro (approved):

PR #591, opened 2023-12-07
Merge dependency-update workflows and fix automatic-release being triggered when executing action

vegaro:
I noticed `automatic-release` was getting triggered after executing the `dependency-update` action manually in CircleCI.

I also noticed `weekly-run-workflow` and `dependency-update` could be merged by merging their starting conditions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -412,17 +412,13 @@ workflows:
           requires:
             - make-github-release
   
-  weekly-run-workflow:
+  dependency-update:
     when:
-      and:
-        - equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -435,8 +431,11 @@ workflows:
 
   automatic-release:
     when:
-      not:
-        equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+      and:
+        - not:
+            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+        - not:
+            equal: [ dependency-update, << pipeline.parameters.action >> ]
```
tonidero, line 438: Hmm I wonder if this might be error prone in case we add other actions in the future...

tonidero (approved):

PR #602, opened 2023-12-20
Add PaywallFragmentViewModel to share data with caller fragment/activity

tonidero:
This is an attempt to be able to have some sort of completion listener from the calling fragment/activity. The idea is to use a shared view model, which doesn't need to be exposed, to share data using the underlying activity. The flow would be:
- The calling activity/fragment would pass us a new `PaywallResultListener` to listen to the result of the paywall
- Internally, we create a `PaywallFragmentViewModel` using the underlying activity and set that listener there
- Then, in the `PaywallFragment` itself, we also get the same instance of the `PaywallFragmentViewModel`, giving us access to the same result listener.

Note that I haven't exposed a `PaywallListener` directly. The reason for that is that we're using a different activity for the paywall, so we can't use the underlying activity to communicate with that (we would need the underlying application context). This would require making the Paywall activity as a fragment instead of an activity in purchases-android.

### DIFF
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/PaywallFragment.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/PaywallFragment.kt
@@ -5,7 +5,6 @@ import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
 import androidx.lifecycle.ViewModelProvider
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
-import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityLauncher
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResult
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler
--- a/android/src/main/java/com/revenuecat/purchases/hybridcommon/PaywallFragment.kt
+++ b/android/src/main/java/com/revenuecat/purchases/hybridcommon/PaywallFragment.kt
@@ -2,16 +2,41 @@ package com.revenuecat.purchases.hybridcommon
 
 import android.os.Bundle
 import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentActivity
+import androidx.lifecycle.ViewModelProvider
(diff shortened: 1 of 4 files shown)


tonidero:
Tested this in flutter. I was able to make it work!! (get the result returned through the VM)

android/src/main/java/com/revenuecat/purchases/hybridcommon/PaywallFragment.kt
```diff
@@ -2,16 +2,40 @@ package com.revenuecat.purchases.hybridcommon
...
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityLauncher
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResult
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler
 
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
-internal class PaywallFragment(
-    private val requiredEntitlementIdentifier: String?,
-) : Fragment(), PaywallResultHandler {
+internal class PaywallFragment : Fragment(), PaywallResultHandler {
```
tonidero, line 13: Note that since we're using a fragment, this does require the hybrid to use some subclass of a `FragmentActivity`... But most hybrids provide some of those, without having to provide our own version.

NachoSoto: Remember that we found out that we needed to use an activity on Flutter, as they say that's the preferred method. Are you thinking of forcing users to use our own fragment parent class?

tonidero: No, the idea is for the developer to use `FlutterFragmentActivity` (provided by flutter) instead of the `PurchasesFlutterActivity`. That way, the current `presentPaywallFromFragment` method we have here should take care of launching this fragment and from there the activity.

I believe the reason we created that custom activity was to try to get a result back from the paywall activity right? @NachoSoto

tonidero: That would be solved with the view model we're adding here

tonidero:
> If this works then sounds great!

Yup it works well! The only thing I see is that the activity in Android has the title bar, which it didn't use to have, but I think that's unrelated to this change and possibly related to https://github.com/RevenueCat/purchases-android/pull/1511

PR #603, opened 2023-12-21
Re-organize android library

tonidero:
This is a pretty big change in terms of changes, but it's mostly moving files around. Until now, the `android` folder was both a module and a project. That's against how gradle projects are usually structured, where you have a root project and one or more modules.

In this PR, we change this organization to use the proper one of having a root project with multiple modules. Now, we have a project with 2 modules:
- api-tests
- hybridcommon

Some advantages of this:
- Api tests are now actually separated from the library module, so they can more accurately reflect what API is really available to consumers. 
- Now we can separate the revenuecat ui library much more easily by creating a different module (Will be added in a followup PR)


android/gradle.properties
```diff
@@ -15,17 +15,15 @@ org.gradle.jvmargs=-Xmx1536m
...
+# resources declared in the library itself and none from the library's dependencies,
+# thereby reducing the size of the R class for that library
+android.nonTransitiveRClass=true
 
-# Maven
 GROUP=com.revenuecat.purchases
 VERSION_NAME=8.1.2
-POM_NAME=purchases-hybrid-common
-POM_PACKAGING=aar
-POM_ARTIFACT_ID=purchases-hybrid-common
```
tonidero, line 28: This will now be part of each module deployed

android/gradlew
```diff
@@ -28,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
tonidero, line 47: Many of these changes were added when I created a new project from Android Studio... I don't think they should hurt so I left them.

android/hybridcommon/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
+@Suppress("DSL_SCOPE_VIOLATION") // TODO: Remove once KTIJ-19369 is fixed
```
tonidero, line 1: I took the chance to migrate to kotlin DSL instead of groovy.

fastlane/Fastfile
```diff
@@ -13,7 +13,6 @@
 files_with_version_number = {
     './.version' => ['{x}'],
     './PurchasesHybridCommon.podspec' => ['s.version          = "{x}"'],
-    './android/build.gradle' => ['versionName "{x}"'],
```
tonidero, line 16: versionCode/versionName is not needed in libraries anymore so I took the opportunity to remove it from there

android/.idea/runConfigurations/All_tests.xml
```diff
@@ -0,0 +1,24 @@
+<component name="ProjectRunConfigurationManager">
```
tonidero, line 1: Added this new run configuration to be able to run all tests more easily

android/api-tests/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
+@Suppress("DSL_SCOPE_VIOLATION") // TODO: Remove once KTIJ-19369 is fixed
+plugins {
+    alias(libs.plugins.androidLibrary)
+    alias(libs.plugins.kotlinAndroid)
+    alias(libs.plugins.detekt)
```
tonidero, line 5: Right now I'm applying detekt to each individual module. In purchases-android, we created a task to run it on all modules instead. I thought about doing that, but I thought to follow the default route first.

android/hybridcommon/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+}
+
+dependencies {
+
```
vegaro, line 50: nit pick, extra space

android/gradle.properties
```diff
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx1536m
+org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
```
vegaro, line 9: great

android/.idea/codeStyles/Project.xml
```diff
@@ -1,135 +0,0 @@
-<component name="ProjectCodeStyleConfiguration">
```
vegaro, line 1: I think we had this in the repo to share the codeStyle? I am not sure, might need to check when this file was added

android/hybridcommon/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+detekt {
+    buildUponDefaultConfig = true
+    baseline = file("config/detekt/detekt-baseline.xml")
+}
+
+android {
+    namespace = "com.revenuecat.purchases.hybridcommon"
+    compileSdk = 34
```
vegaro, line 21: This is being updated from 33, not sure if it can cause any issue (I don't think so)

android/hybridcommon/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+    api(libs.purchases.ui)
+    api(libs.purchases.amazon)
+    testImplementation(libs.kotlin.test)
+    testImplementation(libs.junit)
+    testImplementation(libs.assertj.core)
+    testImplementation(libs.mockk)
+    testImplementation(libs.junit.jupiter.api)
+    testRuntimeOnly(libs.junit.jupiter.engine)
+    androidTestImplementation(libs.androidx.test.ext.junit)
+    androidTestImplementation(libs.espresso.core)
```
vegaro, line 62: do we need these two?

android/hybridcommon/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+task("disableLocalBuild") {
+    group = "Tools"
+    description = "Disable composite build"
+    doLast {
+        val file = File(".composite-enable")
+        if (file.exists()) {
+            file.delete()
+        }
+    }
+}
```
vegaro, line 85: I believe these two tasks should be in the root's build.gradle because the settings.gradle that checks for its existence is in the root? I am not sure tho...

android/hybridcommon/build.gradle.kts
```diff
@@ -0,0 +1,85 @@
...
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+}
+
+dependencies {
+
+    implementation(libs.core.ktx)
```
vegaro, line 51: Do these include annotation and fragment? Because we used to have 

```
    implementation 'androidx.annotation:annotation:1.3.0'
    implementation 'androidx.fragment:fragment-ktx:1.6.2'
    ```

android/hybridcommon/build.gradle.kts
tonidero: Yeah, it was part of adding the new module. I believe it should be fine after checking the changes. Thanks for checking!

android/hybridcommon/build.gradle.kts
tonidero: Espresso at least no. Will remove it! As for the other one, I will check 👍

android/.idea/codeStyles/Project.xml
tonidero: Hmm yeah it's from this PR: https://github.com/RevenueCat/purchases-hybrid-common/pull/73... I do wonder whether it's needed, considering we now have detekt... But I'm ok bringing it back

android/hybridcommon/build.gradle.kts
tonidero: You're right! Thanks!

android/hybridcommon/build.gradle.kts
tonidero: I believe so, at least, it compiles and runs tests fine.

vegaro (approved):

PR #604, opened 2023-12-21
Separate paywalls into a different module

tonidero:
This PR separates paywalls code into a different module. This way we don't force users of the main hybrid common sdk to have to bump their min version, which is a breaking change.

This new dependency can be used with `implementation "com.revenuecat.purchases:purchases-hybrid-common-ui:$phc_version"`


android/hybridcommon/build.gradle.kts
```diff
@@ -21,8 +21,7 @@ android {
     compileSdk = 34
 
     defaultConfig {
-        // TODO: revert this to 19
-        minSdk = 24
+        minSdk = 19
```
tonidero, line 24: 🎉

android/hybridcommon-ui/build.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+    }
+}
+
+dependencies {
+    implementation(libs.core.ktx)
+    implementation(libs.fragment.ktx)
+    api(libs.purchases)
```
tonidero, line 47: Need to also expose the main library here since we're exposing some classes from the native SDK.

android/hybridcommon/build.gradle.kts
NachoSoto: Yayyyyyyy

NachoSoto (approved):
❤️

PR #606, opened 2023-12-22
Add shouldDisplayDismissButton

vegaro:
Add shouldDisplayDismissButton to launchPaywall so we can add this parameter to the hybrids

### DIFF
--- a/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
+++ b/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
@@ -4,6 +4,7 @@ import android.os.Bundle
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
 import androidx.lifecycle.ViewModelProvider
+import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityLauncher
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResult
@@ -13,6 +14,7 @@ import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler
 internal class PaywallFragment : Fragment(), PaywallResultHandler {
     companion object {
         private const val requiredEntitlementIdentifierKey = "requiredEntitlementIdentifier"
+        private const val shouldDisplayDismissButtonKey = "shouldDisplayDismissButton"
         const val tag: String = "revenuecat-paywall-fragment"
 
         @JvmStatic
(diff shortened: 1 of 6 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -26,13 +26,15 @@ import UIKit
     }
 
     @objc
-    public func presentPaywall() {
+    public func presentPaywall(
+        displayCloseButton: Bool = false
```
tonidero, line 30: I wonder if we should make it nullable instead? So we can use the default defined in `purchases-ios`

tonidero: Actually, not sure if that would work in objc, since nullable booleans didn't work well if I remember correctly

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -26,13 +26,15 @@ import UIKit
     }
 
     @objc
-    public func presentPaywall() {
+    public func presentPaywall(
```
tonidero, line 29: Should we add this to the api tests?

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
```diff
@@ -13,19 +14,22 @@ import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler
...
         private const val requiredEntitlementIdentifierKey = "requiredEntitlementIdentifier"
+        private const val shouldDisplayDismissButtonKey = "shouldDisplayDismissButton"
         const val tag: String = "revenuecat-paywall-fragment"
 
         @JvmStatic
         fun newInstance(
             activity: FragmentActivity,
             requiredEntitlementIdentifier: String? = null,
             paywallResultListener: PaywallResultListener? = null,
+            shouldDisplayDismissButton: Boolean? = null,
```
tonidero, line 25: Do we need to add this parameter to the `PaywallsHelper`?

tonidero:
I would change the label to `RevenueCatUI`

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
vegaro: I tried and I get 

`Method cannot be marked @objc because the type of the parameter cannot be represented in Objective-C`

I can maybe do two functions?

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -32,11 +32,19 @@ import UIKit
...
-        rootController.present(controller, animated: true)
+    @objc
+    public func presentPaywall(displayCloseButton: Bool) {
+        guard let rootController = UIApplication.shared.keyWindow?.rootViewController else {
+            NSLog("Unable to find root UIViewController")
+            return
+        }
+
+        let controller = PaywallViewController(displayCloseButton: displayCloseButton)
+        configureAndPresentPaywall(controller, rootController: rootController)
```
tonidero, line 47: Hmm considering this is very close to the other `presentPaywall`, could we create a private function with the optional `displayCloseButton` parameter instead?

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -53,6 +61,26 @@ import UIKit
         }
     }
 
+    @objc
+    public func presentPaywallIfNeeded(requiredEntitlementIdentifier: String, displayCloseButton: Bool) {
```
tonidero, line 65: Same for this one and the previous `presentPaywallIfNeeded` (into a different method)

tonidero, line 65: We're missing this from the api tests I think?

tonidero (approved):

aboedo (approved):

PR #612, opened 2023-12-22
Add offering parameter for launching paywalls

vegaro:
Based off #606

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
@@ -27,37 +27,48 @@ import UIKit
 
     @objc
     public func presentPaywall() {
-        presentPaywall(optionalDisplayCloseButton: nil)
+        presentPaywall(optionalDisplayCloseButton: nil, optionalOffering: nil)
     }
 
     @objc
     public func presentPaywall(displayCloseButton: Bool) {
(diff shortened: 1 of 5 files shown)


aboedo (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -27,53 +27,65 @@ import UIKit
...
                 }
             } catch {
                 NSLog("Failed presenting paywall: \(error)")
             }
         }
     }
 
-    private func presentPaywall(optionalDisplayCloseButton: Bool?) {
+    private func presentPaywall(optionalDisplayCloseButton: Bool?,
+                                optionalOffering: Offering?) {
```
NachoSoto, line 80: Nit: this is against the Swift guidelines, optional shouldn't be in the name:
```suggestion
    private func presentPaywall(displayCloseButton: Bool?,
                                offering: Offering?) {
```

NachoSoto (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
vegaro: What would be a good naming so it doesn't colide with the other `presentPaywall` function? That one doesn't have optionals as parameters since it's `@objc`

PR #615, opened 2023-12-26
Add lefthook to configure a pre-push hook that runs detekt

vegaro:
Added [lefthook](https://github.com/evilmartians/lefthook) and configured a pre-push hook that runs detekt locally before pushing


aboedo (approved):
oh this is cool!

tonidero:
Oh interesting!

PR #621, opened 2023-12-26
add pricePerPeriod properties

aboedo:
Added `pricePerWeek`, `pricePerMonth`, `pricePerYear` and string variants of them. Based on https://github.com/RevenueCat/purchases-ios/pull/3546

**Edits by @JayShortway:**
- Added variants for Android, iOS and Typescript.
- This seems to be all that's needed for [React Native](https://github.com/RevenueCat/react-native-purchases/pull/1045) and [Kotlin Multiplatform](https://github.com/RevenueCat/purchases-kmp/pull/131), but I didn't check other hybrids yet. Please let me know if you see something that wont work in a certain hybrid.


ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
```diff
@@ -21,27 +21,40 @@ internal extension StoreProduct {
...
 
-        if #available(iOS 11.2, macOS 10.13.2, tvOS 11.2, watchOS 6.2, *) {
-            if let subscriptionPeriod = self.subscriptionPeriod {
-                dictionary["subscriptionPeriod"] = StoreProduct.rc_normalized(subscriptionPeriod: subscriptionPeriod)
+            if #available(iOS 12.2, tvOS 12.2, macOS 10.14.4, *) {
+                dictionary["discounts"] = self.discounts.map { $0.rc_dictionary }
             }
-        }
 
+            if #available(watchOS 6.2, *) {
```
JayShortway, line 52: Lint complained about the check that was previously here: 
```swift
if #available(iOS 11.2, macOS 10.13.2, tvOS 11.2, watchOS 6.2, *)
```
saying that `iOS` was already covered by the enclosing check a few lines up:
```swift
if #available(iOS 11.2, tvOS 11.2, macOS 10.13.2, *)
```
I figured that Lint warning is also valid for `macOS` and `tvOS`, so I kept only `watchOS`.

I'm not 100% sure this is correct, so please double check.

typescript/src/offerings.ts
```diff
@@ -99,6 +99,51 @@ export interface PurchasesStoreProduct {
...
+  readonly pricePerWeekString: string;
+    /**
+   * Null for INAPP products. The price of the PurchasesStoreProduct formatted for the current
+   * locale in a monthly recurrence. This means that, for example, if the period is annual,
+   * the price will be divided by 12. It uses a currency formatter to format the price in the 
+   * given locale. Note that this value may be an approximation. For Google subscriptions,
+   * this value will use the basePlan to calculate the value.
+   */
+    readonly pricePerMonthString: string;
+      /**
```
tonidero, line 139: Indentation seems wrong?
```suggestion
  /**
```

typescript/src/offerings.ts
```diff
@@ -99,6 +99,51 @@ export interface PurchasesStoreProduct {
...
+  readonly pricePerYear: number;
+  /**
+   * Null for INAPP products. The price of the PurchasesStoreProduct formatted for the current
+   * locale in a weekly recurrence. This means that, for example, if the period is monthly,
+   * the price will be divided by 4. It uses a currency formatter to format the price in the 
+   * given locale. Note that this value may be an approximation. For Google subscriptions,
+   * this value will use the basePlan to calculate the value.
+   */
+  readonly pricePerWeekString: string;
+    /**
```
tonidero, line 131: Same
```suggestion
  /**
```

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
```diff
@@ -44,6 +44,12 @@ fun StoreProduct.map(): Map<String, Any?> =
         "currencyCode" to priceCurrencyCode,
         "introPrice" to mapIntroPrice(),
         "discounts" to null,
+        "pricePerWeek" to pricePerWeek()?.amountMicros,
+        "pricePerMonth" to pricePerMonth()?.amountMicros,
+        "pricePerYear" to pricePerYear()?.amountMicros,
+        "pricePerWeekString" to pricePerWeek()?.formatted,
+        "pricePerMonthString" to pricePerMonth()?.formatted,
+        "pricePerYearString" to pricePerYear()?.formatted,
```
tonidero, line 52: Ohh right these are only in the `StoreProduct` level right now... I wonder if we should add something similar at the `PricingPhase` level as well in Android (which is used inside the `SubscriptionOption` model). These are also useful though, so not needed for this PR.

ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
tonidero: Hmm I don't know why there is a requirement for watchOS on this one TBH... But this seems the best way to avoid a breaking change. Maybe we could try to remove the checks and try on an older emulator...

tonidero (approved):
Left a couple comments but nothing blocking IMO.

typescript/src/offerings.ts
JayShortway: https://github.com/RevenueCat/purchases-hybrid-common/pull/621/commits/0805b7da184c9666155043eacbec4ab1b44c2ea7

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreProductMapper.kt
JayShortway: Good idea! As discussed, this would need some changes to purchases-android first, to add the other periods besides monthly, so I'll leave them out of this PR.

PR #622, opened 2023-12-27
Fix dismiss button not showing in Android when not specified

vegaro:
If the `shouldDisplayDismissButtonKey` was missing, the button wasn't being displayed

### DIFF
--- a/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
+++ b/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
@@ -45,7 +45,15 @@ internal class PaywallFragment : Fragment(), PaywallResultHandler {
         get() = arguments?.getString(requiredEntitlementIdentifierKey)
 
     private val shouldDisplayDismissButton: Boolean?
-        get() = arguments?.getBoolean(shouldDisplayDismissButtonKey)
+        get() {
+            return arguments?.let {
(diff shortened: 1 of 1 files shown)


android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
```diff
@@ -45,7 +45,15 @@ internal class PaywallFragment : Fragment(), PaywallResultHandler {
         get() = arguments?.getString(requiredEntitlementIdentifierKey)
 
     private val shouldDisplayDismissButton: Boolean?
-        get() = arguments?.getBoolean(shouldDisplayDismissButtonKey)
```
MarkVillacampa, line 48: Does using a default value of null not work in this case?

`get() = arguments?.getBoolean(shouldDisplayDismissButtonKey, null)`

MarkVillacampa (approved):

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
vegaro: ahhh that would be easier lol, I will make the update

vegaro: Actually, not, the default value needs to be a boolean so that won't work

PR #623, opened 2023-12-28
Release 8.10.1-beta.8

vegaro:
### RevenueCatUI
* Fix dismiss button not showing in Android when not specified (#622) via Cesar de la Vega (@vegaro)


aboedo (commented):
why is it both a hotfix and a new beta? wouldn't it just be a new beta if we're only fixing a bug?

vegaro:
Right... I will make it 8.10.0-beta.8 then

PR #624, opened 2024-01-02
Add `createFooterPaywallView` to `PaywallProxy`

vegaro:
Added a function to create a `PaywallFooterViewController`

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
@@ -26,6 +26,11 @@ import UIKit
     }
 
     @objc
+    public func createFooterPaywallView() -> UIViewController {
+        return PaywallFooterViewController()
+    }
+
+    @objc
     public func presentPaywall() {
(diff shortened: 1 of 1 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -25,6 +25,11 @@ import UIKit
         return UIHostingController(rootView: PaywallView())
     }
 
+    @objc
+    public func createFooterPaywallView() -> UIViewController {
+        return PaywallFooterViewController()
```
tonidero, line 30: Maybe you're planning to add this in future PRs but do we need to pass in optional parameters like the offering?

tonidero (approved):
Makes sense, for optional parameters, we can add them later I think.

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
vegaro: yes! I was planning on adding those whenever we add start adding the parameters to the hybrids, after we have the views fully working

PR #625, opened 2024-01-02
Release 8.10.0-beta.8

vegaro:
### RevenueCatUI
* Fix dismiss button not showing in Android when not specified (#622) via Cesar de la Vega (@vegaro)
* Add createFooterPaywallView to PaywallProxy (#624) via Cesar de la Vega (@vegaro)


vegaro:
We are adding a new function, so should this be a new minor + new beta number? I think just a new beta right? cc @aboedo @tonidero

tonidero:
> We are adding a new function, so should this be a new minor + new beta number? I think just a new beta right? cc @aboedo @tonidero

I think just a new beta right, since it hasn't shipped as stable yet

tonidero (approved):

PR #628, opened 2024-01-04
Support result present paywall

tonidero:
This supports having a listener for the result of presenting a paywall which can be used by the hybrids. Currently it's sent as a string, both in iOS and android and possible values are:
- `NOT_PRESENTED`
- `CANCELLED`
- `PURCHASED`
- `RESTORED`
- `ERROR` (only in android right now)


ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -30,50 +32,107 @@ import UIKit
...
 
+    @available(*, deprecated, message: "Use presentPaywall with onPaywallResult instead")
     @objc
     public func presentPaywall(offering: Offering, displayCloseButton: Bool) {
-        privatePresentPaywall(displayCloseButton: displayCloseButton, offering: offering)
+        self.privatePresentPaywall(displayCloseButton: displayCloseButton, offering: offering)
+    }
+
+    @objc
+    public func presentPaywall(onPaywallResult: @escaping (String) -> Void) {
```
aboedo, line 60: this name on a parameter feels very android-y, on iOS I'd have expected it to be called something like `paywallResultHandler`

aboedo (approved):
looking great, left a comment regarding naming for callback params on iOS

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
tonidero: Made the rename in https://github.com/RevenueCat/purchases-hybrid-common/pull/628/commits/2551cf217ae977773950ed0e9a6c5bea4f53c113

PR #637, opened 2024-01-08
Do not update hybrids automatically on prereleases

tonidero:
Avoid updating hybrids automatically whenever we make a new prerelease. We should only update the hybrids automatically on stable releases. For prereleases, I believe the bump should be performed manually

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -273,6 +273,10 @@ end
 
 desc "Run automatic bump in other repos"
 lane :bump_hybrid_dependencies do |options|
+  if current_version_prerelease?
+    UI.error("Hybrids won't update automatically for PHC prerelase versions")
+    next
+  end
   version_number_for_bump = current_version_number
   repo_name = options[:repo_name]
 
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -273,6 +273,10 @@ end
 
 desc "Run automatic bump in other repos"
 lane :bump_hybrid_dependencies do |options|
+  if current_version_prerelease?
+    UI.error("Hybrids won't update automatically for PHC prerelase versions")
+    next
```
NachoSoto, line 278: This isn't needed cause `error` throws I believe?

fastlane/Fastfile
```diff
@@ -273,6 +273,10 @@ end
 
 desc "Run automatic bump in other repos"
 lane :bump_hybrid_dependencies do |options|
+  if current_version_prerelease?
+    UI.error("Hybrids won't update automatically for PHC prerelase versions")
```
NachoSoto, line 277: This will make the pipeline fail, but I guess that's ok.

NachoSoto (approved):
Thanks for doing this! 🙏🏻

fastlane/Fastfile
tonidero: Actually, this won't make the pipeline fail. In order to do that, I would need to use `user_error!`. In this case, this is just logging but ends successfully

fastlane/Fastfile
tonidero: See my other comment, but `error` doesn't throw. That would be `user_error!`

NachoSoto:
👍🏻

PR #638, opened 2024-01-08
Don't trigger updates on dependents on non-stable releases

vegaro:
Only trigger dependent updates on stable releases


tonidero:
Ohh I followed a different approach #637 😅. I like this approach since it avoids extra CI time, so I'm happy with this. Do you mind reverting that change then?

vegaro:
Lol I didn't notice, I will revert that change and point this branch against paywalls

tonidero (approved):

PR #652, opened 2024-01-12
Fix update native versions job after hybrid common android split

tonidero:
After the split in #604, we didn't update this command to use the new location when updating the native versions in PHC. This fixes that.


NachoSoto (approved):
🙏🏻

PR #666, opened 2024-01-19
`Paywalls`: add methods for presenting paywalls with an offering identifier (iOS)

NachoSoto:
Depends on https://github.com/RevenueCat/purchases-ios/pull/3587.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
@@ -24,17 +24,17 @@ import UIKit
                                                      result: PaywallResult)] = [:]
 
     @objc
-    public func createPaywallView() -> UIViewController {
+    public func createPaywallView() -> PaywallViewController {
         return PaywallViewController()
     }
 
     @objc
-    public func createPaywallView(offeringIdentifier: String) -> UIViewController {
+    public func createPaywallView(offeringIdentifier: String) -> PaywallViewController {
         return PaywallViewController(offeringIdentifier: offeringIdentifier)
     }
 
     @objc
-    public func createFooterPaywallView() -> UIViewController {
+    public func createFooterPaywallView() -> PaywallFooterViewController {
         let controller = PaywallFooterViewController()
         controller.delegate = self
 
@@ -42,7 +42,7 @@ import UIKit
     }
 
     @objc
-    public func createFooterPaywallView(offeringIdentifier: String) -> UIViewController {
+    public func createFooterPaywallView(offeringIdentifier: String) -> PaywallFooterViewController {
         let controller = PaywallFooterViewController(offeringIdentifier: offeringIdentifier)
         controller.delegate = self
 
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
@@ -25,7 +25,12 @@ import UIKit
 
     @objc
     public func createPaywallView() -> UIViewController {
-        return UIHostingController(rootView: PaywallView())
+        return PaywallViewController()
+    }
+
+    @objc
+    public func createPaywallView(offeringIdentifier: String) -> UIViewController {
+        return PaywallViewController(offeringIdentifier: offeringIdentifier)
     }
 
     @objc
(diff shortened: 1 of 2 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -36,46 +41,28 @@ import UIKit
...
 
     @objc
     public func presentPaywall(displayCloseButton: Bool, paywallResultHandler: @escaping (String) -> Void) {
         self.privatePresentPaywall(displayCloseButton: displayCloseButton,
-                                   offering: nil,
                                    paywallResultHandler: paywallResultHandler)
     }
 
     @objc
     public func presentPaywall(offering: Offering, paywallResultHandler: @escaping (String) -> Void) {
```
NachoSoto, line 64: There's really no reason we have all these overloads. Better to have just a couple of methods and define default values for things like "displayCloseButton" in the hybrids, instead of checking 100 different permutations of parameters to decide which method to call.

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -216,4 +214,51 @@ extension PaywallProxy: PaywallViewControllerDelegate {
 
 }
 
+// MARK: - Deprecations
+
+@available(iOS 15.0, *)
+extension PaywallProxy {
```
NachoSoto, line 220: We can probably remove all these. I don't think we gain anything keeping them. I know it's a public SDK, but I'd be surprised if anyone was actually using these.

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
vegaro: I agree, it makes adding parameters very complicated. I thought about creating an options object instead so we can have just that one parameter and defer the defaults to the natives

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
vegaro: Hmm, yes... https://github.com/RevenueCat/purchases-hybrid-common/pull/666/files#diff-3a3cff3944b3db826f3f47f7e980ca36ad3f9589beac3ba0aabc52f5f5a28b29R105 is also breaking

vegaro (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
tonidero: Right... I think the idea was to try to avoid duplicating the defaults in the hybrids and rely on the ones in the natives... But yeah, it's likely not worth that 👍

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
tonidero: Hmm, I'm not opposed... But in that case, I would probably stop using semver in PHC and use some different versioning system. If we are using semver and making a breaking change, I think we should still make a major (which is probably ok since it doesn't force us to make a major in the hybrids)

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
vegaro: actually, that way we guarantee both platforms default to the same, because right now the default in iOS for `displayCloseButton` is false and for Android true

vegaro:
@NachoSoto I've rebased this since I need to root view controller fix (I am pointing locally to this branch)

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
NachoSoto: Yeah that's what I was thinking too.

ios/PurchasesHybridCommon/PurchasesHybridCommon/PaywallProxy.swift
```diff
@@ -24,58 +24,45 @@ import UIKit
                                                      result: PaywallResult)] = [:]
 
     @objc
-    public func createPaywallView() -> UIViewController {
-        return UIHostingController(rootView: PaywallView())
+    public func createPaywallView() -> PaywallViewController {
```
NachoSoto, line 27: @vegaro I've updated this here too.

NachoSoto:
Accidentally merged on the wrong branch. Reopened on #675.

PR #667, opened 2024-01-22
Fix getting `rootViewController`

vegaro:
As suggested by @MarkVillacampa I changed `windows.first` with `keyWindow`. This was returning null in React Native, which broke the `presentPaywall` and `presentPaywallIfNeeded` functions in 7.16.0

The regression was introduced in #656


MarkVillacampa (approved):

tonidero (approved):

NachoSoto (commented):
Thanks 🙏🏻

PR #669, opened 2024-01-22
Update Android orb

vegaro:
We were using a very old android image (from 2021) which is actually getting deprecated. Updating the version of the orb uses so we use the `default` tag, which automatically gets updated every 3 months with the latest stuff.

Screenshot 2024-01-22 at 14 56 12

https://discuss.circleci.com/t/android-image-deprecations-and-eol-for-2024/50180


tonidero (approved):
🙌

PR #670, opened 2024-01-22
`Paywalls`: add methods for presenting paywalls with an offering identifier (Android)

vegaro:
Added `newInstance` function that accepts an `offeringIdentifier` instead of an `Offering`

### DIFF
--- a/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
+++ b/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
@@ -26,24 +26,7 @@ internal class PaywallFragment : Fragment(), PaywallResultHandler {
             requiredEntitlementIdentifier: String? = null,
             paywallResultListener: PaywallResultListener? = null,
             shouldDisplayDismissButton: Boolean? = null,
-            offering: Offering? = null,
-        ): PaywallFragment {
-            return newInstance(
-                activity,
-                requiredEntitlementIdentifier,
-                paywallResultListener,
-                shouldDisplayDismissButton,
-                offering?.identifier,
-            )
-        }
-
-        @JvmStatic
-        fun newInstance(
-            activity: FragmentActivity,
-            requiredEntitlementIdentifier: String? = null,
-            paywallResultListener: PaywallResultListener? = null,
-            shouldDisplayDismissButton: Boolean? = null,
-            offeringIdentifier: String? = null,
+            paywallSource: PaywallSource,
         ): PaywallFragment {
             val paywallFragmentViewModel = ViewModelProvider(activity)[PaywallFragmentViewModel::class.java]
             paywallFragmentViewModel.paywallResultListener = paywallResultListener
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -223,6 +225,7 @@ jobs:
     executor:
       name: android/android-machine
       resource-class: large
+      tag: default
```
tonidero, line 228: This is from #669 right? Also, seems that this PR might need a rebase since it has some unrelated changes.

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
```diff
@@ -27,14 +27,31 @@ internal class PaywallFragment : Fragment(), PaywallResultHandler {
...
+            )
+        }
+
+        @JvmStatic
+        fun newInstance(
+            activity: FragmentActivity,
+            requiredEntitlementIdentifier: String? = null,
+            paywallResultListener: PaywallResultListener? = null,
+            shouldDisplayDismissButton: Boolean? = null,
+            offeringIdentifier: String? = null,
```
tonidero, line 46: I wonder if we should deprecate the old one (and remove it?), I don't think we will be using it since it would be much trickier.

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallHelpers.kt
```diff
@@ -4,25 +4,67 @@ import androidx.fragment.app.FragmentActivity
...
+            )
+            .commit()
+    }
+}
+
+data class PresentPaywallOptions(
+    val requiredEntitlementIdentifier: String? = null,
+    val paywallResultListener: PaywallResultListener? = null,
+    val shouldDisplayDismissButton: Boolean? = null,
+    val offering: Offering? = null,
```
tonidero, line 68: Hmm could we just entirely remove this one in favor of `offeringIdentifier`? If it's for the deprecated method, we can just pass in the id there I think.

tonidero (commented):
Left a couple of comments but it's looking good

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallHelpers.kt
NachoSoto: Wouldn't that mean we have to fetch offerings again potentially? 

Can we define a type like this:
```kotlin
sealed class PaywallSource {
    class Offering(val value: PurchasesOffering): PaywallSource()
    class DefaultOffering()
    class OfferingIdentifier(val value: String): PaywallSource()
}
```

Like I did in iOS, to make the 3 options type safe?

tonidero: Right... Yeah makes sense. We have something similar as an internal API in native android: https://github.com/RevenueCat/purchases-android/blob/main/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt#L6... But I don't know if we want to expose that. But yeah, in here that could make sense.

vegaro: > Wouldn't that mean we have to fetch offerings again potentially?

The hybrids are always going to use the offering identifier since there's no way to send a whole offering from JS, Dart, etc to the native side

.circleci/config.yml
vegaro: sorry, I had the wrong base branch, these are gone now

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallHelpers.kt
vegaro: But I think I am going to add the type just in case

vegaro: Yes that's what I'm doing in the opened PR, but this helper in phc will always get an offering identifier, it will never be an offering

NachoSoto: Yup yup 👍🏻

NachoSoto (approved):

NachoSoto:
Merging this, I hope that's ok 😇 
I want to make a release with this + #675 before releasing #672.

PR #672, opened 2024-01-22
Separate `PurchasesHybridCommonUI`

NachoSoto:
For https://github.com/RevenueCat/purchases-flutter/issues/939.


PurchasesHybridCommon.podspec
```diff
@@ -17,7 +17,6 @@ Pod::Spec.new do |s|
   s.framework      = 'SwiftUI'
 
   s.dependency 'RevenueCat', '4.32.0'
-  s.dependency 'RevenueCatUI', '4.32.0'
```
NachoSoto, line 20: This is the core performance improvement.

tonidero (approved):

.circleci/config.yml
```diff
@@ -171,16 +171,28 @@ jobs:
...
           cache-version: v1
       - trust-github-key
       - run:
-          name: Deploy new version to Cocoapods
-          command: bundle exec fastlane ios push_pods
+          name: Deploy PurchasesHybridCommon to Cocoapods
+          command: bundle exec fastlane ios push_pod_PHC
+          no_output_timeout: 30m
+
+  deploy-ios-phcui:
```
NachoSoto, line 186: This needs to depend on the first one

PurchasesHybridCommonUI.podspec
```diff
@@ -0,0 +1,35 @@
...
+  s.dependency 'RevenueCatUI', '4.32.0'
+  s.swift_version = '5.7'
+
+  s.ios.deployment_target = '11.0'
+  s.osx.deployment_target = '10.13'
+  s.tvos.deployment_target = '11.0'
+
+  s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
+
+  s.source_files = ['ios/PurchasesHybridCommon/PurchasesHybridCommonUI/**/*.{h,m,swift}']
```
tonidero, line 29: I do wonder if it would be better to separate into 2 separate folders inside the `ios` folder, so the separation is clearer... I'm ok with this though

.circleci/config.yml
vegaro: Does this need to wait for the other pod to propagate? I think it will try to build and fail because it won't find the pod

PurchasesHybridCommonUI.podspec
```diff
@@ -0,0 +1,35 @@
...
+  s.license          =  { :type => 'MIT' }
+  s.author           = { "RevenueCat, Inc." => "support@revenuecat.com" }
+  s.source           = { :git => "https://github.com/revenuecat/purchases-hybrid-common.git", :tag => s.version.to_s }
+  s.documentation_url = "https://docs.revenuecat.com/"
+
+  s.framework      = 'StoreKit'
+  s.framework      = 'SwiftUI'
+
+  s.dependency 'RevenueCat', '4.32.0'
+  s.dependency 'RevenueCatUI', '4.32.0'
```
vegaro, line 20: Does this need both defined? Or will the `RevenueCat` get to the project transitively

fastlane/Fastfile
vegaro, line 1: `files_with_version_number` and `files_to_update_ios_dependency` also need to be updated at the top of the file

PurchasesHybridCommonUI.podspec
NachoSoto: Good point, let me clean this

.circleci/config.yml
NachoSoto: Yup, updated.

fastlane/Fastfile
NachoSoto: Done.

PurchasesHybridCommonUI.podspec
NachoSoto: Actually `ios/PurchasesHybridCommon` is where the Xcode project is, and since `PurchasesHybridCommonUI` is still part of the same project (just a different target), it makes sense that it's inside.

PR #685, opened 2024-01-25
`CommonFunctionality`: `purchase` methods also return `StoreTransaction`

NachoSoto:
This will be used by some hybrids, like for https://github.com/RevenueCat/react-native-purchases/issues/659.

PR #686, opened 2024-01-25
`Common`: purchase methods also return `StoreTransaction`

NachoSoto:
Android version of #685.


android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/StoreTransactionMapper.kt
```diff
@@ -0,0 +1,12 @@
+package com.revenuecat.purchases.hybridcommon.mappers
+
+import com.revenuecat.purchases.models.StoreTransaction
+import java.util.Date
+
+fun StoreTransaction.map(): Map<String, Any?> =
```
NachoSoto, line 6: We didn't have this, we just had `Transaction.map()`

tonidero (approved):
Looks great!

PR #724, opened 2024-02-09
Set delegate for PaywallViewController

vegaro:
We were not setting the delegate for `PaywallViewController`


tonidero (approved):
Ah right, looks good!

NachoSoto (commented):
👍🏻

PR #725, opened 2024-02-09
Add `paywallViewControllerDidStartPurchase` to wrapper delegate

vegaro:
Added new `paywallViewControllerDidStartPurchase`

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
@@ -183,6 +183,10 @@ import UIKit
 
 @available(iOS 15.0, *)
 extension PaywallProxy: PaywallViewControllerDelegate {
+    
+    public func paywallViewControllerDidStartPurchase(_ controller: PaywallViewController) {
+        self.delegate?.paywallViewControllerDidStartPurchase?(controller)
+    }
 
     public func paywallViewController(_ controller: PaywallViewController,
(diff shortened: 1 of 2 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
```diff
@@ -183,6 +183,10 @@ import UIKit
 
 @available(iOS 15.0, *)
 extension PaywallProxy: PaywallViewControllerDelegate {
+    
+    public func paywallViewControllerDidStartPurchase(_ controller: PaywallViewController) {
```
tonidero, line 187: Do we also need one for restore starting?

tonidero (approved):
This looks good! Just a question

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
vegaro: yes, but that one is not available in purchases-ios yet so we need to add it there first 😢

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
```diff
@@ -183,6 +183,10 @@ import UIKit
...
+        self.delegate?.paywallViewControllerDidStartPurchase?(controller)
```
NachoSoto, line 188: Thanks for adding this!

PR #729, opened 2024-02-09
Add lanes to trigger dependency updates and bump

vegaro:
Created two lanes, one that triggers a dependency update and one that creates a bump. Inspired by https://github.com/RevenueCat/purchases-ios/blob/b99c8a6f6392038f26c68240a87c50519f053316/fastlane/Fastfile#L458

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -91,6 +91,16 @@ lane :github_release do |options|
   )
 end
 
+desc "Trigger update dependencies to latest GitHub releases"
+lane :trigger_update_dependencies do
+    trigger_action(action: "dependency-update")
+end
+
+desc "Trigger bump"
+lane :trigger_bump do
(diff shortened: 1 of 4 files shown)


fastlane/Fastfile
```diff
@@ -204,6 +214,32 @@ lane :tag_current_branch do |options|
   push_git_tags(tag: version_number)
 end
 
+private_lane :trigger_action do |options|
```
vegaro, line 217: We can move this to the common fastlane plugin, but since I had already added it here, I figured we can have it here until its moved over there, which takes time due to unit tests and such

tonidero: Hmm yeah, I think this would ideally be there, since it seems like something we might want to reuse... But I'm ok with this for now.

tonidero (approved):

PR #735, opened 2024-02-16
Add `didStartPurchaseWithPackage`

vegaro:
Needed for https://github.com/RevenueCat/react-native-purchases/pull/911

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
@@ -9,8 +9,12 @@
 import Foundation
 import RevenueCat
 
-internal extension Package {
+public extension Package {
 
+    var dictionary: [String: Any] {
+        return dictionary(offeringIdentifier)
+    }
+    
     func dictionary(_ offeringIdentifier: String) ->  [String: Any] {
         return [
             "identifier": identifier,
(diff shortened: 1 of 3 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
```diff
@@ -9,8 +9,12 @@
 import Foundation
 import RevenueCat
 
-internal extension Package {
+public extension Package {
```
tonidero, line 12: Do we need to access these from the hybrids now?

tonidero (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
vegaro: It's from `PurchasesHybridCommonUI` https://github.com/RevenueCat/purchases-hybrid-common/pull/735/files#diff-c673e3957e7bea968ef2fa452b8c46a98aac22c132cf33e54ab4ba6bbb02f025R200

vegaro: for example, CustomerInfo and Offerings are also public for the same reason :)

tonidero: Ohh right, I forgot 😅 Makes sense, thanks!

PR #741, opened 2024-02-23
[WIP] Release 10.0.0

MarkVillacampa:
Version 10.0.0 of `purchases-hybrid-common` bumps the iOS SDK to 5.0.0.

- Removed all iOS 13 and equivalent availability checks since we bumped the minimum target.
- Exposed new APIs: StoreKitVersion and SK2 observer mode.
- Update Typescript APIs to use PERIOD_UNIT type instead of string: https://github.com/RevenueCat/purchases-hybrid-common/pull/797

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
@@ -41,6 +41,8 @@
 		35F6FD602673FE5600ABCB53 /* ErrorContainerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 35F6FD5F2673FE5600ABCB53 /* ErrorContainerTests.swift */; };
 		37E357D5014169A093450AD5 /* MockPurchases.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37E351F858AE15D5BD2EAA5A /* MockPurchases.swift */; };
 		3F7F287D27E208F39745436C /* Pods_ObjCAPITester.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 4FAA49047A2AE8963CF73C94 /* Pods_ObjCAPITester.framework */; };
+		4D322E642B83BC650004AF53 /* StoreKitVersion+HybridAdditions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4D322E632B83BC650004AF53 /* StoreKitVersion+HybridAdditions.swift */; };
+		4D6276F42B8A0AD30062E553 /* StoreKitVersion+HybridAdditionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4D6276F32B8A0AD30062E553 /* StoreKitVersion+HybridAdditionsTests.swift */; };
 		4F7974562B5EDDEF003085E7 /* PaywallProxy.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4F72072B2AF310390017395F /* PaywallProxy.swift */; };
 		4F79745C2B5EDDEF003085E7 /* PaywallResult.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1EFF59942B46D1450022FD77 /* PaywallResult.swift */; };
 		4F79746A2B5EDF82003085E7 /* PurchasesHybridCommon.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2D54BF002437AD5800FF4EE4 /* PurchasesHybridCommon.framework */; };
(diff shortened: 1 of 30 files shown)


vegaro:
This is using a beta version of iOS, shouldn't it be beta as well? Not sure if we've talked about it somewhere else and I missed the conversation

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -654,3 +640,26 @@ private extension CommonFunctionality {
...
+            let result = await StoreKit.Transaction.latest(for: productID)
+            if let result = result {
+                do {
+                    let transaction = try await Self.sharedInstance.handleObserverModeTransaction(.success(result))
+                    completion?(transaction?.dictionary, nil)
+                } catch {
+                    completion?(nil, ErrorContainer(error: error, extraPayload: [:]))
+                }
+            } else {
+                completion?(nil, nil)
```
MarkVillacampa, line 660: We could either return an error here (but I couldn't find any existing error code we could reuse, and I'm not sure this warrants adding a new one?) or return nil but log a message. Thoughts?

ios/PurchasesHybridCommon/PurchasesHybridCommon/NSDate+HybridAdditions.swift
```diff
@@ -8,8 +8,6 @@
 
 import Foundation
 
-// note: This extension is only temporary, since we're
-// dropping iOS 11 in the upcoming major
```
MarkVillacampa, line 12: We already replaced the implementation with `ISO8601DateFormatter` and dropped iOS 11, and there's no need to remove this extensions since they're useful.

https://github.com/RevenueCat/purchases-hybrid-common/commit/dacb949cb7db73560fd057deea6026b727aa8717

ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
```diff
@@ -1047,13 +1055,12 @@
...
 				INFOPLIST_KEY_NSHumanReadableCopyright = "Copyright © 2022 RevenueCat. All rights reserved.";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
-				MACOSX_DEPLOYMENT_TARGET = 10.12;
```
tonidero, line 1056: Is it ok to remove these?

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
tonidero: In what situations could this happen? (result being nil). 

I think ideally we return either success or error. Since it's not a success (unless I'm missing something) I think we should try to return an error... Could we return an unknown error or some store error with some extra info?

typescript/src/enums.ts
```diff
@@ -184,3 +184,25 @@ export enum PAYWALL_RESULT {
...
+   * Always use StoreKit 2 (StoreKit 1 will be used if StoreKit 2 is not available in the current device.)
+   * - Warning: Make sure you have an In-App Purchase Key configured in your app.
+   * Please see https://rev.cat/in-app-purchase-key-configuration for more info.
+   */
+  STOREKIT_2 = "STOREKIT_2",
+
+  /**
+   * Let RevenueCat use the most appropiate version of StoreKit
+  */
+ DEFAULT = "DEFAULT",
```
tonidero, line 207: The indentation here seems wrong?

tonidero (commented):
I don't see the bump to 10.0.0. I guess that will be done in a separate PR?

In any case, this is looking good. Just left some comments

ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
MarkVillacampa: Yep, I moved the deployment target requirements to the project level instead of the target level.

MarkVillacampa:
Yes, the bump to 10.0.0-beta.1 will happen in a separate release PR cut from this branch

tonidero:
Hmm I just noticed, I think we didn't include 
> Update Typescript APIs to use PERIOD_UNIT type instead of string: https://github.com/RevenueCat/purchases-hybrid-common/pull/797

In the latest major right?... I guess we should leave it for the next one :( But this PR should be good to close otherwise?

aboedo:
@MarkVillacampa I'm gonna guess this PR is safe to kill?

PR #746, opened 2024-02-28
Add `paywallViewControllerRequestedDismissal`

vegaro:
Please read https://github.com/RevenueCat/react-native-purchases/pull/913

Depends on https://github.com/RevenueCat/purchases-ios/pull/3738

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
@@ -27,7 +27,7 @@ import UIKit
     public func createPaywallView() -> PaywallViewController {
         let controller = PaywallViewController()
         controller.delegate = self
-        controller.updateAutomaticallyDismiss(with: false)
+        controller.requestedDismissalDelegate = self
         return controller
     }
 
@@ -38,7 +38,7 @@ import UIKit
     ) -> PaywallViewController {
         let controller = PaywallViewController(offeringIdentifier: offeringIdentifier)
         controller.delegate = self
-        controller.updateAutomaticallyDismiss(with: false)
+        controller.requestedDismissalDelegate = self
         return controller
     }
 
@@ -248,13 +248,18 @@ extension PaywallProxy: PaywallViewControllerDelegate {
         paywallResultHandler(result.name)
     }
 
-    public func paywallViewControllerRequestedDismissal(_ controller: PaywallViewController) {
-        self.delegate?.paywallViewControllerRequestedDismissal?(controller)
-    }
-
     public func paywallViewController(_ controller: PaywallViewController,
                                       didChangeSizeTo size: CGSize) {
         self.delegate?.paywallViewController?(controller, didChangeSizeTo: size)
+    }
+
+}
+
+@available(iOS 15.0, *)
+extension PaywallProxy : PaywallViewControllerRequestedDismissalDelegate {
+
+    public func paywallViewControllerRequestedDismissal(_ controller: PaywallViewController) {
+        self.delegate?.paywallViewControllerRequestedDismissal?(controller)
     }
 
 }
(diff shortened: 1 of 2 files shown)


tonidero (commented):
Changes LGTM! Lmk when this is ready for review

vegaro:
@tonidero I am waiting for https://github.com/RevenueCat/purchases-ios/pull/3738, which needs to be released before this can be merged

tonidero (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
```diff
@@ -28,15 +27,18 @@ import UIKit
...
     @objc
-    public func createPaywallView(offeringIdentifier: String) -> PaywallViewController {
+    public func createPaywallView(
+        offeringIdentifier: String,
+        dismissHandler: (() -> Void)? = nil
+    ) -> PaywallViewController {
         let controller = PaywallViewController(offeringIdentifier: offeringIdentifier)
         controller.delegate = self
-
+        controller.dismissalDelegate = self
```
tonidero, line 41: We need to set this delegate for the footer methods as well. Otherwise it won't be called after a purchase completes

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
```diff
@@ -26,17 +25,23 @@ import UIKit
...
 
     @objc
-    public func createPaywallView(offeringIdentifier: String) -> PaywallViewController {
-        let controller = PaywallViewController(offeringIdentifier: offeringIdentifier)
+    public func createPaywallView(
+        offeringIdentifier: String,
+        dismissHandler: (() -> Void)? = nil
+    ) -> PaywallViewController {
+        let controller = PaywallViewController(offeringIdentifier: offeringIdentifier,
+                                               dismissalRequest: { controller in
```
tonidero, line 41: We need to update the name of this parameter and add it to the footer methods... Maybe we can even extract to a function to avoid repeating the code?

tonidero: Though thinking about it, we might need to make `self` weak here I think?

tonidero (approved):
LGTM!

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
```diff
@@ -26,31 +25,42 @@ import UIKit
 
     @objc
     public func createPaywallView() -> PaywallViewController {
-        let controller = PaywallViewController()
+        let controller = PaywallViewController(dismissRequestedHandler: { controller in
+            self.delegate?.paywallViewControllerRequestedDismissal?(controller)
```
tonidero, line 29: As mentioned, I think we might want to weakify `self` here?

vegaro: I think so yes

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
```diff
@@ -185,6 +184,13 @@ import UIKit
 
     }
 
+    private func createDismissHandler() -> (PaywallViewController) -> Void {
+        return { [weak self] controller in
+            guard let delegate = self?.delegate else { return }
```
tonidero, line 189: Hmm thinking about it, we could also hold just the delegate (weakly) instead of self... But this should be fine.

tonidero (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/PaywallProxy.swift
vegaro: actually the delegate is already weak

    public weak var delegate: PaywallViewControllerDelegateWrapper?

vegaro: but yes, let's leave it like this, I tested it just in case and it works fine.

PR #748, opened 2024-03-01
Add targeting placements and custom attributes

joshdholtz:
## Motivation

Add support for new targeting features
- Placements
- Custom attributes

## Description

‼️ This is a breaking change PR (replacing offering identifier with offering context in the calls)

- Bumps iOS to `4.38.0`
- Bumps Android to `7.70`

### PresentedOfferingContext
- Android
  - Adds to `purchaseProduct`
  - Adds to `purchasePackage`
  - Adds to `purchaseSubscriptionOption`
- iOS
  - Adds to `purchasePackage`

### Placements

- Android
  - `getCurrentOfferingForPlacement()`
- iOS
  - `getCurrentOffering()`

### Custom Attributes

- Android
  - `syncAttributesAndOfferingsIfNeeded()`
- iOS
  - `syncAttributesAndOfferingsIfNeeded()`


typescript/src/offerings.ts
```diff
@@ -1,617 +1,651 @@
...
+  readonly defaultOption: SubscriptionOption | null;
+  /**
+   * Collection of subscription options for a product. Google Play only.
+   */
+  readonly subscriptionOptions: SubscriptionOption[] | null;
+  /**
+   * Offering identifier the store product was presented from.
+   * Null if not using offerings or if fetched directly from store via getProducts.
+   */
+  readonly presentedOfferingIdentifier: string | null;
```
tonidero, line 132: Should we deprecate this one?

typescript/src/offerings.ts
```diff
@@ -1,617 +1,651 @@
...
+   */
+  readonly packageType: PACKAGE_TYPE;
+  /**
+   * Product assigned to this package.
+   */
+  readonly product: PurchasesStoreProduct;
+  /**
+   * Offering this package belongs to.
+   */
+  readonly offeringIdentifier: string;
```
tonidero, line 267: Same here about maybe deprecating this.

typescript/src/offerings.ts
```diff
@@ -1,617 +1,651 @@
...
+   * The intro trial PricingPhase of the subscription.
+   * Looks for the first pricing phase of the SubscriptionOption where amountMicros is greater than 0.
+   * There can be a freeTrialPhase and an introductoryPhase in the same SubscriptionOption.
+   */
+  readonly introPhase: PricingPhase | null;
+
+  /**
+   * Offering identifier the subscription option was presented from
+   */
+  readonly presentedOfferingIdentifier: string | null;
```
tonidero, line 519: Same here

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -283,14 +283,19 @@ import RevenueCat
         }
     }
 
-    @objc(purchasePackage:offering:signedDiscountTimestamp:completionBlock:)
+    @objc(purchasePackage:offering:presentedOfferingContext:signedDiscountTimestamp:completionBlock:)
     static func purchase(package packageIdentifier: String,
                          offeringIdentifier: String,
```
tonidero, line 288: This would be a breaking change but should we remove the `offeringIdentifier` parameter? FWIW, we decided we would be ok making majors in PHC as often as needed, since it's mostly meant for internal usage.

We can also do this in a separate PR.

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -586,12 +618,44 @@ private extension CommonFunctionality {
...
                         offeringIdentifier: String,
+                        presentedOfferingContext: PresentedOfferingContext?,
                         completion: @escaping(Package?) -> Void) {
         Self.sharedInstance.getOfferings { offerings, error in
             let offering = offerings?.offering(identifier: offeringIdentifier)
-            let package = offering?.package(identifier: packageIdentifier)
+            let foundPackage = offering?.package(identifier: packageIdentifier)
+
+            let package = foundPackage.flatMap { pkg in
+                Package(
```
tonidero, line 651: I wonder if we should change the completion block in this function to return a map instead of a `Package` to avoid having to recreate the package here... We can go with this for now though.

ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
```diff
@@ -9,6 +9,21 @@
...
+
+    var dictionary: [String: Any] {
+        let dict: [String: Any?] = [
+            "offeringIdentifier": self.offeringIdentifier,
+            "placementIdentifier": self.placementIdentifier,
+            "targetingRevision": self.targetingContext?.revision,
+            "targetingRuleId": self.targetingContext?.ruleId,
+        ]
+
+        return dict.compactMapValues { $0 }
```
tonidero, line 22: Since we are using `compactMapValues`, that means we won't have keys with null values right? My main concerns are:
- It's a slight difference in behavior from Android where we can have these keys with null values
- In the typescript code, we've set the types to `PresentedOfferingContext | null`, but do not support `undefined` which I think would be what would happen if we strip null key/values... 
I think we should either not compact the values here or fix it in android and typescript

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/CommonFunctionality+async.swift
```diff
@@ -19,6 +32,23 @@ extension CommonFunctionality {
         }
     }
 
+    static func currentOffering(forPlacement placement: String) async throws -> [String: Any]? {
+        let value: PlacementOffering = try await withCheckedThrowingContinuation { continuation in
+            Self.getCurrentOffering(forPlacement: placement) { dictionary, error in
+                let v: PlacementOffering
```
tonidero, line 38: I'm not fully understanding this but maybe it's my swift 😅. Why is it needed to map the dictionary back to a `PlacementOffering` before sending it as part of the `Result`?

typescript/src/offerings.ts
```diff
@@ -1,617 +1,651 @@
...
+   */
+  readonly placementIdentifier: string | null;
+  /**
+   * The revision of the targeting used to obtain this object.
+   */
+  readonly targetingRevision: string | null;
+  /**
+   * The rule id from the targeting used to obtain this object.
+   */
+  readonly targetingRuleId: string | null;
```
tonidero, line 294: Note that we're degrouping the `targetingRevision` and `targetingRuleId` here (in the natives, they are part of another object. I'm not sure if it's worth introducing another type for this though... I'm ok keeping it like this.

tonidero (commented):
Left some comments but looking good!

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/CommonFunctionality+async.swift
joshdholtz: This got me too 😅 An error gets thrown if a `Result` is created like `Result(nil, nil)` which _is_ what we want with an offerings by placement because `nil` is a valid value. So to get _around_ that, we need to map a `nil` to temporary enum to pass into the `Result` and then map that back for the return. It's... a weird Swift thing 😬

ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
joshdholtz: Ahhhhh, good point good point! I'll remove the compact map 👍

typescript/src/offerings.ts
joshdholtz: Yeah yeah, this crossed my mind ☝️ It kind of gets messy either direction... I'll do a little POC and see what it takes because _maybe_ more structure if better if we end up adding more things to the context in the future 🤔

joshdholtz: DONE! 💪

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -586,12 +616,50 @@ private extension CommonFunctionality {
...
+            targetingContext: targetingContext
+        )
+    }
+
     static func package(withIdentifier packageIdentifier: String,
-                        offeringIdentifier: String,
+                        presentedOfferingContext: PresentedOfferingContext?,
                         completion: @escaping(Package?) -> Void) {
+        guard let presentedOfferingContext else {
+            return completion(nil)
```
tonidero, line 647: Hmm should we print something in this case?

ios/PurchasesHybridCommon/PurchasesHybridCommon/Package+HybridAdditions.swift
```diff
@@ -9,6 +9,29 @@
 import Foundation
 import RevenueCat
 
+public extension PresentedOfferingContext {
+
+    var dictionary: [String: Any] {
+        return [
+            "offeringIdentifier": self.offeringIdentifier,
+            "placementIdentifier": self.placementIdentifier  ?? NSNull(),
```
tonidero, line 17:
```suggestion
            "placementIdentifier": self.placementIdentifier ?? NSNull(),
```

tonidero (approved):
LGTM!

PR #754, opened 2024-03-07
Release 6.3.2

vegaro:
### Dependency Updates
* Bump android to 6.9.7


tonidero (approved):

typescript/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@revenuecat/purchases-typescript-internal",
   "title": "Purchases typescript internal shared code",
-  "version": "6.3.1",
+  "version": "6.3.2",
```
tonidero, line 4: I believe this deployment won't have the fix so the hotfix is not deployed as the latest right? We would need to retag the latest version after deploying this

vegaro: just did this

PR #792, opened 2024-04-17
fix(PaywallProxy): check if root view controller is already presenting

MegaMaddin:
In case our root view controller is already presenting another view controller, go to the top of the "view controller food chain" and present the paywall there.

This will fix the issue, where it was not possible to present a paywall on top of a modal presentation style view controller.

I previously discussed this issue with Michael from the developer support team. I think there should be an issue in Zendesk about this (id39622).


aboedo:
Thanks for opening this up @MegaMaddin! It makes a lot of sense, we'll get this merged 💪

tonidero (approved):
Agreed, thanks for doing this!! Will get this merged now

PR #796, opened 2024-04-19
Add api-extractor to automate API tests

tonidero:
This adds a new dev dependency `api-extractor` to automatically generate files containing our entire public API, eliminating the need to add api tests manually.

In the process, I had to fix some documentation that doesn't follow the typedoc standard.


.circleci/config.yml
```diff
@@ -378,7 +378,10 @@ jobs:
       - node/install:
           install-yarn: true
       - run:
-          command: yarn && yarn build && yarn apitests
+          command: yarn && yarn build && yarn extract-api
+      - run:
+          name: Check if API has changed
+          command: git diff --exit-code api-report/**
```
tonidero, line 384: The idea is to have the typescript files act as a kind of "snapshot" test, so if it changes, we need to manually update them by running `yarn extract-api` and commiting the difference.

typescript/src/callbackTypes.ts
```diff
@@ -7,16 +7,43 @@ import {
...
 * Listener used on updated customer info
-* @callback CustomerInfoUpdateListener
-* @param {Object} customerInfo Object containing info for the customer
+* @public
+* @param customerInfo - Object containing info for the customer
 */
 export type CustomerInfoUpdateListener = (customerInfo: CustomerInfo) => void;
+/**
+ * Listener used to determine if a user should be able to purchase a promo product.
+ * @public
```
tonidero, line 16: I had to add some missing docs and visibility modifiers to all APIs

typescript/src/callbackTypes.ts
```diff
@@ -7,16 +7,43 @@ import {
...
+   */
+  productIdentifier: string;
+  /**
+   * The Customer Info for the user.
+   */
+  customerInfo: CustomerInfo;
+  /**
+   * The transaction object for the purchase
+   */
+  transaction: PurchasesStoreTransaction;
```
tonidero, line 35: I wanted to make these `readonly`, but that's a breaking change...

typescript/src/customerInfo.ts
```diff
@@ -65,31 +66,31 @@ export interface PurchasesEntitlementInfo {
     /**
      * The date an unsubscribe was detected in ISO8601 format. Can be `null`.
      *
-     * @note: Entitlement may still be active even if user has unsubscribed. Check the `isActive` property.
```
tonidero, line 68: `@note` is not accepted in TSDoc: https://tsdoc.org/

typescript/src/customerInfo.ts
```diff
@@ -179,15 +184,17 @@ export interface CustomerInfo {
...
+    /**
+     * List of all non subscription transactions. Use this to fetch the history of
+     * non-subscription purchases
+     */
     readonly nonSubscriptionTransactions: PurchasesStoreTransaction[];
 }
 
 /**
- * List of all non subscription transactions. Use this to fetch the history of
- * non-subscription purchases
```
tonidero, line 188: This was in the wrong place. Moved it up to where it should be.

typescript/src/offerings.ts
```diff
@@ -426,6 +466,10 @@ export interface PurchasesPromotionalOffer {
   readonly timestamp: number;
 }
 
+/**
+ * Enum with possible proration modes in a subscription upgrade or downgrade in the Play Store. Used only for Google.
+ * @public
```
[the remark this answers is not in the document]
tonidero: Well, these are actually warnings the library mentions. Basically this is only really useful if you use this tool to also generate docs (we don't generate docs for this library, and we use `tsdoc`/`@capacitor/docgen` in our other libraries). So we could just disable the warning... But I thought it would be good to follow the `TSDoc` standard, instead of just silencing the warning, but lmk if you feel strongly about it, and I would be ok to change it.

typescript/api-report/purchases-typescript-internal.api.md
```diff
@@ -0,0 +1,521 @@
+## API Report File for "@revenuecat/purchases-typescript-internal"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
```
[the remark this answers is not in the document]
tonidero: Done in https://github.com/RevenueCat/purchases-hybrid-common/pull/796/commits/9e9a0d69d62a49198eb395a98b9f7c3bf5fd3ccb

Issue #786, opened 2024-04-10
Typescript: `periodUnit` should be `PERIOD_UNIT`


christophby:
### The type of `periodUnit` should be more strict.

In [offerings.ts](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/typescript/src/offerings.ts) the key `periodUnit` should have the type **`PERIOD_UNIT`** instead of `string`. 

The [comment above](https://github.com/RevenueCat/purchases-hybrid-common/blob/main/typescript/src/offerings.ts#L212) already says
> Unit for the billing period of the discount, can be DAY, WEEK, MONTH or YEAR.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/39454) and notified the team. Thank you for reporting, we're checking this out!

mshmoustafa:
@christophby thanks for reporting! I opened an internal ticket for the engineering team to work on this.

PR #797, opened 2024-04-19
Update `periodUnit` fields to use `PERIOD_UNIT` type

tonidero:
We had a couple types for a periodUnit type using string instead of the existing `PERIOD_UNIT` enum. This enum already has the same underlying string types, but we should be using the enum.

This is a breaking change so will be part of the next major.

This will address #786


tonidero:
Failing tests are unrelated (this PR only changes the typescript interfaces which are not covered in the ios tests) and also happen in the base branch. Seems there is a mismatch of versions that we need to fix, but will hold that until we release v5 in iOS.

vegaro (approved):

PR #814, opened 2024-05-07
Add PrivacyInfo.xcprivacy

vegaro:
As reported in https://github.com/RevenueCat/purchases-flutter/issues/1064 we were missing a PrivacyInfo.xcprivacy

Added CA92.1 as reason which reads:

> Declare this reason to access user defaults to read and write information that is only accessible to the app itself.
>
> This reason does not permit reading information that was written by other apps or the system, or writing information that can be accessed by other apps.

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393

### DIFF
--- a/PurchasesHybridCommon.podspec
+++ b/PurchasesHybridCommon.podspec
@@ -31,4 +31,6 @@ Pod::Spec.new do |s|
     'ios/PurchasesHybridCommon/PurchasesHybridCommon/*.h'
   ]
 
+  s.resource_bundles = {'PurchasesHybridCommon' => ['ios/PurchasesHybridCommon/PurchasesHybridCommon/PrivacyInfo.xcprivacy']}
+
 end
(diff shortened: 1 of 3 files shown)


PurchasesHybridCommon.podspec
```diff
@@ -31,4 +31,6 @@ Pod::Spec.new do |s|
     'ios/PurchasesHybridCommon/PurchasesHybridCommon/*.h'
   ]
 
+  s.resource_bundles = {'PurchasesHybridCommon' => ['ios/PurchasesHybridCommon/PurchasesHybridCommon/PrivacyInfo.xcprivacy']}
```
tonidero, line 34: Hmm is this not added transitively from `purchases-ios`?

tonidero: Good point yeah

tonidero (approved):

PR #817, opened 2024-05-15
Add support for new OneSignal `setOnesignalUserID` method for OneSignal 11+

tonidero:
This adds support for the new method to set the user id when using Onesignal SDK 11+

### DIFF
--- a/ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
+++ b/ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
@@ -151,6 +151,8 @@ NS_ASSUME_NONNULL_BEGIN
     [RCCommonFunctionality setMparticleID:nil];
     [RCCommonFunctionality setOnesignalID:@""];
     [RCCommonFunctionality setOnesignalID:nil];
+    [RCCommonFunctionality setOnesignalUserID:@""];
+    [RCCommonFunctionality setOnesignalUserID:nil];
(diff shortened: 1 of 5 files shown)


ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
```diff
@@ -151,6 +151,8 @@ - (void)testSubscriberAttributes {
     [RCCommonFunctionality setMparticleID:nil];
     [RCCommonFunctionality setOnesignalID:@""];
     [RCCommonFunctionality setOnesignalID:nil];
+    [RCCommonFunctionality setOnesignalUserID:@""];
+    [RCCommonFunctionality setOnesignalUserID:nil];
```
jamesrb1, line 155: Why this pattern?

jamesrb1 (approved):
LGTM, just a question!

ios/PurchasesHybridCommon/ObjCAPITester/RCHybridCommonAPITest.m
tonidero: These API tests basically make sure that the APIs don't change inadvertently. We only perform some basic verification (compile them) right now. So these are basically making sure that there is a `setOnesignalUserID` method that accepts a string or nil.

This new method is needed when using OneSignal SDK 11+, when using versions <=9, developers should still use the existing methods.

PR #838, opened 2024-06-19
Exposes PHC version

JayShortway:
## Changes
- Adds a new file, `Constants.swift`, which contains the PHC version.
- This file is updated through Fastlane, which is run on CI. 
- Adds a new public property, exposed as function in ObjC, which provides this PHC version. 

## Other approaches I tried
- Reading the `CFBundleShortVersionString` from the `Bundle`, but that doesn’t work because it returns the version of the iOS app, not the PHC framework (even when specifying a PHC class in `Bundle(for: <class>)`).
- Somehow making `PurchasesHybridCommonVersionString` accessible, following [this SO answer](https://stackoverflow.com/a/56543847), but that fails with linker errors.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
@@ -61,10 +61,11 @@
 		57FFDE3F286A777D0089A57B /* StoreKitIntegrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 57FFDE3E286A777D0089A57B /* StoreKitIntegrationTests.swift */; };
 		57FFDE41286A77E30089A57B /* CommonFunctionality+async.swift in Sources */ = {isa = PBXBuildFile; fileRef = 57FFDE40286A77E30089A57B /* CommonFunctionality+async.swift */; };
 		57FFDE46286B606A0089A57B /* XCTestCase+Expectations.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DD3D1842810A81F00A19EC6 /* XCTestCase+Expectations.swift */; };
+		77D5E72A2C233E2C002ECC77 /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77D5E7292C233E2C002ECC77 /* Constants.swift */; };
 		AE0B58267EE48614BBF43A9E /* Pods_PurchasesHybridCommonUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = D02631302E453BBC6C5F5471 /* Pods_PurchasesHybridCommonUI.framework */; };
 		B1C9A7EEA6961CC6DE22BCAD /* Pods_PurchasesHybridCommon_PurchasesHybridCommonTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E05FA8EACFF88F020CF8B9BB /* Pods_PurchasesHybridCommon_PurchasesHybridCommonTests.framework */; };
 		B3105D7A287E3CC200CA935B /* MockSandboxEnvironmentDetector.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3105D79287E3CC200CA935B /* MockSandboxEnvironmentDetector.swift */; };
-		C5335FED2A63D9B5B628774F /* BuildFile in Frameworks */ = {isa = PBXBuildFile; };
+		C5335FED2A63D9B5B628774F /* (null) in Frameworks */ = {isa = PBXBuildFile; };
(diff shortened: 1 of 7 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
```diff
@@ -61,10 +61,11 @@
...
 		57FFDE46286B606A0089A57B /* XCTestCase+Expectations.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DD3D1842810A81F00A19EC6 /* XCTestCase+Expectations.swift */; };
+		77D5E72A2C233E2C002ECC77 /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77D5E7292C233E2C002ECC77 /* Constants.swift */; };
 		AE0B58267EE48614BBF43A9E /* Pods_PurchasesHybridCommonUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = D02631302E453BBC6C5F5471 /* Pods_PurchasesHybridCommonUI.framework */; };
 		B1C9A7EEA6961CC6DE22BCAD /* Pods_PurchasesHybridCommon_PurchasesHybridCommonTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E05FA8EACFF88F020CF8B9BB /* Pods_PurchasesHybridCommon_PurchasesHybridCommonTests.framework */; };
 		B3105D7A287E3CC200CA935B /* MockSandboxEnvironmentDetector.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3105D79287E3CC200CA935B /* MockSandboxEnvironmentDetector.swift */; };
-		C5335FED2A63D9B5B628774F /* BuildFile in Frameworks */ = {isa = PBXBuildFile; };
+		C5335FED2A63D9B5B628774F /* (null) in Frameworks */ = {isa = PBXBuildFile; };
```
tonidero, line 68: Hmm does this mean it's missing a reference to this file? Seems like there might be a bad reference on the `Frameworks` folder but not sure...

ios/PurchasesHybridCommon/PurchasesHybridCommon/Constants.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Constants.swift
+//  PurchasesHybridCommon
+//
+//  Created by Jay Shortway on 19/06/2024.
+//  Copyright © 2024 RevenueCat. All rights reserved.
+//
+
+import Foundation
+
+public struct Constants {
```
tonidero, line 11: Hmm I think we shouldn't make this public, but `internal` and just expose it through the `CommonFunctionality` class right?

tonidero (approved):
LGTM! Just a couple comments

ios/PurchasesHybridCommon/PurchasesHybridCommon/Constants.swift
JayShortway: Very much so yes! Good catch! https://github.com/RevenueCat/purchases-hybrid-common/pull/838/commits/5e44a5751f4c3ed47fac548fc1853ae7274d7ec1

ios/PurchasesHybridCommon/PurchasesHybridCommon.xcodeproj/project.pbxproj
JayShortway: Hmm according to https://github.com/CocoaPods/Xcodeproj/issues/574, this is an orphaned file, since it has no `fileRef`. The other files do, if you look at the lines above this one. 

- If Xcode is the one writing the `project.pbxproj` file, it will write `(null) in Frameworks` for an orphaned file.
- If Xcodeproj is the one writing the `project.pbxproj` file, it will write `BuildFile in Frameworks` for an orphaned file.

So what I think happened here is that the `project.pbxproj` file was last written with Xcodeproj, and on my local machine it was last written with Xcode, causing the diff.

My understanding from that Xcodeproj issue thread is that this was already an orphaned file, and this entire line should just be removed. Let me know if you come to the same or different conclusion!

tonidero: Yup! I think this should be removed, but ideally this happens from the Xcode IDE, which should reference this as a not found file or something like that, since there might be another reference in this project file.

JayShortway: As discussed on Slack, it seems that https://github.com/RevenueCat/purchases-hybrid-common/pull/672 changed the reference to `Pods_PurchasesHybridCommon.framework`, orphaning the previous reference. I have removed the orphaned references. It builds fine locally. Let's see what CI thinks.

JayShortway: CI says :shipit:

PR #843, opened 2024-06-24
Adds .aiexclude

JayShortway:
Adds `.aiexclude` to exclude files with sensitive info from being accessed by Gemini in Android Studio. More info can be read [here](https://developer.android.com/studio/preview/gemini/aiexclude).

Adds it to the `android` folder, as I reckon that's the folder you'll open in Android Studio. Not sure if the `fastlane` sibling folder is necessary to add, but better safe than sorry.

### DIFF
--- a/android/.aiexclude
+++ b/android/.aiexclude
@@ -1,0 +1,2 @@
+../fastlane/.env
+local.properties


tonidero (approved):

vegaro (approved):

tonidero:
@JayShortway going to change the label for this, since I don't think we should consider it a feature for devs using the SDK. Will use `refactor` for now.

JayShortway:
@tonidero Thanks! I was in doubt about the label. 😅 I updated the other PRs too.

PR #858, opened 2024-07-08
adopt purchases-ios 5.2.2

fire-at-will:
This PR updates the purchases-hybrid-common to use purchases-ios 5.2.2.

Specifically, it:
- Updates the purchases-ios dependency to 5.2.2
- Obsoletes CommonFunctionality.setAutomaticAppleSearchAdsAttributionCollection and removes tests related to it
- Replaces the `usesStoreKit2IfAvailable: Bool` params with the `StoreKitVersion` enum
- Introduces the new 5.2.2 enums to the typescript files
- Moves the `PurchasesAreCompletedBy` to/from `String` logic to its own file 
- Bumps the simulator used to run the integration tests to `iPhone 15 Pro (17.2)`

This PR is based heavily off of some previous work by @MarkVillacampa, but ultimately it was easier to create a new branch/PR needed due to some merge conflicts with `main`. 

### Do Not Merge Yet
We're waiting for the latest purchases-ios release to get out the door so we can adopt it instead of 5.0.0


ios/PurchasesHybridCommon/PurchasesHybridCommon/NSDate+HybridAdditions.swift
```diff
@@ -8,8 +8,6 @@
 
 import Foundation
 
-// note: This extension is only temporary, since we're
-// dropping iOS 11 in the upcoming major
```
MarkVillacampa, line 12: Should we do some more cleaning here? or was this comment already outdated?

MarkVillacampa (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
```diff
@@ -36,13 +36,18 @@ import RevenueCat
         if let appUserID = appUserID {
             configurationBuilder = configurationBuilder.with(appUserID: appUserID)
         }
-        configurationBuilder = configurationBuilder.with(purchasesAreCompletedBy: purchasesAreCompletedBy)
+        if let storeKitVersion {
+            if let version = StoreKitVersion(name: storeKitVersion) {
+                configurationBuilder = configurationBuilder.with(
+                    purchasesAreCompletedBy: purchasesAreCompletedBy,
+                    storeKitVersion: version
```
tonidero, line 43: The current implementation means it wouldn't work if the caller only passed a value for `purchasesAreCompletedBy` but passed `nil` for `storeKitVersion`. Can those be separated?

tonidero (approved):
Looks good! Just a couple of comments

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -502,6 +497,32 @@ import RevenueCat
 
 }
 
+// MARK: StoreKit 2 Observer Mode
+@objc public extension CommonFunctionality {
+
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    @objc(handleObserverModeTransactionForProductID:completion:)
+    static func recordPurchase(productID: String, completion: (([String: Any]?, ErrorContainer?) -> Void)?) {
```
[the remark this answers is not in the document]
fire-at-will: Great catch!! I'll rename it to 

```swift
@objc(recordPurchaseForProductID:completion:)
```

ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
fire-at-will: The reason that these are bundled is that they're bundled in the native iOS SDK's configuration options when `purchasesAreCompletedBy` is provided:

```swift
@objc public func with(
            purchasesAreCompletedBy: PurchasesAreCompletedBy,
            storeKitVersion: StoreKitVersion
        ) -> Configuration.Builder {
```

The way it boils down, you can provide a `StoreKitVersion` without `PurchasesAreCompletedBy`, but if you provide a `PurchasesAreCompletedBy`, then you _must_ provide a `StoreKitVersion` 😅

ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
```diff
@@ -12,15 +12,15 @@ import RevenueCat
...
+            platformFlavorVersion:storeKitVersion:dangerousSettings:shouldShowInAppMessagesAutomatically:
             verificationMode:)
     static func configure(apiKey: String,
                           appUserID: String?,
                           purchasesAreCompletedBy: PurchasesAreCompletedBy,
                           userDefaultsSuiteName: String?,
                           platformFlavor: String?,
                           platformFlavorVersion: String?,
-                          usesStoreKit2IfAvailable: Bool = false,
+                          storeKitVersion: String? = "DEFAULT",
```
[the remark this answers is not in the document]
fire-at-will: This is a great point! When I look at it further, `StoreKitVerison.init(name: String)` can't handle a `nil` value, so I'll modify it to be of type `String` instead of `String?`. Good find!

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -502,6 +497,32 @@ import RevenueCat
...
+            let result = await StoreKit.Transaction.latest(for: productID)
+            if let result = result {
+                do {
+                    let transaction = try await Self.sharedInstance.recordPurchase(.success(result))
+                    completion?(transaction?.dictionary, nil)
+                } catch {
+                    completion?(nil, ErrorContainer(error: error, extraPayload: [:]))
+                }
+            } else {
+                completion?(nil, transactionNotFoundError(
```
MarkVillacampa, line 516: Should these completion blocks be dispatched in the main actor?

MarkVillacampa: I see we're not dispatching to the main thread in any of the other methods so we're probably fine for the hybrids.

ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
```diff
@@ -113,22 +118,6 @@ extension LogLevel {
     )
 }
 
-extension PurchasesAreCompletedBy {
```
fire-at-will, line 116: I moved this to a new `PurchasesAreCompletedBy+HybridAdditions.swift` file to remove this responsibility from `Purchases+HybridAdditions`

ios/PurchasesHybridCommon/PurchasesHybridCommon/PurchasesAreCompletedBy+HybridAdditions.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Created by Will Taylor on 7/26/24.
+//  Copyright © 2024 RevenueCat. All rights reserved.
+//
+
+import Foundation
+import RevenueCat
+
+public extension PurchasesAreCompletedBy {
+
+    init?(name: String) {
```
fire-at-will, line 14: I switched the implementation to this over the existing `fromString()` function to match what we're doing in `StoreKitVersion+HybridAdditions` and because it felt a little more "Swift-like" to me - let me know if y'all think we should revert to the previous `fromString()` function!

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/__Snapshots__/StoreKitIntegrationTests/SK2-testCanPurchasePackage.1.json
```diff
@@ -15,7 +15,7 @@
           "isActive" : true,
           "isSandbox" : true,
           "ownershipType" : "PURCHASED",
-          "periodType" : "INTRO",
+          "periodType" : "TRIAL",
```
fire-at-will, line 18: This changed because we were sending up SK1 receipts in iOS v4.X with SK2, and with iOS v5.X we're sending up SK2 JWTs, which treat this as a trial instead of intro

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -17,10 +17,77 @@ import XCTest
...
+
+    override class var purchasesAreCompletedBy: PurchasesAreCompletedBy {
+        return .myApp
+    }
+
+    override class var storeKitVersion: StoreKitVersion { return .storeKit2 }
+
+    private var testSession: SKTestSession!
+
+    override func setUp() async throws {
```
MarkVillacampa, line 34: Should we move this up to `BaseIntegrationTests` to avoid duplication?

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -95,7 +162,7 @@ class StoreKit1IntegrationTests: BaseIntegrationTests {
             try await self.purchaseMonthlyOffering()
             fail("Expected error")
         } catch {
-            expect(error).to(matchError(ErrorCode.invalidPromotionalOfferError))
+            expect(error).to(matchError(ErrorCode.storeProblemError))
```
MarkVillacampa, line 165: 🤔 why did this change?

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
fire-at-will: Done! 😄

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
MarkVillacampa: Turned out to be change in behavior when running in the iOS 17 simualtor

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -87,15 +105,15 @@ class StoreKit1IntegrationTests: BaseIntegrationTests {
         await self.assertSnapshot(data)
     }
 
+    @available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, visionOS 1.0, *)
     func testPurchaseFailuresAreReportedCorrectly() async throws {
-        self.testSession.failTransactionsEnabled = true
-        self.testSession.failureError = .invalidSignature
+        try await self.testSession.setSimulatedError(.purchase(.purchaseNotAllowed), forAPI: .purchase)
```
fire-at-will, line 110: This test has been failing on iOS 17 for a while now since the `failureError` API has been deprecated. After talking with @MarkVillacampa about this offline, we decided to use the new `setSimulatedError()` API here and test for a `purchaseNotAllowedError` instead, since the `setSimulatedError` returns an unknown `StoreKitError` when making a purchase.

MarkVillacampa (approved):

typescript/src/purchasesConfiguration.ts
```diff
@@ -34,16 +38,15 @@ export interface PurchasesConfiguration {
   userDefaultsSuiteName?: string;
   /**
    * iOS-only, will be ignored for Android.
-   * Set this to TRUE to enable StoreKit2.
-   * Default is FALSE.
+   * Default is STOREKIT_2.
```
tonidero, line 41: This is slightly strange considering we have a `DEFAULT` value in the enum... Should we add that the `DEFAULT` is  currently `STOREKIT_2` in the enum?

tonidero (approved):
LGTM! 🙌

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -568,6 +589,20 @@ import RevenueCat
...
+
+    // Obsoleted in purchases-ios 5.0.0
+    @available(iOS, obsoleted: 1, message: "Use enableAdServicesAttributionTokenCollection() instead")
+    @available(watchOS, obsoleted: 1, message: "Use enableAdServicesAttributionTokenCollection() instead")
+    @available(macOS, obsoleted: 1, message: "Use enableAdServicesAttributionTokenCollection() instead")
+    @available(tvOS, obsoleted: 1, message: "Use enableAdServicesAttributionTokenCollection() instead")
+    @available(swift, obsoleted: 1, message: "Use enableAdServicesAttributionTokenCollection() instead")
+    @objc static func setAutomaticAppleSearchAdsAttributionCollection(_ enabled: Bool) {
+        fatalError()
+    }
```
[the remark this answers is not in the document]
fire-at-will: Works for me, I'll remove it!

ios/PurchasesHybridCommon/PurchasesHybridCommonIntegrationTests/BaseIntegrationTests.swift
```diff
@@ -36,7 +57,17 @@ class BaseIntegrationTests: XCTestCase {
...
 
         self.clearReceiptIfExists()
+
+        // Initialize `Purchases` *after* the fresh new session has been created
+        // (and transactions has been cleared), to avoid the SDK posting receipts from
+        // a previous test.
         self.configurePurchases()
+
+        // SDK initialization begins with an initial request to offerings
+        // Which results in a get-create of the initial anonymous user.
```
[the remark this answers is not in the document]
fire-at-will: To be honest, this was copy/pasted from another part of the code, so I was assuming it to be correct. I just dug into the `/v1/subscribers/{SUBSCRIBER_ID}/offerings` endpoint, and while the logic is a bit complex, but it looks like it creates the subscriber if the project is currently running experiments, and doesn't create the subscriber if there are no active experiments.

PR #859, opened 2024-07-09
Update Android to v8 of the native SDK

tonidero:
This updates Android to the latest native version including some needed changes:
- Adds `installmentsInfo` to `SubscriptionOption`
- Adds `pendingTransactionsForPrepaidPlansEnabled` to `configure` method.
- Renames `ProrationMode` to `ReplacementMode`
- Updates int value of `DEFERRED` proration mode to use the new value used in `ReplacementMode`.
- Updates minSdk to 21
- In purchase methods changes `googleProrationMode` parameter to `googleReplacementModeInt`


tonidero:
I've added support for these changes in RN and flutter: https://github.com/RevenueCat/react-native-purchases/pull/1034 and https://github.com/RevenueCat/purchases-flutter/pull/1113/ Will continue with the rest but I think this should be safe to review now

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -538,12 +540,13 @@ fun configure(
     context: Context,
     apiKey: String,
     appUserID: String?,
-    purchasesAreCompletedBy: PurchasesAreCompletedBy? = null,
+    purchasesAreCompletedBy: String? = null,
```
tonidero, line 543: I changed the APIs here and iOS to be strings instead of the enum itself. The reason is that most hybrids require us to pass something from the hybrid layer to the native (a string right now) and then we need to convert that string to the enum in each hybrid. I think it's better to just pass the string and perform the conversion at the PHC level. Lmk if you think otherwise!

typescript/api-report/purchases-typescript-internal.api.md
```diff
@@ -298,7 +310,10 @@ export interface PurchasesConfiguration {
     apiKey: string;
     appUserID?: string | null;
     entitlementVerificationMode?: ENTITLEMENT_VERIFICATION_MODE;
+    // @deprecated
     observerMode?: boolean;
+    pendingTransactionsForPrepaidPlansEnabled?: boolean;
```
tonidero, line 315: Note that I'm keeping the `observerMode` property to avoid the breaking change, but `purchasesAreCompletedBy` has priority if it's set over `observerMode`.

android/hybridcommon/config/detekt/detekt-baseline.xml
```diff
@@ -8,20 +8,13 @@
...
-    <ID>LongParameterList:common.kt$( activity: Activity?, productIdentifier: String, optionIdentifier: String, googleOldProductId: String?, googleProrationMode: Int?, googleIsPersonalizedPrice: Boolean?, presentedOfferingIdentifier: String?, onResult: OnResult, )</ID>
-    <ID>LongParameterList:common.kt$( activity: Activity?, productIdentifier: String, type: String, googleBasePlanId: String?, googleOldProductId: String?, googleProrationMode: Int?, googleIsPersonalizedPrice: Boolean?, presentedOfferingIdentifier: String?, onResult: OnResult, )</ID>
-    <ID>NestedBlockDepth:common.kt$fun purchaseProduct( activity: Activity?, productIdentifier: String, type: String, googleBasePlanId: String?, googleOldProductId: String?, googleProrationMode: Int?, googleIsPersonalizedPrice: Boolean?, presentedOfferingIdentifier: String?, onResult: OnResult, )</ID>
-    <ID>NestedBlockDepth:common.kt$fun purchaseSubscriptionOption( activity: Activity?, productIdentifier: String, optionIdentifier: String, googleOldProductId: String?, googleProrationMode: Int?, googleIsPersonalizedPrice: Boolean?, presentedOfferingIdentifier: String?, onResult: OnResult, )</ID>
```
JayShortway, line 19: Nice to remove some of these!

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
JayShortway: I think that makes sense! It's also consistent with other enums, like `verificationMode`.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -574,6 +578,17 @@ fun getPromotionalOffer(): ErrorContainer {
...
+private fun String.toPurchasesAreCompletedBy(): PurchasesAreCompletedBy? {
+    return when (this) {
+        "REVENUECAT" -> PurchasesAreCompletedBy.REVENUECAT
+        "MY_APP" -> PurchasesAreCompletedBy.MY_APP
+        else -> {
+            Log.e("PurchasesHybridCommon", "Unrecognized PurchasesAreCompletedBy: $this")
+            null
+        }
+    }
+}
```
JayShortway, line 590: We can use `enumValueOf()` or `PurchasesAreCompletedBy.valueOf()` to avoid having to write out the literal enum entries. 

```kotlin
private fun String.toPurchasesAreCompletedBy(): PurchasesAreCompletedBy? {
    return try {
        enumValueOf<PurchasesAreCompletedBy>(this)
    } catch (e: IllegalArgumentException) {
        null
    }
}
```
I'd also be okay not catching the exception and failing fast. It means we have made a mistake integrating PHC in a hybrid, and a crash is easier to notice (and thus more likely to get fixed) than unusual behavior due to a null `PurchasesAreCompletedBy`.

ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
```diff
@@ -109,6 +113,22 @@ extension LogLevel {
...
+extension PurchasesAreCompletedBy {
+
+    static func fromString(_ purchasesAreCompletedByString: String) -> PurchasesAreCompletedBy? {
+        var purchasesAreCompletedBy: PurchasesAreCompletedBy? = nil
+        if purchasesAreCompletedByString == "MY_APP" {
+            purchasesAreCompletedBy = .myApp
+        } else if purchasesAreCompletedByString == "REVENUECAT" {
+            purchasesAreCompletedBy = .revenueCat
+        } else {
+            NSLog("Error: Unrecognized purchasesAreCompletedBy \(purchasesAreCompletedByString)")
```
JayShortway, line 125: Same comment about crash vs log here. Not a blocker.

typescript/api-report/purchases-typescript-internal.api.md
```diff
@@ -204,7 +210,7 @@ export enum PRODUCT_TYPE {
 
 // @public
 export enum PRORATION_MODE {
-    DEFERRED = 4,
+    DEFERRED = 6,
```
JayShortway, line 213: 🙌

typescript/api-report/purchases-typescript-internal.api.md
JayShortway: Imo a breaking change in an internal library is okay, to make sure we always use the correct api call. Or is PHC not strictly internal?

JayShortway (commented):
Very nice, thanks for taking care of this! Just had a few suggestions, but looking good!

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
tonidero: Good point! Will do that change.

As for crash vs log, we've been very reticent to do crashes even for things that are likely our mistake like this... Happy to discuss this policy more though!

typescript/api-report/purchases-typescript-internal.api.md
tonidero: Well this is a bit confusing actually, this typescript library actually provides some interfaces that are used as the public interface in the React Native and Capacitor SDKs (we basically expose these interfaces from those libraries). So a breaking change here would mean a breaking change in those 2 SDKs and I would like to avoid that. 

If it was only PHC I would completely agree with you, though I would probably make a major just in case (there have been people using PHC at some point 😅 )

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
tonidero: Done in https://github.com/RevenueCat/purchases-hybrid-common/pull/859/commits/dab8a3fce431c80955f4f707c3b5bc5bbf90a9f0

JayShortway: Yea let's not deviate from a previously discussed policy here! I would like to learn a bit more about it, but let's do that in our weekly for instance.

typescript/api-report/purchases-typescript-internal.api.md
JayShortway: Oh yes that totally makes sense then! Thanks for the context!

JayShortway (approved):

PR #860, opened 2024-07-10
Fix typescript API tests

tonidero:
The command is executed from the fastlane folder, so the diff was failing and not catching issues


JayShortway (approved):
Good one!!

PR #863, opened 2024-07-15
Release/11.2.0

MarkVillacampa:
### Other Changes
* Add method for checking Amazon LWA consent status via Mark Villacampa (@MarkVillacampa)


tonidero (approved):
Looks good! We should also get the Amazon change merged into `main`, to make sure we don't forget once we make the major.

PR #875, opened 2024-07-24
Add Kotlin version of the syncPurchases method that returns the customerInfo

MarkVillacampa:
The iOS already returns it, but not the Android one. This adds a method overload on Kotlin so we can potentially implement this in the hybrids.

### DIFF
--- a/android/api-tests/src/test/java/com/revenuecat/apitests/kotlin/CommonApiTests.kt
+++ b/android/api-tests/src/test/java/com/revenuecat/apitests/kotlin/CommonApiTests.kt
@@ -173,8 +173,9 @@ private class CommonApiTests {
         getCustomerInfo(onResult)
     }
 
-    fun checkSyncPurchases() {
+    fun checkSyncPurchases(onResult: OnResult) {
         syncPurchases()
+        syncPurchases(onResult)
     }
 
     fun checkIsAnonymous() {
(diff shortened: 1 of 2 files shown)


android/api-tests/src/test/java/com/revenuecat/apitests/kotlin/CommonApiTests.kt
```diff
@@ -173,8 +173,9 @@ private class CommonApiTests {
         getCustomerInfo(onResult)
     }
 
-    fun checkSyncPurchases() {
+    fun checkSyncPurchases(onResult: OnResult) {
         syncPurchases()
+        syncPurchases(onResult)
```
fire-at-will, line 178: Shouldn't this replace the existing syncPurchases() call so we don't call syncPurchases twice?

MarkVillacampa: Kotlin method overload means both methods should be fine to live side by side

vegaro: fwiw you could do:

```
@JvmOverloads
fun syncPurchases(
    onResult: OnResult? = null,
) {
```

to only have a single `syncPurchases`, but that requires adding logic inside to check which `syncPurchases` function to call, so I think what you  did is also ok

fire-at-will: @MarkVillacampa This was my bad, I thought that this change was added somewhere in the codebase's `syncPurchases()` function, but I now realize that this is in an API tester so it's totally okay :D

fire-at-will (approved):

PR #884, opened 2024-07-31
Require STOREKIT_VERSION when PURCHASES_ARE_COMPLETED_BY is MY_APP

fire-at-will:
As of iOS native SDK 5.0.0, developers must specify the StoreKit version they're using when setting `PurchasesAreCompletedBy`. This PR adjusts the typescript APIs to:
- Remove the existing `observerMode: bool` parameter
- Change `purchasesAreCompletedBy` to be a new `PURCHASES_ARE_COMPLETED_BY` combined type, which can be either `PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT` or a new type, `PURCHASES_ARE_COMPLETED_BY_MY_APP`, which requires developers specify their StoreKit version.

### Code Samples
Setting `purchasesAreCompletedBy` to `REVENUECAT`:
```typescript
Purchases.configure({
 apiKey: "123",
 purchasesAreCompletedBy: PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT,
});
```

Settings `purchasesAreCompletedBy` to `MY_APP`:
```typescript
Purchases.configure({
 apiKey: "123",
 purchasesAreCompletedBy: {
    type: PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP,
    storeKitVersion: STOREKIT_VERSION.STOREKIT_1
 },
});
```

### Android-Only Apps with `PURCHASES_ARE_COMPLETED_BY` == `MY_APP`
Android-only apps that want to complete their own purchases can specify any StoreKit version as that parameter is ignored by the Android SDK. This fact has been detailed in `PURCHASES_ARE_COMPLETED_BY_MY_APP`'s docstring.

### Acknowledgements
Huge thanks to @vegaro for coming up with the idea and prototyping it!

### DIFF
--- a/typescript/api-report/purchases-typescript-internal.api.md
+++ b/typescript/api-report/purchases-typescript-internal.api.md
@@ -226,15 +226,6 @@ export enum PURCHASE_TYPE {
 }
 
 // @public
-export type PURCHASES_ARE_COMPLETED_BY = PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT | PURCHASES_ARE_COMPLETED_BY_MY_APP;
-
-// @public
-export type PURCHASES_ARE_COMPLETED_BY_MY_APP = {
-    type: PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP;
-    storeKitVersion: STOREKIT_VERSION;
-};
-
-// @public
 export enum PURCHASES_ARE_COMPLETED_BY_TYPE {
     MY_APP = "MY_APP",
     REVENUECAT = "REVENUECAT"
@@ -315,12 +306,21 @@ export enum PURCHASES_ERROR_CODE {
 }
 
 // @public
+export type PurchasesAreCompletedBy = PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT | PurchasesAreCompletedByMyApp;
+
+// @public
+export type PurchasesAreCompletedByMyApp = {
+    type: PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP;
+    storeKitVersion: STOREKIT_VERSION;
+};
+
+// @public
 export interface PurchasesConfiguration {
     apiKey: string;
     appUserID?: string | null;
     entitlementVerificationMode?: ENTITLEMENT_VERIFICATION_MODE;
     pendingTransactionsForPrepaidPlansEnabled?: boolean;
-    purchasesAreCompletedBy?: PURCHASES_ARE_COMPLETED_BY;
+    purchasesAreCompletedBy?: PurchasesAreCompletedBy;
     shouldShowInAppMessagesAutomatically?: boolean;
     storeKitVersion?: STOREKIT_VERSION;
     useAmazon?: boolean;
--- a/typescript/api-report/purchases-typescript-internal.api.md
+++ b/typescript/api-report/purchases-typescript-internal.api.md
@@ -226,7 +226,16 @@ export enum PURCHASE_TYPE {
 }
 
 // @public
-export enum PURCHASES_ARE_COMPLETED_BY {
+export type PURCHASES_ARE_COMPLETED_BY = PURCHASES_ARE_COMPLETED_BY_TYPE.REVENUECAT | PURCHASES_ARE_COMPLETED_BY_MY_APP;
+
+// @public
+export type PURCHASES_ARE_COMPLETED_BY_MY_APP = {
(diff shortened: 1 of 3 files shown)


typescript/api-report/purchases-typescript-internal.api.md
```diff
@@ -310,8 +319,6 @@ export interface PurchasesConfiguration {
     apiKey: string;
     appUserID?: string | null;
     entitlementVerificationMode?: ENTITLEMENT_VERIFICATION_MODE;
-    // @deprecated
-    observerMode?: boolean;
```
MarkVillacampa, line 314: 🔥

typescript/src/purchasesConfiguration.ts
```diff
@@ -17,13 +17,6 @@ export interface PurchasesConfiguration {
    * A unique id for identifying the user
    */
   appUserID?: string | null;
-  /**
-   * An optional boolean. Set this to TRUE if you have your own IAP implementation and
-   * want to use only RevenueCat's backend. Default is FALSE. If you are on Android and setting this to ON, you will have
-   * to acknowledge the purchases yourself.
-   * @deprecated Use purchasesAreCompletedBy instead.
-   */
-  observerMode?: boolean;
```
tonidero, line 26: It’s probably fine but this is a breaking change so I would label this PR as breaking. We believe it’s ok to make many breaking changes in PHC as long as we don’t propagate them to the hybrids

typescript/src/enums.ts
```diff
@@ -244,3 +244,45 @@ export enum PURCHASES_ARE_COMPLETED_BY {
    */
   REVENUECAT = "REVENUECAT",
 }
+
+/**
+ * Configuration option that specifies that your app will complete purchases.
+ * @public
+ */
+export type PURCHASES_ARE_COMPLETED_BY_MY_APP = {
```
tonidero, line 252: Hmm I’m honestly not sure about naming this uppercase. Feels like it should be an enum. Probably better to do: `PurchasesAreCompletedByMyApp` for the type?

typescript/src/enums.ts
```diff
@@ -244,3 +244,45 @@ export enum PURCHASES_ARE_COMPLETED_BY {
...
+ *  apiKey: "123",
+ *  purchasesAreCompletedBy: {
+ *    type: PURCHASES_ARE_COMPLETED_BY.MY_APP,
+ *    storeKitVersion: STOREKIT_VERSION.STOREKIT_1
+ *  },
+ * });
+ * ```
+ * @public
+ */
+export type PURCHASES_ARE_COMPLETED_BY =
```
tonidero, line 286: Same here about the all case naming. I would do `PurchasesAreacompletedBy` here?

tonidero (commented):
Some naming suggestions but looking good!

typescript/src/enums.ts
fire-at-will: Works for me, I'll make the change!

typescript/src/enums.ts
fire-at-will: Updated! :D

typescript/src/purchasesConfiguration.ts
```diff
@@ -17,19 +17,12 @@ export interface PurchasesConfiguration {
...
   /**
    * Set this to MY_APP if you have your own IAP implementation and
```
tonidero, line 21: Need to update the docs here?

typescript/src/enums.ts
```diff
@@ -244,3 +244,45 @@ export enum PURCHASES_ARE_COMPLETED_BY {
...
+ * @public
+ */
+export type PurchasesAreCompletedByMyApp = {
+  type: PURCHASES_ARE_COMPLETED_BY_TYPE.MY_APP;
+
+  /**
+   * The version of StoreKit that your app is using to make purchases. This value is ignored
+   * on Android, so if your app is Android-only, you may provide any value.
+   */
+  storeKitVersion: STOREKIT_VERSION;
```
tonidero, line 259: I guess with this approach there is a chance of people setting conflicting versions here and in the `storeKitVersion` parameter in the `PurchasesConfiguration`... IMO in those cases we should just log a warning and default to this one.

tonidero (approved):
I think this approach makes sense! Mostly we need to update the docs.

typescript/src/enums.ts
fire-at-will: Great point - I'll be sure to do that in the hybrid SDKs 😄

PR #911, opened 2024-09-02
Adds StoreProduct.priceAmount and StoreProductDiscount.priceAmount on iOS.

JayShortway:
As the title says. 

## Motivation
Because these were not exposed before this PR, purchases-kmp was reconstructing the price by parsing the formatted price. Besides being very brittle, this _might_ also be the cause of https://github.com/RevenueCat/purchases-kmp/issues/161, as we return `0` if the price formatter is null.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
@@ -12,7 +12,11 @@ import StoreKit
 
 @objc public extension StoreProduct {
     
-    // Re-exports pricePerPeriod properties with different names to avoid recursion.
+    // Re-exports price properties with different names to avoid recursion.
+
+    @objc(price) var priceAmount: NSDecimalNumber {
+        return self.priceDecimalNumber
+    }
 
     @available(iOS 11.2, macOS 10.13.2, tvOS 11.2, watchOS 6.2, *)
     @objc var pricePerWeekAmount: NSDecimalNumber? {
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
@@ -14,7 +14,7 @@ import StoreKit
     
     // Re-exports price properties with different names to avoid recursion.
 
-    @objc(price) var priceAmount: NSDecimalNumber {
+    @objc var priceAmount: NSDecimalNumber {
         return self.priceDecimalNumber
     }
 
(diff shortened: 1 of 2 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
```diff
@@ -12,7 +12,11 @@ import StoreKit
 
 @objc public extension StoreProduct {
     
-    // Re-exports pricePerPeriod properties with different names to avoid recursion.
+    // Re-exports price properties with different names to avoid recursion.
+
+    @objc(price) var priceAmount: NSDecimalNumber {
```
tonidero, line 17: Hmm I noticed that the underlying property is `priceDecimalNumber` so couldn't we just expose it in objc as `priceAmount`? (Not sure if there is other benefit to change the name in objc to `price`)
```suggestion
    @objc var priceAmount: NSDecimalNumber {
```

ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProductDiscount+HybridAdditions.swift
```diff
@@ -10,6 +10,16 @@ import Foundation
 import StoreKit
 import RevenueCat
 
+@objc public extension StoreProductDiscount {
+    
+    // Re-exports price property with a different name to avoid recursion.
+
+    @objc(price) var priceAmount: NSDecimalNumber {
```
tonidero, line 17: Same

tonidero (approved):
Just a question, but not a blocker. Looks good!

ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProduct+HybridAdditions.swift
JayShortway: This was inspired by purchases-ios, which also renames `priceDecimalNumber` to `price` for ObjC, see [here](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift#L206). Do you think there are potential downsides to this renaming?

tonidero: Mostly that there is a mismatch between the names in swift and objc, which can be confusing, specially when comparing different SDKs. As in, if we're going to be doing this, I would suggest to also rename the `priceAmount` in swift to `price` (though I very slightly prefer `priceAmount`?).

JayShortway: Yea no you're right, I've also started to prefer `priceAmount`. `price` is too ambiguous. (E.g. is it formatted, is it the amount?) https://github.com/RevenueCat/purchases-hybrid-common/pull/911/commits/9f94b1745fca341794d5d6e082c8023c08d1afc7

ios/PurchasesHybridCommon/PurchasesHybridCommon/StoreProductDiscount+HybridAdditions.swift
JayShortway: https://github.com/RevenueCat/purchases-hybrid-common/pull/911/commits/9f94b1745fca341794d5d6e082c8023c08d1afc7

PR #922, opened 2024-09-23
Support integration via Swift Package Manager

JayShortway:
This adds a `Package.swift` file, allowing PHC to be integrated via SPM. This is mainly useful for purchases-kmp users, who need to integrate PHC manually, many of which aren't using CocoaPods. Some requests for this: 

- https://github.com/RevenueCat/purchases-kmp/issues/177#issuecomment-2331069006
- https://github.com/RevenueCat/purchases-kmp/issues/216#issuecomment-2367751406 

I was able to show a Paywall on iOS in a fresh KMP project by pointing it at the `Package.swift` file on my machine, without integrating CocoaPods.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,15 @@
+{
+  "originHash" : "1c8e7b22f143659dfd3d9b4d8325e3eacccb1e50f24657cf6a87483bc3d6717a",
+  "pins" : [
+    {
+      "identity" : "purchases-ios-spm",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/RevenueCat/purchases-ios-spm",
+      "state" : {
+        "revision" : "dabbf506344b6f924b44bdfabbaec96bc7943d5d",
+        "version" : "5.3.3"
+      }
+    }
+  ],
+  "version" : 3
+}
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -24,11 +24,13 @@ files_with_version_number_without_prerelease_modifiers = {
     './ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Info.plist' => PLIST_VERSION_PATTERNS,
 }
 POD_VERSION_PATTERNS = ['\'RevenueCat\', \'{x}\'', '\'RevenueCatUI\', \'{x}\'']
+SPM_VERSION_PATTERNS = ['.package(url: "https://github.com/RevenueCat/purchases-ios-spm", exact: "{x}"),']
 files_to_update_ios_dependency = {
     'PurchasesHybridCommon.podspec' => POD_VERSION_PATTERNS,
     'PurchasesHybridCommonUI.podspec' => POD_VERSION_PATTERNS,
     'ios/PurchasesHybridCommon/Podfile' => POD_VERSION_PATTERNS,
     'DEVELOPMENT.md' => POD_VERSION_PATTERNS,
+    'Package.swift' => SPM_VERSION_PATTERNS,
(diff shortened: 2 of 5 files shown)


Package.resolved
```diff
@@ -0,0 +1,15 @@
+{
+  "originHash" : "1c8e7b22f143659dfd3d9b4d8325e3eacccb1e50f24657cf6a87483bc3d6717a",
+  "pins" : [
+    {
+      "identity" : "purchases-ios-spm",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/RevenueCat/purchases-ios-spm",
+      "state" : {
+        "revision" : "dabbf506344b6f924b44bdfabbaec96bc7943d5d",
+        "version" : "5.3.3"
```
tonidero, line 10: Hmm can we automate updating this as well on each release?

tonidero (approved):
LGTM! Just a comment. Might be good checking with some iOS devs as well before merging though.

Package.resolved
JayShortway: Good one! This seems to do the trick: https://github.com/RevenueCat/purchases-hybrid-common/pull/922/commits/dfd4d0103d6e2f1b075f9635384e5a78f8affd78

aboedo: we can also not version control package.resolved

fastlane/Fastfile
```diff
@@ -24,11 +24,15 @@ files_with_version_number_without_prerelease_modifiers = {
     './ios/PurchasesHybridCommon/PurchasesHybridCommonTests/Info.plist' => PLIST_VERSION_PATTERNS,
 }
 POD_VERSION_PATTERNS = ['\'RevenueCat\', \'{x}\'', '\'RevenueCatUI\', \'{x}\'']
+SPM_VERSION_PATTERNS = ['.package(url: "https://github.com/RevenueCat/purchases-ios-spm", exact: "{x}"),']
+SPM_RESOLVED_VERSION_PATTERNS = ['        "version" : "{x}"']
```
aboedo, line 28: do we need all the whitespace before the version here?

aboedo (approved):

fastlane/Fastfile
JayShortway: Not necessarily. The intention was to make the pattern as strict as possible, but it works without it too. Removed here: https://github.com/RevenueCat/purchases-hybrid-common/pull/922/commits/13b2b231ada16b2852f6e7bc95b479d2acbac15f.

Package.resolved
JayShortway: I don't have a strong opinion, so I Googled some "official" advice. From [swiftlang/swift-evolution](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0175-package-manager-revised-dependency-resolution.md#proposed-solution): 

> Some users may chose to add the `Package.resolved` file to their package's `.gitignore` file. When this file is checked in, it allows a team to coordinate on what versions of the dependencies they should use. If this file is gitignored, each user will separately choose when to get new versions based on when they run the `swift package update` command, and new users will start with the latest eligible version of each dependency. Either way, for a package which is a dependency of other packages (e.g. a library package), that package's `Package.resolved` file will not have any effect on its client packages.

Seems slightly safer to check it in? (Very slightly, because our dependencies have `exact` versions currently.) And it won't affect library consumers apparently.

vegaro (approved):

Package.resolved
JayShortway: Will merge it for now, but we can always remove it later.

PR #930, opened 2024-10-01
[CI] Fix update dependencies job to use correct labels

tonidero:
We were using the old labels. This updates the job to use the new labels and get the version numbers right.


JayShortway (approved):
Nice catch!

PR #938, opened 2024-10-16
Adds `Offerings.currentOfferingForPlacement()`

JayShortway:
As the title says. Used the same approach as we did for `StoreProduct.priceAmount` in https://github.com/RevenueCat/purchases-hybrid-common/pull/911. 

Note that there's an existing `getCurrentOfferingForPlacement()` in `CommonFunctionality`, but that's an async call and thus not entirely the same thing. 

- Dependant: https://github.com/RevenueCat/purchases-kmp/pull/243
- Addresses: https://github.com/RevenueCat/purchases-kmp/issues/242


tonidero (approved):
:shipit:

PR #952, opened 2024-10-29
Add Amazon `freeTrialPeriod` to `introPrice`

vegaro:
We are not exposing Amazon's `freeTrialPeriod` anywhere, even if we have that information in `AmazonStoreProduct`


vegaro:
Tested it in unity and it works

tonidero (approved):

PR #974, opened 2024-11-28
Add methods to set Tenjin and Kochava integration IDs

tonidero:
This brings the `setKochavaDeviceID` and `setTenjinAnalyticsInstallationID` setter methods to PHC, to be used by the hybrids.


JayShortway (approved):
Looks good!

PR #979, opened 2024-12-03
Change dependency-update job to macos machine

tonidero:
After updating cocoapods in https://github.com/RevenueCat/purchases-hybrid-common/pull/957, it seems the dependency-update job has been failing. I thought about reverting to a previous cocoapods version, but I think it might be good to stay on the latest...


JayShortway (approved):
🚢! 

>but I think it might be good to stay on the latest...
  
I definitely agree

PR #1004, opened 2025-01-03
Add `parseAsWebPurchaseRedemption` API to iOS

tonidero:
This adds a new API to be used by the KMP SDK. This is because we need to have access to the iOS native `WebPurchaseRedemption` object in order to use the native `redeemWebPurchase` method, as we do in the KMP SDK.

 This replaces the PR in https://github.com/RevenueCat/purchases-ios/pull/4629, since we don't really need access to those APIs from the native SDK.


JayShortway (approved):
LGTM!

PR #1008, opened 2025-01-08
Introduce IOSAPIAvailabilityChecker for KMP

fire-at-will:
This PR introduces an `IOSAPIAvailabilityChecker` class to allow for easy checking of iOS API availabilities in the KMP SDK.

### IOSAPIAvailabilityChecker
The win-back offer APIs are only available on iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, and visionOS 2.0. We don't have a great way of checking this at runtime in the KMP SDK - the best we can do is to perform string comparisons on `UIDevice.currentDevice.systemVersion`, which is far from ideal. Instead, this PR introduces a `IOSAPIAvailabilityChecker` class, which allows the KMP SDK to check the availability of the win-back APIs in a more robust manner since Swift can perform safer checks without relying on `UIDevice`.


ios/PurchasesHybridCommon/PurchasesHybridCommon/PurchaseParams+HybridAdditions.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright © 2025 RevenueCat. All rights reserved.
+//
+
+import Foundation
+import RevenueCat
+
+@objc
+public extension PurchaseParams.Builder {
+
+    // For some reason, PurchaseParams.Builder isn't exposed as a part of the PHC's public API to KMP by default.
```
JayShortway, line 15: I think it makes sense it's not exposed. PHC doesn't have any public method that either takes `PurchaseParams.Builder` as an argument or returns it, so the type is not part of it's public API. (Or am I missing something?)

If what I'm saying makes sense, maybe we can remove "For some reason" here? 
```suggestion
    // PurchaseParams.Builder isn't exposed as a part of the PHC's public API to KMP by default.
```

JayShortway (approved):
LGTM! Just a suggestion on a comment.

JayShortway (approved):

JayShortway (approved):
Nice nice!

PR #1011, opened 2025-01-13
Expose storeKitError for iOS if any

tonidero:
This adds the `storeKitError` property that includes the `code`, `domain` and `message` properties for the original storekit error if any.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
@@ -27,6 +27,14 @@ import RevenueCat
 
         info["underlyingErrorMessage"] = underlyingErrorMessage ?? ""
 
+        if let storeKitError = self.findStoreKitErrorCodeIfAny(nsError) {
+            info["storeKitError"] = [
+                "code": storeKitError.code,
+                "domain": storeKitError.domain,
+                "message": storeKitError.localizedDescription
+            ]
+        }
+
         // todo: remove "readable_error_code" and instead send whole user info instead
         // also: code name is already exposed as error.code
         if let readableErrorCode = nsError.userInfo["readable_error_code"] {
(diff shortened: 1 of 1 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
```diff
@@ -27,6 +27,14 @@ import RevenueCat
 
         info["underlyingErrorMessage"] = underlyingErrorMessage ?? ""
 
+        if let storeKitError = ErrorContainer.findStoreKitErrorCodeIfAny(nsError) {
+            info["storeError"] = [
```
tonidero, line 31: Note that this is not used by some hybrids (like RN, which uses the original error). But this will include this information for those hybrids that do.

ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
```diff
@@ -53,4 +61,18 @@ import RevenueCat
 
         self.info = info
     }
+
+    private static func findStoreKitErrorCodeIfAny(_ error: Error) -> NSError? {
+        var currentError: NSError? = error as NSError
+        var storeKitError: NSError?
+        while let underlyingNSError = currentError?.userInfo[NSUnderlyingErrorKey] as? NSError {
+            if !underlyingNSError.domain.starts(with: "RevenueCat") {
```
aboedo, line 69: where do we define the `RevenueCat` domain? I couldn't find it in a quick search

tonidero: I believe it uses the module name and the file/class name. In my tests, our domains looked like: `RevenueCat.ErrorCode` or `RevenueCat.ProductsFetcherSK2.Error`

aboedo: I see 👍 thanks! Asking because I remember we used to set it explicitly back in the objc days, guess we moved away from it.

aboedo (approved):

ajpallares (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift
tonidero: That's correct... We would need to figure out a way since in RN, we just pick the original error, but yeah, I think that makes a lot of sense 👍

PR #1016, opened 2025-01-17
Dismiss PaywallFragment after PaywallActivity is dismissed

tonidero:
We were not dismissing the `PaywallFragment` after the `PaywallActivity` was closed. This caused some crashes like: https://github.com/RevenueCat/purchases-flutter/issues/1090.

Repro steps were:
- Present paywall
- Dismiss it
- Background the app
- Wait for app to be killed by the system or kill it with adb
- Reopen app and crash

### DIFF
--- a/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
+++ b/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
@@ -99,7 +99,11 @@ internal class PaywallFragment : Fragment(), PaywallResultHandler {
         // This should normally never happen, but just in case, we don't want to try to present the paywall
         // if the SDK is not configured.
         if (!Purchases.isConfigured) {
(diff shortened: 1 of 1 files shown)


android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallFragment.kt
```diff
@@ -94,6 +96,14 @@ internal class PaywallFragment : Fragment(), PaywallResultHandler {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        // This should normally never happen, but just in case, we don't want to try to present the paywall
+        // if the SDK is not configured.
+        if (!Purchases.isConfigured) {
```
tonidero, line 101: I don't think this is needed... But I added it as a failsafe, in case we're missing any more cases.

JayShortway (approved):
Nice to squash this! 💪

PR #1030, opened 2025-01-28
[CI] Remove requirement for pod push to succeed to continue release process

tonidero:
Pushing new pods has been very brittle lately. It's been a problem since our current CI doesn't allow to continue the workflow after the job fails. This makes it so the release process can continue even if pushing the pods fail.

Some notes about this:
- We are still running the pod lint command for the tests before releasing
- There is still a dependency of the `phcui` library to `phc`. This is because the first one depends on the second one, so it doesn't make sense to continue if the first one failed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -441,8 +441,6 @@ workflows:
       - make-github-release:
           <<: *release-tags
           requires:
-            - deploy-ios-phc
-            - deploy-ios-phcui
             - deploy-android
             - deploy-typescript
             - deploy-typescript-esm


tonidero:
> If the job fails, will we still be notified in some way so that we know that the release needs to be performed manually?

Well, whoever approved/started the job should check that everything is green before merging (same as now). We do get notified through email if you're subscribed to the project in CircleCI though!

> Maybe we could modify our publish jobs so that if pod trunk push returns an error, we search for the new version in Cocoapods with pod search, and if the output contains the new version, then we could make the deploy job pass.

That's a good one! Do you know if/how much delay there is between pushing and being available? Asking because that has happened pretty frequently in the hybrids (finishing deploying PHC but tests in the bump jobs in the hybrids failing because of a delay for the pod to be available). In any case, this can be done in a separate PR.

PR #1047, opened 2025-02-19
chore: Fix Gemfile versions to avoid Gemfile.lock updates

facumenzella:
This PR fixes the Gemfile to avoid unnecessary .lock updates

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,10 @@
 source "https://rubygems.org"
 
-gem "fastlane"
+gem 'fastlane', '2.226.0'
 
 eval_gemfile("fastlane/Pluginfile")
 
 gem "cocoapods", "~> 1.16"
-gem 'danger'
-gem 'rest-client'
-gem "lefthook", "~> 1.10"
+gem 'danger', '9.5.1'
+gem 'rest-client', '2.1.0'
+gem "lefthook", "1.10.10"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,10 +1,10 @@
...
 
 eval_gemfile("fastlane/Pluginfile")
 
 gem "cocoapods", "~> 1.16"
-gem 'danger'
-gem 'rest-client'
-gem "lefthook", "~> 1.10"
+gem 'danger', '9.5.1'
+gem 'rest-client', '2.1.0'
+gem "lefthook", "1.10.10"
```
tonidero, line 10: Hmm I actually think we are leaving these with the version unspecified so we can get auto-updates from dependabot. Since they are only related to CI, it's mostly ok to get auto-updates... Are you getting any issues with the updates?

facumenzella: Well, I am used to fixing versions in any Gemfile. I tried running a fastlane command locally, and ended up with an updated lock file.

Does this prevent updates? I am used to dependabot opening PRs whenever there's an update. I think we might just need to add a few more lines to `dependabot.yml`

That keeps track of any change, and we can control what we update and when we update. Great to avoid unexpected broken release trains.

facumenzella:
@tonidero I'll close this 👍

PR #1050, opened 2025-02-20
fix: improve period unit ts type

itsramiel:
Thank you for contributing to purchases-hybrid-common. Before pressing the "Create Pull Request" button, please provide the following:

  - [x] A description about what and why you are contributing, even if it's trivial: 
  		If periodUnit can be only `DAY, WEEK, MONTH or YEAR` as mentioned in the comment, then it is better to type it that way. Makes it easier to switch on the predefined union or perform if checks with intellisense

  - [x] The issue number(s) or PR number(s) in the description if you are contributing in response to those:
  		There is no issue number

  - [x] If applicable, unit tests.
  		Not applicable

### DIFF
--- a/typescript/src/offerings.ts
+++ b/typescript/src/offerings.ts
@@ -311,7 +311,7 @@ export interface PurchasesIntroPrice {
   /**
    * Unit for the billing period of the discount, can be DAY, WEEK, MONTH or YEAR.
    */
-  readonly periodUnit: string;
+  readonly periodUnit: "DAY" | "WEEK" | "MONTH" | "YEAR";
   /**
    * Number of units for the billing period of the discount.
    */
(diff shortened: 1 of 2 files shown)


typescript/src/offerings.ts
```diff
@@ -311,7 +311,7 @@ export interface PurchasesIntroPrice {
   /**
    * Unit for the billing period of the discount, can be DAY, WEEK, MONTH or YEAR.
    */
-  readonly periodUnit: string;
+  readonly periodUnit: "DAY" | "WEEK" | "MONTH" | "YEAR";
```
tonidero, line 314: This can also be `UNKNOWN` in unexpected scenarios:
```suggestion
  readonly periodUnit: "DAY" | "WEEK" | "MONTH" | "YEAR" | "UNKNOWN";
```

tonidero (commented):
Hi @itsramiel, thank you very much for your contribution! 

Looking at it, it does make a lot of sense, unfortunately this would be a breaking change in the API, so we might hold this until the next major release. Hope that's ok!

PR #1067, opened 2025-03-03
Add `subscriptionsByProductIdentifier` map of SubscriptionInfo by productId to CustomerInfo

tonidero:
This adds some info we've recently added to the natives with information that might be useful to devs.

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
@@ -25,7 +25,7 @@ internal extension SubscriptionInfo {
             "isSandbox": isSandbox,
             "unsubscribeDetectedAt": unsubscribeDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
             "billingIssuesDetectedAt": billingIssuesDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
-            "gracePeriodExpiresDate": billingIssuesDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
+            "gracePeriodExpiresDate": gracePeriodExpiresDate?.rc_formattedAsISO8601() ?? NSNull(),
             "ownershipType": ownershipType.ownershipTypeString,
             "periodType": periodType.periodTypeString,
             "refundedAt": refundedAt?.rc_formattedAsISO8601() ?? NSNull(),
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
@@ -1,0 +1,39 @@
+//
+//  EntitlementInfo+HybridAdditions.swift
+//  PurchasesHybridCommon
+//
+//  Created by Andrés Boedo on 4/13/22.
+//  Copyright © 2022 RevenueCat. All rights reserved.
+//
+
+import Foundation
+import RevenueCat
+
+internal extension SubscriptionInfo {
+
+    var dictionary: [String: Any] {
+        var priceObject: [String: Any]? = nil
+        if let priceCurrency = price?.currency, let priceAmount = price?.amount {
+            priceObject = ["currency": priceCurrency, "amount": priceAmount]
+        }
+        return [
+            "productIdentifier": productIdentifier,
+            "purchaseDate": purchaseDate.rc_formattedAsISO8601(),
+            "originalPurchaseDate": originalPurchaseDate?.rc_formattedAsISO8601() ?? NSNull(),
+            "expiresDate": expiresDate?.rc_formattedAsISO8601() ?? NSNull(),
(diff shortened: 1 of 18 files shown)


typescript/src/customerInfo.ts
```diff
@@ -50,7 +69,7 @@ export interface PurchasesEntitlementInfo {
     /**
      * The store where this entitlement was unlocked from.
      */
-    readonly store: "PLAY_STORE" | "APP_STORE" | "STRIPE" | "MAC_APP_STORE" | "PROMOTIONAL" | "AMAZON" | "RC_BILLING" | "EXTERNAL" | "UNKNOWN_STORE";
+    readonly store: Store;
```
tonidero, line 72: I believe this isn't a breaking change since it's actually the same type.

ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    var dictionary: [String: Any] {
+        var priceObject: [String: Any]? = nil
+        if let priceCurrency = price?.currency, let priceAmount = price?.amount {
+            priceObject = ["currency": priceCurrency, "amount": priceAmount]
+        }
+        return [
+            "productIdentifier": productIdentifier,
+            "purchaseDate": purchaseDate.rc_formattedAsISO8601(),
+            "originalPurchaseDate": originalPurchaseDate?.rc_formattedAsISO8601() ?? NSNull(),
```
ajpallares, line 22: I know this is already the same for other objects (e.g. `CustomerInfo`), but I still want to ask. What's the benefit of having
```
key: optionalValue ?? NSNull()
```
instead of
```
key: optionalValue as Any
```
?
To me it seems simpler to just not have the key present if the value is `nil`.

ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
```diff
@@ -0,0 +1,39 @@
...
+            "expiresDate": expiresDate?.rc_formattedAsISO8601() ?? NSNull(),
+            "store": store.storeString,
+            "isSandbox": isSandbox,
+            "unsubscribeDetectedAt": unsubscribeDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
+            "billingIssuesDetectedAt": billingIssuesDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
+            "gracePeriodExpiresDate": billingIssuesDetectedAt?.rc_formattedAsISO8601() ?? NSNull(),
```
ajpallares, line 28:
```suggestion
            "gracePeriodExpiresDate": gracePeriodExpiresDate?.rc_formattedAsISO8601() ?? NSNull(),
```

ajpallares (changes requested):
Looks good. Only there's a mistake in the `SubscriptionInfo`'s `dictionary` mapping

ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
tonidero: Whoops great catch, thanks!!

ios/PurchasesHybridCommon/PurchasesHybridCommon/SubscriptionInfo+HybridAdditions.swift
tonidero: Well I can see that this could be a change in the typescript hybrids (RN + Capacitor). As in, the presence of a key but with a null value is different than the abscence of that key 😅. In this case, I think it helps discoverability to have the keys there with null values vs not having the keys at all. This would be my only point in favor of the current system.

As for the other hybrids, I think it doesn't really matter what we do here, as long as we map to the corresponding object correctly...

ajpallares: Makes sense, yes. Thanks for clarifying!

ajpallares (approved):

typescript/api-report/purchases-typescript-internal.api.md
```diff
@@ -120,6 +120,9 @@ export enum OFFER_PAYMENT_MODE {
     SINGLE_PAYMENT = "SINGLE_PAYMENT"
 }
 
+// @public
+export type OwnershipType = "PURCHASED" | "FAMILY_SHARED" | "UNKNOWN";
```
JayShortway, line 124: I have no idea if it matters, mentioning it just in case, but the order of the items are different than before: 
```typescript
readonly ownershipType: "FAMILY_SHARED" | "PURCHASED" | "UNKNOWN";
```

JayShortway (approved):
Nice! Just a very minor thing to double check.

typescript/api-report/purchases-typescript-internal.api.md
tonidero: I think it doesn't matter (it's only deciding the valid values, but the order shouldn't really matter) but thanks!

PR #1076, opened 2025-03-06
Create `CustomerCenterListenerWrapper`

vegaro:
This helps with converting the inputs of `CustomerCenterListenerWrapper` since the mappers are not public

Did some cleanup on `ExperimentalPreviewRevenueCatUIPurchasesAPI`


android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/CustomerCenterListenerWrapper.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun onFeedbackSurveyCompleted(feedbackSurveyOptionId: String) {
+        this.onFeedbackSurveyCompletedWrapper(feedbackSurveyOptionId)
+    }
+
+    override fun onRestoreCompleted(customerInfo: CustomerInfo) {
+        this.onRestoreCompletedWrapper(customerInfo = customerInfo.map())
+    }
+
+    override fun onRestoreFailed(error: PurchasesError) {
+        this.onRestoreFailedWrapper(error = error.map().info)
```
tonidero, line 19: Hmm any reason we're only sending the userInfo instead of the entire ErrorContainer?

tonidero (approved):
Looks great! Just a question about how we're mapping the errors (haven't checked iOS but we should make sure it matches)

android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/CustomerCenterListenerWrapper.kt
vegaro: I mimicked what we do in paywalls in the equivalent listener wrapper. Maybe we should send the whole error 🤔

vegaro: I reviewed this and when we build the ErrorContainer we add the following as the info

```
mapOf(
            "code" to code.code,
            "message" to message,
            "readableErrorCode" to code.name,
            "readable_error_code" to code.name,
            "underlyingErrorMessage" to (underlyingErrorMessage ?: ""),
        ) + extra
```        

it has nothing to do with `userInfo`. I was also confused by it

PR #1077, opened 2025-03-06
Add automatic changes on `.idea` files

vegaro:
These files got automatically updated. I do wonder if we need them in the repository though

I removed an empty `.idea` folder we had in the root. It just had a `.gitignore` file


tonidero (approved):
I also got these changes... I do wonder if it would be worth not checking out the .idea folder... But I think this is ok for now.

PR #1078, opened 2025-03-06
Update detekt and enable autocorrect

vegaro:
I noticed detekt didn't have autocorrect enabled. I also updated it to the latest version


tonidero (approved):
Makes sense thank you!

PR #1100, opened 2025-03-27
Expose Virtual Currency Balances

fire-at-will:
This PR exposes the virtual currency balances through the `VirtualCurrencyInfo` object. Specifically, it:
- Adds the ability to map the VirtualCurrencyInfo objects in CustomerInfo to the JSON dictionaries
- Includes unit tests for the new mapping capabilities
- Adds the new VirtualCurrency APIs to the typescript APIs

Once this is approved and merged into the `virtual-currency-dev` branch, we can do a beta release from that branch to use in the beta hybrid SDKs.


Package.swift
```diff
@@ -22,7 +22,7 @@ let package = Package(
             targets: ["PurchasesHybridCommonUI"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/RevenueCat/purchases-ios-spm", exact: "5.20.0"),
+        .package(url: "https://github.com/RevenueCat/purchases-ios-spm", exact: "5.20.2-vc-beta.1"),
```
tonidero, line 25: Not a huge deal but personally I prefer to keep version bumps off of the beta branches to avoid conflicts later. This should ideally just be bumped by the automations/fastlane commands when making a release. NABD though, FFTI!

ios/PurchasesHybridCommon/PurchasesHybridCommon/VirtualCurrencyInfo+HybridAdditions.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Will Taylor on 3/27/25.
+//  Copyright © 2025 RevenueCat. All rights reserved.
+//
+
+import Foundation
+import RevenueCat
+
+internal extension VirtualCurrencyInfo {
+
+    var rc_dictionary: [String: Any] {
```
tonidero, line 14: Hmm NABD but I noticed that in Android, the value `Any` is nullable. Should it also be in iOS? or maybe not nullable in Android?

typescript/src/customerInfo.ts
```diff
@@ -311,3 +317,16 @@ export interface PurchasesSubscriptionInfo {
...
+ * Describes information about a virtual currency in the app.
+ * 
+ * Use this class to access information about a virtual currency, such as its current balance.
+ * @public
+ */
+export interface PurchasesVirtualCurrencyInfo {
+    /**
+     * The current balance of the virtual currency.
+     */
+    readonly balance: bigint;
```
tonidero, line 331: Huh interesting... I've been using `number` for the types of these... I guess `bigint` might be more specific but I wonder if we should keep `number` for simplicity + consistency...

tonidero (approved):
Some comments but nothing really blocking!

Package.swift
MarkVillacampa: Yeah running the fastlane command with the beta version should bump the dependency

typescript/src/customerInfo.ts
MarkVillacampa: yeah I think we might be limited by int64 size elsewhere so using bigint here might not bring any benefit

ios/PurchasesHybridCommon/PurchasesHybridCommon/VirtualCurrencyInfo+HybridAdditions.swift
fire-at-will: Ah, great catch! These should definitely be consistent. The `balance` value in `VirtualCurrencyInfo` is non-optional on both iOS & Android, so I've removed the optionality on the Android side in d2fd6b2d874fc954d3b9d4b2ebf1ada60d0f7411

typescript/src/customerInfo.ts
fire-at-will: @tonidero We had a conversation in the virtual currencies Slack channel and decided to limit the max VC balance size to a signed 32-bit integer, so it'll definitely be safe to change this to a `number`. I'll make the change 😄

fire-at-will: Changed in 6f4849002637793a4982251c34237b525a3c09cb :D

PR #1104, opened 2025-04-03
Create `purchases-js-hybrid-mappings` module with `CustomerInfo` model mappings

tonidero:
This PR creates a new npm library to be used to hold the mappings from the `purchases-js` SDK to what the hybrids expect, so we can share that mapping logic accross all our hybrids. This follows a similar line to what the android and iOS PHC libraries do.

In follow-up PRs I will be adding more mappings for other models and more methods to the `PurchasesCommon` class.


purchases-js-hybrid-mappings/api-report/purchases-js-hybrid-mappings.api.md
```diff
@@ -0,0 +1,20 @@
+## API Report File for "purchases-js-hybrid-mappings"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+
+// @public (undocumented)
+export class PurchasesCommon {
+    // (undocumented)
```
tonidero, line 9: I don't think it's necessary to add documentation to this class, since it's mostly meant for internal hybrid usage.

purchases-js-hybrid-mappings/src/index.ts
```diff
@@ -0,0 +1 @@
+export { PurchasesCommon } from './purchases-common';
```
tonidero, line 1: We only need to export `PurchasesCommon` at least for now. We won't be exposing other types from the library, since everything will be basic types.

purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
```diff
@@ -0,0 +1,125 @@
...
+
+function mapEntitlementInfos(entitlementInfos: EntitlementInfos): Record<string, unknown> {
+  return {
+    "all": Object.fromEntries(
+      Object.entries(entitlementInfos.all).map(([key, value]) => [key, mapEntitlementInfo(value)])
+    ),
+    "active": Object.fromEntries(
+      Object.entries(entitlementInfos.active).map(([key, value]) => [key, mapEntitlementInfo(value)])
+    ),
+    "verification": "NOT_REQUESTED" // TODO: Trusted entitlements not available in JS SDK
```
tonidero, line 53: For now, I'm hardcoding this to `NOT_REQUESTED`. I think supporting this in web would be much much trickier

purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
```diff
@@ -0,0 +1,125 @@
...
+    "ownershipType": entitlementInfo.ownershipType,
+    "verification": "NOT_REQUESTED", // TODO: Trusted entitlements not available in JS SDK
+  };
+}
+
+function mapNonSubscriptionTransactions(nonSubscriptionTransactions: NonSubscriptionTransaction[]): Record<string, unknown>[] {
+  return nonSubscriptionTransactions.map((transaction) => {
+    return {
+      "transactionIdentifier": transaction.transactionIdentifier,
+      "revenueCatId": transaction.transactionIdentifier, // Deprecated: Use transactionIdentifier in this map instead
```
tonidero, line 86: Kept these to keep the same interface as the mobile PHC libraries... I think it's probably not needed, but haven't checked all hybrids.

purchases-js-hybrid-mappings/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "build": "tsc",
+    "test": "jest",
+    "api-test": "npm run build && api-extractor run --local --verbose",
+    "api-test:ci": "npm run build && api-extractor run --verbose",
+    "prepack": "npm run build && api-extractor run --local",
+    "lint": "eslint . --ext .ts",
+    "lint:fix": "eslint . --ext .ts --fix"
+  },
+  "dependencies": {
+    "@revenuecat/purchases-js": "^1.1.0"
```
tonidero, line 29: purchases-js 1.1.0 hasn't been released yet. Tests will fail until then, but we need the changes there to have the full mappings needed here (tested locally while pointing to a local version of the package)

purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
```diff
@@ -0,0 +1,125 @@
...
+    "allExpirationDatesMillis": Object.fromEntries(
+      Object.entries(customerInfo.allExpirationDatesByProduct).map(([key, value]) => [key, value ? value.getTime() : null])
+    ),
+    "allPurchaseDates": Object.fromEntries(
+      Object.entries(customerInfo.allPurchaseDatesByProduct).map(([key, value]) => [key, value ? value.toISOString() : null])
+    ),
+    "allPurchaseDatesMillis": Object.fromEntries(
+      Object.entries(customerInfo.allPurchaseDatesByProduct).map(([key, value]) => [key, value ? value.getTime() : null])
+    ),
+    "originalApplicationVersion": null,
```
JayShortway, line 36: Is this null on purpose?

purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
JayShortway: Good call

purchases-js-hybrid-mappings/src/purchases-common.ts
```diff
@@ -0,0 +1,42 @@
...
+    )
+    PurchasesCommon.instance = new PurchasesCommon(purchasesInstance);
+    return PurchasesCommon.instance;
+  }
+
+  static getInstance(): PurchasesCommon {
+    if (!PurchasesCommon.instance) {
+      throw new Error("Purchases not configured. Call configure() first.");
+    }
+    return PurchasesCommon.instance;
```
JayShortway, line 31: Is it possible to check the `instance` from `purchases-js` here, and wrap that if needed? Instead of keeping our own instance in `purchases-js-hybrid-mappings`. We have a similar scenario on iOS I believe. If someone configures the underlying iOS SDK, PHC won't know that. (See [this thread](https://revenuecat.slack.com/archives/CEU9GNWTW/p1735590350211529) for some discussion.)

purchases-js-hybrid-mappings/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "purchases-js-hybrid-mappings",
```
JayShortway, line 2: Just for my understanding, could this have been part of `purchases-typescript-internal`?

JayShortway (approved):
Nice work! Only 1 comment on the singleton instance.

purchases-js-hybrid-mappings/package.json
tonidero: Not really, for clarification:
- `purchases-typescript-internal`. Is used by TS hybrids (right now only React Native + Capacitor). It exposes types that those hybrids will use in their public API. It doesn't contain any logic, just types, so we can have a "mostly" common API among them
- `purchases-js-hybrid-mappings`. Will be used by all hybrids supporting web and will have logic to map the methods/models from `purchases-js` to/from what the hybrids expect.

purchases-js-hybrid-mappings/src/purchases-common.ts
tonidero: I thought about that, but then I think we would have 2 options:
- Make all functions in `PurchasesCommon` static, which internally would call the private instance of the `purchases-js` `Purchases`. I was trying to avoid this
- Make a `getInstance` in `PurchasesCommon` that returns a new instance of `PurchasesCommon` every time it's called, but passing the cached purchases-js instance.

I didn't like those options, which is what I went this way. Wdyt? FWIW, if a user calls `configure` again, it would still overwrite the `purchases-js` instance.

purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
tonidero: Yes, it's null in Android as well.. it's only not null in iOS. I guess we could try to add it, but we should add it at the `purchases-android` as well. For now, leaving it empty since it seems to be a non-critical API.

JayShortway: Makes sense, thanks for explaining!

purchases-js-hybrid-mappings/package.json
JayShortway: Ah yes, makes sense. Thanks!

purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
```diff
@@ -0,0 +1,125 @@
...
+      "periodType": mapPeriodType(subscriptionInfo.periodType),
+      "refundedAt": subscriptionInfo.refundedAt ? subscriptionInfo.refundedAt.toISOString() : null,
+      "storeTransactionId": subscriptionInfo.storeTransactionId,
+      "isActive": subscriptionInfo.isActive,
+      "willRenew": subscriptionInfo.willRenew,
+    }])
+  )
+}
+
+function mapPeriodType(periodType: PeriodType): string {
```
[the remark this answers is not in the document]
tonidero: This already accounts for all possible cases.

PR #1109, opened 2025-04-07
[purchases-js-hybrid-mapping] Add mapping for errors and run prettier in the library

tonidero:
2 main changes in this PR:
- Adds mapping for `purchases-js` `PurchasesError` into the error format expected in hybrids
- Adds prettier and runs it over the codebase.


JayShortway (approved):
So pretty! It's a bit hard to isolate the error-mapping changes from the prettier changes, but I'm assuming all error-mapping happens in `purchases_error_mapper.ts` and `purchases_error_mapper.test.ts`. If so: looks good!

tonidero:
Right, sorry about that 🙇. But yeah, you're correct, that's mostly contained there and a bit in the `purchases_common.ts` file where the actual mapping happens.

PR #1115, opened 2025-04-14
[purchases-js-hybrid-mappings] Add login/logout functions and set PlatformInfo

tonidero:
- Adds `logIn` and `logOut` functions. Note that `logIn` won't behave the same way the mobile SDKs do, since they won't alias users together.
- Sets the `PlatformInfo` in the web SDK to have proper headers.
- Adds a new npm task to run all checks in one command

### DIFF
--- a/purchases-js-hybrid-mappings/api-report/api.json
+++ b/purchases-js-hybrid-mappings/api-report/api.json
@@ -199,7 +199,7 @@
                 },
                 {
                   "kind": "Content",
-                  "text": "{\n        apiKey: string;\n        appUserId: string | undefined;\n    }"
+                  "text": "{\n        apiKey: string;\n        appUserId: string | undefined;\n        flavor: string;\n        flavorVersion: string;\n    }"
                 },
                 {
                   "kind": "Content",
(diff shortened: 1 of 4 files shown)


purchases-js-hybrid-mappings/api-report/api.json
ajpallares, line 1: Just curious: this is autogenerated, right?

purchases-js-hybrid-mappings/src/purchases-common.ts
ajpallares, line 1: I'm sure this is explained by your PR description comment
> Note that logIn won't behave the same way the mobile SDKs do, since they won't alias users together.
But I still don't understand. Why don't the login/logout methods call the login/logout methods in the SDKs?

ajpallares (commented):
Looks good to me, but I don't really understand why the difference with the mobile SDKs

purchases-js-hybrid-mappings/src/purchases-common.ts
ajpallares: Ohhh makes sense to me now. Thanks so much for the explanation! 🫶

ajpallares (approved):

PR #1116, opened 2025-04-14
[purchases-js-hybrid-mapping] Add `purchasePackage` function

tonidero:
This adds a function to perform purchases, by passing a package ID + presented offering context.


purchases-js-hybrid-mappings/src/purchases-common.ts
```diff
@@ -80,6 +89,76 @@ export class PurchasesCommon {
     }
   }
 
+  public async purchasePackage(purchaseParams: {
+    packageIdentifier: string;
+    presentedOfferingContext: Record<string, unknown>;
+    optionIdentifier?: string;
+    customerEmail?: string;
+    selectedLocale?: string;
+    defaultLocale?: string;
```
tonidero, line 98: We will need to add these new parameters to the hybrid APIs... for now, adding them here doesn't hurt.

purchases-js-hybrid-mappings/src/purchases-common.ts
```diff
@@ -80,6 +89,76 @@ export class PurchasesCommon {
...
+      purchaseParams.presentedOfferingContext['offeringIdentifier'];
+    if (!presentedOfferingIdentifier || typeof presentedOfferingIdentifier !== 'string') {
+      const purchasesError = new PurchasesError(
+        ErrorCode.PurchaseInvalidError,
+        'Need to provide a valid offering identifier',
+      );
+      this.handleError(purchasesError);
+    }
+    try {
+      const offerings = await this.purchases.getOfferings();
```
tonidero, line 110: TBH, this is not ideal since there is a very small edge case of products changing between the time they are obtained, and the time they are purchased, since purchases-js doesn't cache this data... However, this edge case can potentially exist in android/ios as well, just that the chances are reduced because we have a cache there.

I believe this should be mostly ok, since we don't allow price changes at the moment though. Lmk what you think!

purchases-js-hybrid-mappings/src/purchases-common.ts
JayShortway: - Just curious, how do those locale parameters work? 
- Can the new parameters have default values, at least in the hybrids?

purchases-js-hybrid-mappings/src/purchases-common.ts
JayShortway: Hmm, that does feel like a future bug report that can lead to a rabbit hole. Would it be possible to change the parameter to a package, instead of a package ID? That way we don't need to find the packages again.

JayShortway (commented):
Just some concerns about refetching the products. Otherwise looks okay!

purchases-js-hybrid-mappings/src/purchases-common.ts
tonidero: The `selectedLocale` allows a developer the capability to select the language/locale used during the checkout process. The `defaultLocale` allows a developer the capability to set a fallback locale, in case the `selectedLocale` isn't found (which we default to english ourselves if not set)

> Can the new parameters have default values, at least in the hybrids?

Well, I think these don't apply to iOS + Android, since we don't have any control over the checkout process language there. My idea was to document that these will only apply to web in the hybrids. In web, they already have defaults, which is to pick the locale from the browser's settings. Lmk if you were referring to something else!

purchases-js-hybrid-mappings/src/purchases-common.ts
tonidero: It's complicated... Basically the package that the hybrids expect doesn't match what `purchases-js` expects normally... We would need to add some additional data in the hybrids to hold all the data we need from purchases-js and perform some double-mapping from/to what purchases-js returns/expects... I believe this can lead to some more dirty code IMO. 

Also, as I mentioned, this does potentially exist in iOS and Android already, where we basically do the same thing (refetch offerings and get the package using the package id + presented offering context info), with the main concern being that web doesn't cache offerings, so changes of this edge case happening could be increased.

purchases-js-hybrid-mappings/src/purchases-common.ts
JayShortway: Thanks for explaining! Yea as long as non-web hybrids don't have to explicitly pass `null` to these parameters, I think it's fine.

purchases-js-hybrid-mappings/src/purchases-common.ts
JayShortway: Nice, yea I like it!

JayShortway (approved):
Thanks for being thorough! 🚀

PR #1118, opened 2025-04-14
[purchases-js-hybrid-mappings] Add remaining APIs mappings

tonidero:
This adds the remaining APIs from purchases-js that we want to expose to hybrids at least for now, there are likely more once we start using this library. This includes:
- getCurrentOfferingForPlacement
- getAppUserId
- isSandbox
- setLogLevel

### DIFF
--- a/purchases-js-hybrid-mappings/src/mappers/log_level_mapper.ts
+++ b/purchases-js-hybrid-mappings/src/mappers/log_level_mapper.ts
@@ -1,5 +1,5 @@
 import { LogLevel } from '@revenuecat/purchases-js';
-import { Logger } from 'src/utils/logger';
+import { Logger } from '../utils/logger';
 
 export function mapLogLevel(logLevel: string): LogLevel | null {
   const upperCaseLogLevel = logLevel.toUpperCase();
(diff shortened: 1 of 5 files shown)


purchases-js-hybrid-mappings/src/mappers/log_level_mapper.ts
```diff
@@ -0,0 +1,23 @@
...
+    case 'DEBUG':
+      return LogLevel.Debug;
+    case 'INFO':
+      return LogLevel.Info;
+    case 'WARN':
+      return LogLevel.Warn;
+    case 'ERROR':
+      return LogLevel.Error;
+    case 'SILENT':
+      return LogLevel.Silent;
```
JayShortway, line 18: Interesting, we don't have this in other platforms do we?

JayShortway (approved):
Makes sense!

purchases-js-hybrid-mappings/src/mappers/log_level_mapper.ts
tonidero: Nope we don't. This is the default on web. I think we probably won't support it immediately on the hybrids until we also support a way to suppress all messages in iOS/Android. But wanted to add it here for completeness

PR #1120, opened 2025-04-15
[purchases-js-hybrid-mappings] Add rollup to generate minified library outputs

tonidero:
This condenses the outputs of the library into a few minified files, for less size and complexity.


JayShortway (approved):
R8-all-the-things 🙌  😛

PR #1127, opened 2025-04-17
[purchases-js-hybrid-mappings] Finalize last APIs + export UMD files

tonidero:
Last batch of fixes to get flutter web working, at least for the common cases. This PR includes:
- Add some missing APIs like: `isAnonymous`, `isConfigured`, `setProxyUrl`
- Fixes some existing APIs where Flutter was expecting non null values. Now we default to some values, same way as in Android/iOS
- Adds extra export so we export UMD files which will be used by flutter web, instead of the raw js output.
- Fixes some tests that were completely wrong/incomplete.

### DIFF
--- a/purchases-js-hybrid-mappings/package.json
+++ b/purchases-js-hybrid-mappings/package.json
@@ -3,6 +3,7 @@
   "version": "13.29.0",
   "description": "RevenueCat's JavaScript hybrid mappings for purchases-js",
   "main": "dist/index.js",
+  "browser": "dist/index.umd.js",
   "type": "module",
   "files": [
     "dist"
@@ -29,7 +30,7 @@
     "allchecks": "npm run format && npm run lint:fix && npm run api-test && npm run test"
   },
   "dependencies": {
-    "@revenuecat/purchases-js": "^1.1.0"
+    "@revenuecat/purchases-js": "^1.2.0"
   },
   "devDependencies": {
     "@microsoft/api-extractor": "^7.42.3",
--- a/purchases-js-hybrid-mappings/src/mappers/offerings_mapper.ts
+++ b/purchases-js-hybrid-mappings/src/mappers/offerings_mapper.ts
@@ -25,7 +25,7 @@ export function mapOffering(offering: Offering): Record<string, unknown> {
   return {
     identifier: offering.identifier,
     serverDescription: offering.serverDescription,
-    metadata: offering.metadata,
+    metadata: offering.metadata ?? {},
     availablePackages: offering.availablePackages.map(pkg => mapPackage(pkg)),
     lifetime: offering.lifetime ? mapPackage(offering.lifetime) : null,
     annual: offering.annual ? mapPackage(offering.annual) : null,
@@ -53,7 +53,7 @@ function mapProduct(product: Product): Record<string, unknown> {
   const defaultOptionBasePricingPhase = product.defaultSubscriptionOption?.base;
   return {
     identifier: product.identifier,
-    description: product.description,
+    description: product.description ?? '',
     title: product.title,
     price: product.currentPrice.amountMicros / 1_000_000,
     priceString: product.currentPrice.formattedPrice,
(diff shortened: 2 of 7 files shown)


purchases-js-hybrid-mappings/src/mappers/offerings_mapper.ts
```diff
@@ -93,7 +91,7 @@ function mapIntroPrice(product: Product): Record<string, unknown> | null {
 
   return {
     price: 0,
-    priceString: trialPhase.price?.formattedPrice,
+    priceString: trialPhase.price?.formattedPrice ?? '0.00', // TODO: Improve how we get a formatted price for trials
```
tonidero, line 94: This is slightly tricky... In purchases-js we don't have a currency for a free trial, nor a localized formatted text saying "Free". I think we can use the "base" pricing phase currency in order to at least put the currency here, like "$0.00", but will do that in the future. Want to get this out in beta soon.

purchases-js-hybrid-mappings/src/purchases-common.ts
```diff
@@ -37,9 +40,14 @@ export class PurchasesCommon {
...
       version: configuration.flavorVersion,
     });
+    let httpConfig: HttpConfig | undefined = undefined;
+    if (PurchasesCommon.proxyUrl) {
+      httpConfig = { proxyURL: PurchasesCommon.proxyUrl };
+    }
     const purchasesInstance = Purchases.configure(
       configuration.apiKey,
       configuration.appUserId || Purchases.generateRevenueCatAnonymousAppUserId(),
+      httpConfig,
```
ajpallares, line 50: I'm not too sure about this, but it looks to me like the `httpConfig` parameter cannot be `undefined`, as declared in the `configure` method of `Purchases` that I checked [here](https://github.com/RevenueCat/purchases-js/blob/25af11e92612ca6d23e709c0b843c5a02aa9a250/src/main.ts#L205-L210).
```
static configure(
    apiKey: string,
    appUserId: string,
    httpConfig: HttpConfig = defaultHttpConfig,
    flags: FlagsConfig = defaultFlagsConfig,
  ): Purchases
```

ajpallares (commented):
It mostly looks good to me, but it probably is a good idea to have someone with more knowledge approve this.

purchases-js-hybrid-mappings/package.json
```diff
@@ -29,7 +30,7 @@
     "allchecks": "npm run format && npm run lint:fix && npm run api-test && npm run test"
   },
   "dependencies": {
-    "@revenuecat/purchases-js": "^1.1.0"
+    "@revenuecat/purchases-js": "^1.2.0"
```
JayShortway, line 33: Is this going to be automated? Just curious.

purchases-js-hybrid-mappings/src/mappers/offerings_mapper.ts
JayShortway: That would be good yea!

purchases-js-hybrid-mappings/src/purchases-common.ts
tonidero: I believe `undefined` is the "default" value if the parameter is not passed, in which case, it will pick the default value provided, in this case `defaultHttpConfig`.

JayShortway (commented):
Looks good! No other comments besides Antonio's.

purchases-js-hybrid-mappings/package.json
tonidero: Yes it should! Haven't added the automations yet, but we shouldn't care about bumping these versions soon. Addressing some other problems first.

purchases-js-hybrid-mappings/src/purchases-common.ts
ajpallares: mmm I see your point. I guess this is different from how default method parameters work in Swift 😅 
Thanks for the clarification!

JayShortway (approved):

PR #1130, opened 2025-04-22
[purchases-js-hybrid-mappings] Fix `logIn` result mapping

tonidero:
This PR has 2 changes:
- We were not passing the expected result in Flutter for the `logIn` request. This fixes that. One note however is that, for mobile SDKs, we do have a proper `logIn` method, that returns whether the user was created or not by using the response status code. In `purchases-js`, we don't provide a logIn function, only a changeUser, which returns a basic `CustomerInfo`. I think we will need to deprecate that method in `purchases-js` and add a new one that returns an object including this information, however, if we only change the return type, it would be a conflicting declaration, so we would need to rename it 😞. For now, I'm hardcoding this value to `false` until we get this information.
- Adds missing `close` function to close the configured SDK

### DIFF
--- a/purchases-js-hybrid-mappings/api-report/api.json
+++ b/purchases-js-hybrid-mappings/api-report/api.json
@@ -190,6 +190,42 @@
           "members": [
             {
               "kind": "Method",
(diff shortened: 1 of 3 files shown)


JayShortway (approved):
Makes sense!

PR #1132, opened 2025-04-23
Automate update of `purchases-js` in `purchases-js-hybrid-mappings`

tonidero:
This will make sure we also update the version of `purchases-js` in PHC whenver a new version changes.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -314,7 +314,7 @@ private_lane :update_android_native_version do |options|
   )
 end
 
-desc "Update purchases-android dependency version"
+desc "Update purchases-js dependency version"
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -295,6 +314,19 @@ private_lane :update_android_native_version do |options|
   )
 end
 
+desc "Update purchases-android dependency version"
```
JayShortway, line 317:
```suggestion
desc "Update purchases-js dependency version"
```

JayShortway (approved):
Amazing!!

fastlane/Fastfile
tonidero: Whoops haha thanks!

PR #1138, opened 2025-05-04
Add `Storefront` type to typescript interfaces

tonidero:
This will be used by RN and Capacitor SDKs. Adds this missing work from #1125


ajpallares (commented):
I'm sure I'm missing it, but although I see the `StoreFront` declared, I don't see it used in any other public API

tonidero:
> I'm sure I'm missing it, but although I see the StoreFront declared, I don't see it used in any other public API

That's correct @antoniobg! The typescript library only declares types which are currently used by our RN and Capacitor SDKs so we share the data types between them, minimizing API changes. This change won't do anything (other than adding a new type) until it's actually used in the RN and Capacitor SDKs.

ajpallares (approved):
> That's correct @ajpallares! The typescript library only declares types which are currently used by our RN and Capacitor SDKs so we share the data types between them, minimizing API changes. This change won't do anything (other than adding a new type) until it's actually used in the RN and Capacitor SDKs.

Thanks for the explanation @tonidero. It makes sense!

PR #1147, opened 2025-05-27
Store `appUserId` in `localStorage` for web hybrid support

tonidero:
There was a difference in behavior in Flutter web from Flutter mobile. That is, in mobile, we store the user id in disk, so we reuse the same even if the app is reconfigured normally, but we don't store it on web. This adds some logic to store the user id in local storage on web hybrids, so the behavior is consistent.

### DIFF
--- a/purchases-js-hybrid-mappings/src/purchases-common.ts
+++ b/purchases-js-hybrid-mappings/src/purchases-common.ts
@@ -20,6 +20,7 @@ import { mapLogLevel } from './mappers/log_level_mapper';
 export class PurchasesCommon {
   private static instance: PurchasesCommon | null = null;
   private static proxyUrl: string | null = null;
+  private static readonly STORAGE_KEY = 'revenuecat_user_id';
 
   private purchases: Purchases;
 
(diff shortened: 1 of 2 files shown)


purchases-js-hybrid-mappings/src/purchases-common.ts
```diff
@@ -40,15 +41,33 @@ export class PurchasesCommon {
       flavor: configuration.flavor,
       version: configuration.flavorVersion,
     });
+
+    let appUserId: string;
+    if (configuration.appUserId !== undefined) {
+      appUserId = configuration.appUserId;
+      if (typeof window !== 'undefined') {
```
nicfix, line 48: Why don't we check for the existence of `localStorage` rather than `window`?

nicfix (commented):
Left a question

purchases-js-hybrid-mappings/src/purchases-common.ts
tonidero: That's a very good point 😅 changed in https://github.com/RevenueCat/purchases-hybrid-common/pull/1147/commits/b8bbb77841cdcfcc82f2db736250eb62c27d68f5

ajpallares (approved):
Makes sense!

PR #1150, opened 2025-06-03
Add `PADDLE` store

fire-at-will:
Depends on https://github.com/RevenueCat/purchases-ios/pull/4981 and https://github.com/RevenueCat/purchases-android/pull/2353. 

Adds the `PADDLE` store support to PHC. For Android, we don't need to map it since we use the enum's `name` property.

#### TODO
- [ ] Holding until natives have been deployed and updated in PHC.


fire-at-will:
iOS CI will fail until the native iOS SDK has been released + updated here

tonidero (approved):
:shipit:

PR #1165, opened 2025-06-17
Avoids passing multiple `PaywallResult`s to a single `PaywallResultListener`.

JayShortway:
## Bug
It is possible to have multiple paywalls "alive" (in memory) at the same time, for instance if Paywall A has a button with a deep link that opens Paywall B. 

## Cause
In some of our hybrids, we use an invisible `PaywallFragment` to handle receiving a result from the `PaywallActivity`. In the scenario described above, multiple `PaywallFragment`s are alive, one for each Paywall. The problem arises because they share an Activity-scoped `PaywallFragmentViewModel` instance to hold the `PaywallResultListener`. When a subsequent Paywall is launched, that Paywall's `PaywallResultListener` is set in the shared `PaywallFragmentViewModel`, which all `PaywallFragment`s will now use. (The `PaywallFragmentViewModel` has to be Activity-scoped, because the listener had to be set before the `PaywallFragment` is created.) This causes all of the Paywalls to report their result to the same `PaywallResultListener` instance.  
  
In purchases-flutter, this results in crashes like this: 
```
IllegalStateException: Reply already submitted
```

- https://github.com/RevenueCat/purchases-flutter/issues/1350
- [Play SDK Console](https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/5270760444858964624/crashes/ccfdc45a/details?timeRange=LAST_28_DAYS&affectedApps=AFFECTED_APPS_FIVE_OR_MORE&crashStatuses=SDK_CRASH_STATUS_UNSPECIFIED)

## Fix
The fix implemented in this PR is to use a `FragmentResultListener` tied to the specific `PaywallResultListener`. This means that every `PaywallFragment` only communicates with its own `PaywallResultListener`. This adheres to the "one result per Paywall" contract.

### DIFF
--- a/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallHelpers.kt
+++ b/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallHelpers.kt
@@ -13,14 +13,14 @@ import com.revenuecat.purchases.ui.revenuecatui.fonts.PaywallFontFamily
     ),
 )
 fun presentPaywallFromFragment(
-    fragment: FragmentActivity,
+    activity: FragmentActivity,
+    paywallResultListener: PaywallResultListener,
     requiredEntitlementIdentifier: String? = null,
-    paywallResultListener: PaywallResultListener? = null,
     shouldDisplayDismissButton: Boolean? = null,
     offering: Offering? = null,
 ) {
     presentPaywallFromFragment(
-        fragment,
+        activity,
         PresentPaywallOptions(
             requiredEntitlementIdentifier = requiredEntitlementIdentifier,
             paywallResultListener = paywallResultListener,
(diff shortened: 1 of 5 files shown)


android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallHelpers.kt
```diff
@@ -31,18 +31,26 @@ fun presentPaywallFromFragment(
...
 fun presentPaywallFromFragment(
-    fragment: FragmentActivity,
+    activity: FragmentActivity,
     options: PresentPaywallOptions,
 ) {
     with(options) {
-        fragment
+        val requestKey = System.identityHashCode(paywallResultListener).toString()
+
+        activity.supportFragmentManager.setFragmentResultListener(requestKey, activity) { _, result ->
```
tonidero, line 40: I'm wondering, does this survive activity recreation? Just making sure since I think that was the main reason to using a VM...

PR #1169, opened 2025-06-18
Maps potentially large objects on a background thread to tackle ANRs

JayShortway:
## Bug
Hybrids on Android are experiencing ANRs when certain objects get too big:
- Offerings ([MON-1108](https://linear.app/revenuecat/issue/MON-1108/android-anrs-due-to-too-many-offeringsmetadata-on-main-thread), [Play SDK Console](https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/9178691880686914701/crashes/59e45236/details?timeRange=LAST_28_DAYS&affectedApps=AFFECTED_APPS_FIVE_OR_MORE&errorTypes=ANR&crashStatuses=SDK_CRASH_STATUS_UNSPECIFIED))
- CustomerInfo ([Play SDK Console](https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/9178691880686914701/crashes/d7d319fb/details?timeRange=LAST_28_DAYS&affectedApps=AFFECTED_APPS_FIVE_OR_MORE&errorTypes=ANR&crashStatuses=SDK_CRASH_STATUS_UNSPECIFIED))
- StoreProducts ([Play SDK Console](https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/9178691880686914701/crashes/6a5ed36d/details?timeRange=LAST_28_DAYS&affectedApps=AFFECTED_APPS_FIVE_OR_MORE&errorTypes=ANR&crashStatuses=SDK_CRASH_STATUS_UNSPECIFIED))

## Cause
One of the causes of this, is that we convert these objects to a map on the main thread. If that takes too long, the OS will consider the app non-responsive. 

## Fix
The fix in this PR is to do the mapping on a background thread.

-------------
# Note 👇 
~~Before I go and fix tests,~~ would love to get opinions on using Coroutines for this. I figured we have it in purchases-android already, so there's no change in using it here too. Happy to cobble a thread pool together though if that is preferred.


android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/CustomerInfoMapper.kt
```diff
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.hybridcommon.mappers
 
 import com.revenuecat.purchases.CustomerInfo
+import kotlinx.coroutines.launch
 
 fun CustomerInfo.map(): Map<String, Any?> =
```
JayShortway, line 6: What are your thoughts on making the sync mappers private? I noticed they are part of our API tests. Are some of our hybrids using these directly?

tonidero: I believe we use some... for example: https://github.com/RevenueCat/purchases-flutter/blob/7a21f7f11ab0069dfd38b06afcf920548b5fe19d/android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java#L394

I do think we can potentially remove those usages in favor of new wrappers in PHC, to avoid exposing them, so I wouldn't be opposed to removing usages of those and make these internal TBH.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -643,21 +652,23 @@ fun redeemWebPurchase(
         )
         when (result) {
             is RedeemWebPurchaseListener.Result.Success -> {
-                resultMap["customerInfo"] = result.customerInfo.map()
+                result.customerInfo.mapAsync { map ->
+                    resultMap["customerInfo"] = map
+                    onResult.onReceived(resultMap)
```
tonidero, line 657: If we're moving this to each case, I think we probably don't need `resultMap`? We can probably just create a new map with the relevant data.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/mappersHelpers.kt
```diff
@@ -44,6 +47,8 @@ fun JSONObject.convertToMap(): Map<String, String?> =
         }
     }
 
+internal val mapperScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
tonidero, line 50: ❤️

tonidero (commented):
I like this! I think it might be good to revisit the hybrids to check for any usages of the mapper functions and try to make them private, so we make sure we take full advantage of the new method. Then make the mappers private. I guess that might take a bit longer though, so I'm ok if we want to tackle that separately.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
JayShortway: Good call, done here: https://github.com/RevenueCat/purchases-hybrid-common/pull/1169/commits/891344c03c981b4a93454c1028ac258c3e6584c8

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/mappersHelpers.kt
JayShortway: Had to change it slightly to fix tests: https://github.com/RevenueCat/purchases-hybrid-common/pull/1169/commits/ac05bd98ddba5b1353e003681755ba0613f6ebf7. Normally you would dependency-inject the `CoroutineDispatcher`, but that's not really possible without modifying the public API (which would add Coroutines to our public API).

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/mappers/CustomerInfoMapper.kt
JayShortway: Alright, I made the sync mappers private here: https://github.com/RevenueCat/purchases-hybrid-common/pull/1169/commits/632adaa38fe4542f375ab7d2a9fde180d2b87b4c. Got PRs lined up in the hybrids:
- https://github.com/RevenueCat/purchases-flutter/pull/1373
- https://github.com/RevenueCat/react-native-purchases/pull/1295
- https://github.com/RevenueCat/purchases-capacitor/pull/471
- https://github.com/RevenueCat/purchases-unity/pull/604
- https://github.com/RevenueCat/cordova-plugin-purchases/pull/678

I'll mark this PR as `force_major`, as this is an API-breaking change.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -399,8 +405,8 @@ fun getStorefront(
 fun restorePurchases(
     onResult: OnResult,
 ) {
-    Purchases.sharedInstance.restorePurchasesWith(onError = { onResult.onError(it.map()) }) {
-        onResult.onReceived(it.map())
+    Purchases.sharedInstance.restorePurchasesWith(onError = { onResult.onError(it.map()) }) { customerInfo ->
```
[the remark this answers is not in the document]
JayShortway: Yea, true! Errors are (so far) unlikely to block the main thread I think.

PR #1173, opened 2025-06-23
Update PHC with iOS VC Updates

fire-at-will:
This PR updates the PHC library to support the Virtual Currency iOS API changes in https://github.com/RevenueCat/purchases-ios/pull/5151. It is meant to be merged into the PHC library when we update the library to contain the API changes from https://github.com/RevenueCat/purchases-ios/pull/5151 that will be released in an upcoming release.

Specifically, this PR:

- Adds `CommonFunctionality.virtualCurrencies` to fetch virtual currency information
- Adds `CommonFunctionality.invalidateVirtualCurrenciesCache` to support clearing the virtual currencies cache
- Adds a hybrid dictionary for `VirtualCurrencies` and `VirtualCurrency`
- Adds automated & API tests
- Adds the new virtual currency objects to the typescript API

This PR will be needed when https://github.com/RevenueCat/purchases-ios/pull/5151 is shipped to avoid breaking the PHC `MockPurchase` object's compilation due to the new functions.

## TODO
For now, we're referencing the `break-vc-balances-out-of-customerInfo` branch for the iOS native dependency to ensure that things are building. When this is merged, this will need to be updated to the correct iOS version.

> Note: Since podspec files don't support specifying a `:git` dependency source, the `check-pods` job will not pass on this PR until https://github.com/RevenueCat/purchases-ios/pull/5151 has been released.


ajpallares (approved):
Makes sense! 🚀

PR #1174, opened 2025-06-24
Calls `setFragmentResultListener()` on the UI thread.

JayShortway:
## Description
It is not guaranteed that `presentPaywallFromFragment` is called on the UI thread, but `setFragmentResultListener()` **must** be, otherwise it will crash. This ensures that that's the case.


ajpallares (approved):
Thanks for fixing this!

vegaro (approved):

PR #1190, opened 2025-07-08
Fix clear background of paywalls in iOS

ajpallares:
Paywalls had a transparent background in iOS. This would lead to situations like these

This PR fixes it, by setting the default background of the paywall to the system background color in iOS

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/PaywallProxy.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/PaywallProxy.swift
@@ -182,6 +182,7 @@ import UIKit
 
         controller.delegate = self
         controller.modalPresentationStyle = .pageSheet
+        controller.view.backgroundColor = .systemBackground
 
         if let paywallResultHandler {
             self.resultByVC[controller] = (paywallResultHandler, .cancelled)


ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/PaywallProxy.swift
```diff
@@ -182,6 +182,7 @@ import UIKit
 
         controller.delegate = self
         controller.modalPresentationStyle = .pageSheet
+        controller.view.backgroundColor = .systemBackground
```
tonidero, line 185: Ohhh that explains it! I think that makes sense... I do wonder if this could affect any paywalls if no background is set... But maybe this would only affect the loading paywall?

tonidero (approved):
I think this makes sense! Might be good to give it a quick test on the hybrids, but I don't think it should cause any issues...

ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/PaywallProxy.swift
ajpallares: Yes, this also fixes it for any paywalls. But I think that's the desired behavior

PR #1193, opened 2025-07-08
Ensures async callbacks are called on the main thread.

JayShortway:
## Description
As a result of https://github.com/RevenueCat/purchases-hybrid-common/pull/1169, callbacks such as the PaywallListener could end up not being called on the main thread. At least Flutter doesn't like that: https://github.com/RevenueCat/purchases-flutter/issues/1395. This PR fixes that by moving back to the main thread before calling the callback.


vegaro (approved):
Makes sense!

PR #1204, opened 2025-07-17
Use spi(Interal) Import RC in Tests

fire-at-will:
https://github.com/RevenueCat/purchases-ios/pull/5384 updated the VirtualCurrency object initializers to be @spi(Internal), which broke the compilation of a few tests in PHC. This PR updates the affected test files to use `@spi(Internal) @testable import RevenueCat` to fix the compilation of the tests.


tonidero (approved):
🚢

PR #1206, opened 2025-07-17
Expose VC APIs in Android PHC

fire-at-will:
This PR exposes the following functions to Android PHC clients from the native Android SDK:
- `getVirtualCurrencies()`
- `invalidateVirtualCurrenciesCache()`
- `getCachedVirtualCurrencies()`

The `VirtualCurrency` and `VirtualCurrencies` typescript objects were already introduced in https://github.com/RevenueCat/purchases-hybrid-common/pull/1173, so we don't need to add them here.


tonidero (approved):
🚢

PR #1209, opened 2025-07-17
Update iOS VC APIs

fire-at-will:
### Description
In preparation for adding Virtual Currency functionality to the hybrid SDKs, this PR makes a few changes to the iOS VC APIs exposed through PHC:
- Exposes a new `getCachedVirtualCurrencies()` function. I decided to make this a function instead of a computed property to minimize the risk of running into compatibility issues in any of the hybrid SDKs when trying to access a computed property.
- Updates the signature of the `getVirtualCurrencies()` function's completion handler to return `[String: Any]?` instead of a `VirtualCurrencies` object. 
   - The `getVirtualCurrencies()` function was introduced in https://github.com/RevenueCat/purchases-hybrid-common/pull/1173 and mistakenly returned a `VirtualCurrencies` object instead of a `[String: Any]?` in its completion handler.
   - While this is a breaking change, it shouldn't break any of the individual hybrid SDKs, since they are not consuming the `getVirtualCurrencies()` function yet.
- Updates the existing tests around  `getVirtualCurrencies()` to support the new API signature
- Adds new tests for the `getCachedVirtualCurrencies()` function


tonidero (approved):
Makes sense!

PR #1210, opened 2025-07-18
[purchases-js-hybrid-mappings] Support `setAttributes`

tonidero:
This adds support for the `setAttributes` method and some reserved attributes methods.


ajpallares (approved):
I think it makes sense!

PR #1224, opened 2025-07-30
Add `getApiKey` API to internal purchases-js-hybrid-mappings

tonidero:
In order to support the test store in react native and expo go, we need to do some operations in the RN layer that require us to have the API key. We could cache it in the RN layer, but that's slightly trickier than storing it here. Also, this is meant to be an internal library, so I think it's ok to add it here.


ajpallares (approved):
Yes, I think it makes sense to have the `getApiKey` method here 👌

PR #1233, opened 2025-08-05
Run Tests on iOS 18 in CI

fire-at-will:
This PR updates the `purchases-hybrid-common` CI to run iOS tests on iOS 18. Specifically, it:
- Defaults to using Xcode 16.4.0 instead of 15.2.0
- Adds a `test-ios-18` job to be run when the iOS tests are executed
- Runs the iOS integration tests on iOS 18.5 instead of iOS 17.2
- Explicitly adds `abbrev` as a dependency in the `Gemfile`. Without it, the `check-pods` and `integration-tests-ios` jobs fail, since `abbrev` was "promoted from default gems" in Ruby 3.4.0, meaning that it is no longer a default gem ([source](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/))

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,26 @@ jobs:
           path: ios/PurchasesHybridCommon/test_output
           destination: scan-output
 
+  test-ios-18:
+    <<: *base-ios-job
+    steps:
+      - checkout
+      - revenuecat/install-gem-mac-dependencies:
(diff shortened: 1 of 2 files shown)


tonidero (approved):
🚢

ajpallares (approved):
Nice one! Thank you!

PR #1267, opened 2025-09-04
Add override preferred locale methods

joshdholtz:
## Motivation

Add override preferred locale methods

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/CommonFunctionality+Extensions.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/CommonFunctionality+Extensions.swift
@@ -1,0 +1,23 @@
+//
+//  CommonFunctionality+Extensions.swift
+//  PurchasesHybridCommon
+//
+//  Created by Josh Holtz on 9/3/25.
+//  Copyright © 2025 RevenueCat. All rights reserved.
+//
+
+#if !os(macOS) && !os(tvOS) && !os(watchOS)
+
+import PurchasesHybridCommon
+import RevenueCat
+import RevenueCatUI
+
+extension CommonFunctionality {
+
+    @objc public static func overridePreferredLocale(_ locale: String?) {
+        Purchases.shared.overridePreferredUILocale(locale)
+    }
+
+}
+
+#endif
--- a/purchases-js-hybrid-mappings/src/purchases-common.ts
+++ b/purchases-js-hybrid-mappings/src/purchases-common.ts
@@ -104,6 +104,10 @@ export class PurchasesCommon {
     }
   }
 
+  static overridePreferredLocale(locale: string | null): void {
+    Purchases.overridePreferredLocale(locale);
+  }
+
   private constructor(purchasesInstance: Purchases) {
     this.purchases = purchasesInstance;
   }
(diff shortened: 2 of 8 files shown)


ios/PurchasesHybridCommon/PurchasesHybridCommonUI/Paywalls/CommonFunctionality+Extensions.swift
```diff
@@ -0,0 +1,7 @@
+//
+//  CommonFunctionality+Extensions.swift
+//  PurchasesHybridCommon
+//
+//  Created by Josh Holtz on 9/3/25.
+//  Copyright © 2025 RevenueCat. All rights reserved.
+//
```
tonidero, line 7: File seems unused?

ios/PurchasesHybridCommon/PurchasesHybridCommon/Purchases+HybridAdditions.swift
```diff
@@ -78,12 +79,44 @@ import RevenueCat
             }
         }
 
+        configurationBuilder = configurationBuilder.with(preferredLocale: preferredLocale)
```
tonidero, line 82: Hmm interesting... it seems here we're calling this even if `preferredLocale` is null, but we don't on Android. I think it's probably not needed so we can follow what we do in android?

purchases-js-hybrid-mappings/src/purchases-common.ts
```diff
@@ -104,6 +104,15 @@ export class PurchasesCommon {
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  static overridePreferredLocale(_locale: string | null): void {
```
tonidero, line 108: Hmm do we need this method at all? If it's to just make it fall, we can configure that on each hybrid (at least, that's what we were doing for Flutter + react native. I would probably just remove all changes in `purchases-js-hybrid-mappings` but leave the ones in `purchases-typescript-internal`

tonidero (approved):
Almost there! Nothing really blocking, other than probably removing the new API in `purchases-js-hybrid-mappings` since I think we don't need it.

tonidero:
Ohh nice catch! But yeah, I do think these APIs kinda belong more on the core SDK, since they are related to configuration? Also, it could affect if someone were to use the Paywall types directly (even though they are meant to be internal 😅 )

joshdholtz:
@tonidero @ajpallares So if we do want to move it into the core, it should be just an change in iOS to move it over which... shouldn't be a _really_ breaking change since `RevenueCat` is needed anyway for `RevenueCatUI`

tonidero:
Hmm I'm a bit confused though... how does it work for us to add it in the core library in PHC, but not in the core of iOS? Afaik Core PHC only depends on Core iOS, so it shouldn't have access to anything there right?

ajpallares:
> @ajpallares Ahhhh... you are right 🤔 And now that I think about it... I might have put Android directly on `Purchases` 😬
> 
> I guess... does putting it on `PurchasesUI` have any real benefit? It seems to be a little bit more complicated to split up... especially for allowing it on a configuration (which is `Purchases`) 🤷‍♂️ If we allow it on configuration, it seems like its more of a core thing a should just be on `Purchases` everywhere? 🙃

Right, I understand. Sorry I couldn't take a look at the Android one 🙏 

> Ohh nice catch! But yeah, I do think these APIs kinda belong more on the core SDK, since they are related to configuration?

There's not really a benefit. The original logic was: since it's something that only affects RevenueCatUI (that's also why the public APIs contain the UI term like `preferredUILocaleOverride`), it made more sense to have it there. E.g. someone that does not use RevenueCatUI doesn't have to get autocomplete for the public APIs related to locale override (as they only affect RCUI). Does it make sense?

tonidero:
I think that makes sense! 🫶

PR #1268, opened 2025-09-04
Add internal `_purchaseSimulatedStorePackage`

tonidero:
This method will be used by RN in Expo Go mode when using simulated purchases.


fire-at-will (approved):
:shipit:

PR #1274, opened 2025-09-08
Use `preferredUILocaleOverride` name in `PurchasesConfiguration`

ajpallares:
This PR renames the property `overridePreferredLocale` of `PurchasesConfiguration` to `preferredUILocaleOverride` to make it the same as in the iOS and Android SDKs

### DIFF
--- a/typescript/src/purchasesConfiguration.ts
+++ b/typescript/src/purchasesConfiguration.ts
@@ -93,6 +93,7 @@ export interface PurchasesConfiguration {
 
   /**
    * Override the preferred UI locale for RevenueCat UI components at runtime. This affects both API requests and UI rendering.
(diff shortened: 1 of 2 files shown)


typescript/src/purchasesConfiguration.ts
```diff
@@ -93,9 +93,8 @@ export interface PurchasesConfiguration {
 
   /**
    * Override the preferred UI locale for RevenueCat UI components at runtime. This affects both API requests and UI rendering.
-   * If the locale changes, this will automatically clear the offerings cache and trigger a background refetch to get paywall templates with the correct localizations.
```
tonidero, line 96: Hmm maybe keep it but with something like:
```
This will automatically clear the offerings cache and trigger a background refetch to get paywall templates with the correct localizations.
```

tonidero (approved):
Looks good! Just a comment

PR #1278, opened 2025-09-10
Update iOS SDK to v5.38.1 and adapt to `@_spi` API changes

ajpallares:
This PR:
* Bumps the version of the iOS SDK to v5.38.1
* Adapts to the `@_spi` API changes in that version of the iOS SDK (regarding setting the preferred locale)


facumenzella (approved):
Looks good!

PR #1284, opened 2025-09-15
Add `PurchaseParams` interface

tonidero:
This adds an interface with all mandatory and optional parameters to initiate a purchase. Adding it here so it can be shared between RN and Capacitor.

### DIFF
--- a/typescript/src/purchaseParams.ts
+++ b/typescript/src/purchaseParams.ts
@@ -1,0 +1,38 @@
+import {
+  GoogleProductChangeInfo,
+  PurchasesPackage, PurchasesPromotionalOffer,
+  PurchasesStoreProduct,
+  PurchasesWinBackOffer,
+  SubscriptionOption
+} from "./offerings";
+
+/**
+ * Holds parameters to initialize a purchase in our SDK.
+ * @public
+ */
+export interface PurchaseParams {
+  /**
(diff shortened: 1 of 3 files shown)


typescript/src/purchaseParams.ts
```diff
@@ -0,0 +1,38 @@
...
+   * Google Play only. Optional boolean indicates personalized pricing on products available for purchase in the EU.
+   * For compliance with EU regulations. User will see "This price has been customized for you" in the purchase dialog when true.
+   * See https://developer.android.com/google/play/billing/integrate#personalized-price for more info.
+   */
+  googleIsPersonalizedPrice?: boolean | null;
+  /**
+   * iOS only, requires iOS 18.0 or greater with StoreKit 2. Win-back offer to apply to this purchase.
+   * Retrieve this using getEligibleWinBackOffersForPackage.
+   */
+  winBackOffer?: PurchasesWinBackOffer | null;
```
tonidero, line 33: I wondered about whether to add this... I see in RN, we added separate functions to make this mandatory in those cases... Is there a reason for that or would it be ok to make it an optional parameter like in here? @fire-at-will (This will also inform Flutter, where I didn't move it to `PurchaseParams` [initially](https://github.com/RevenueCat/purchases-flutter/commits/main/)... but I'm thinking we should..

PR #1286, opened 2025-09-15
Implement setLogHandler for web

antoniobg:
Adds support for `setLogHandler` for web

### DIFF
--- a/purchases-js-hybrid-mappings/api-report/api.json
+++ b/purchases-js-hybrid-mappings/api-report/api.json
@@ -1,7 +1,7 @@
 {
   "metadata": {
     "toolPackage": "@microsoft/api-extractor",
-    "toolVersion": "7.52.3",
+    "toolVersion": "7.52.13",
     "schemaVersion": 1011,
     "oldestForwardsCompatibleVersion": 1001,
     "tsdocConfig": {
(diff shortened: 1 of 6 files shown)


purchases-js-hybrid-mappings/api-report/api.json
```diff
@@ -182,7 +182,7 @@
               "text": "declare class PurchasesCommon "
             }
           ],
-          "fileUrlPath": "src/index.ts",
+          "fileUrlPath": "dist/index.d.ts",
```
antoniobg, line 185: Not sure if this change is a problem

purchases-js-hybrid-mappings/package.json
```diff
@@ -30,7 +30,7 @@
     "allchecks": "npm run format && npm run lint:fix && npm run api-test && npm run test"
   },
   "dependencies": {
-    "@revenuecat/purchases-js": "1.13.2"
+    "@revenuecat/purchases-js": "1.14.0"
```
tonidero, line 33: Hmm we could trigger an update of the natives here instead of just updating purchases-is… you can do that from CircleCI dependency-update action

purchases-js-hybrid-mappings/api-report/api.json
tonidero: Well this change itself it’s not a problem since we just use this to diff the public API… I do wonder why it changed though 🤔

tonidero (approved):
I would just say maybe to split the update of the natives to a separate PR which can be created automatically by CircleCI, so we get the latest updates for everything… though having said that, it’s not a blocker for this PR so looks good!

PR #1291, opened 2025-09-17
Add purchase method with options map to Android

tonidero:
As we're adding new purchase functions with a "builder-like" pattern in hybrids, having a function that accepts a free-form map of options and includes the validation of the parameter becomes more useful. The downside is not having compile-time safety... However, it will remove logic from the hybrids which is a good thing.


android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -105,6 +106,131 @@ fun getProductInfo(
...
+fun purchase(
+    activity: Activity?,
+    options: Map<String, Any?>,
+    onResult: OnResult,
+) {
+    val purchaseParams = validatePurchaseParams(options).onFailure {
+        onResult.onError((it as PurchasesException).error.map())
+    }.getOrNull() ?: return
+    when (val purchasableItem = purchaseParams.purchasableItem) {
+        is PurchasableItem.Product -> purchaseProduct(
```
tonidero, line 118: I'm using the existing purchase functions since they all have some logic in them and it was tricky to consolidate them... The idea is to make the hybrids start using the new function, and once all hybrids have updated, we can remove them (which would be a breaking change). I just didn't want to force us to update ALL hybrids at the same time right now.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -105,6 +106,131 @@ fun getProductInfo(
     }
 }
 
+fun purchase(
+    activity: Activity?,
+    options: Map<String, Any?>,
```
tonidero, line 111: As I mentioned, the main downside of this is that we lose any type of compile time safety, in exchange of unifying the logic of parsing the parameters to PHC. Lmk if you have any concerns over the approach!

ajpallares (commented):
I added some initial comments with my opinion about the new approach. Will continue with the review right away!

PR #1307, opened 2025-09-26
[CI] Only `build` pipeline action can trigger `deploy` workflow

ajpallares:
When triggering the CI pipeline with the `bump` action, what we intend is to deploy **the next version** of this dependency. Therefore, the `deploy` workflow should never be triggered in this scenario, since it starts the deployment of **the current version** of the dependency.

This happened to me when trying to release a new version for an older major of PHC. I wanted to release PHC v14.3.0 starting off of v14.2.0. For that I created a `release/14.2.0` branch at the `14.2.0` tag and run the pipeline with the `bump` action. That triggered the `trigger_bump` workflow (as expected) that created the `14.3.0` tag. However, it also triggered the `deploy` workflow because the branch was called `release/X.Y.Z`. 

With this new condition, the `deploy` workflow would'n have been triggered, which is what we want.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -497,8 +497,11 @@ workflows:
       and:
       - not:
           equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
-      - not:
-          equal: [ bump, << pipeline.parameters.action >> ]
+      # Only run this workflow when the action is "build" (the default), 
+      # so it doesn't run for other actions like "bump" on a release branch,
+      # which should only open the release PR but not deploy it.
+      # (Deployment happens after the hold job is approved on the release PR.)
+      - equal: [ build, << pipeline.parameters.action >> ]
     jobs:
       - hold:
           type: approval


.circleci/config.yml
```diff
@@ -494,8 +494,11 @@ workflows:
 
   deploy:
     when:
-      not:
-        equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+      and:
+      - not:
+          equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+      - not:
+          equal: [ bump, << pipeline.parameters.action >> ]
```
JayShortway, line 501: I'm a bit confused. `bump` should `deploy` right? When releasing hotfixes, I believe I usually branch off of the release I want to hotfix, e.g. `release/14.2.0` --> `base/14.2.0`, make changes, and then run `bump` on `base/14.2.0`. (This is from memory, but it's something like that.)

JayShortway (approved):
A bit more safety!

PR #1622, opened 2026-04-29
New Unified StoreReplacementMode APIs

fire-at-will:
Introduces a new unified APIs for product changes that can be used across both the Galaxy and Play Store. 

## New APIs
This PR introduces two new typescript APIs:
### STORE_REPLACEMENT_MODE
```typescript
export enum STORE_REPLACEMENT_MODE {
    CHARGE_FULL_PRICE = "CHARGE_FULL_PRICE",
    CHARGE_PRORATED_PRICE = "CHARGE_PRORATED_PRICE",
    DEFERRED = "DEFERRED",
    WITH_TIME_PRORATION = "WITH_TIME_PRORATION",
    WITHOUT_PRORATION = "WITHOUT_PRORATION"
}
```

### StoreProductChangeInfo
```typescript
export interface StoreProductChangeInfo {
    readonly oldProductIdentifier: string;
    readonly prorationMode?: STORE_REPLACEMENT_MODE;
}
```

The `STORE_REPLACEMENT_MODE` API matches the API found in the native Android SDK's `StoreReplacementMode`.

### PurchaseParams.productChangeInfo
```typescript
export interface PurchaseParams {
  // ...
  productChangeInfo?: StoreProductChangeInfo | null;
}
```

## Migration from .googleReplacementMode() to .storeReplacementMode()
The PR also migrates `common.kt` from using the now-deprecated `GoogleReplacementMode`/`PurchaseParams.Builder.googleReplacementMode()` APIs to instead use the new `StoreReplacementMode`/`PurchaseParams.Builder.replacementMode()` APIs from the native Android SDK.

If both a `googleReplacementMode` and a `storeReplacementMode` are provided, the `googleReplacementMode` is ignored and the `storeReplacementMode` is used.

## Tests
- Added plenty of unit tests for the changes in `common.kt`.
- Added API tests for the new purchase function variations in `common.kt`.
- Linked this branch into the RN SDK, and everything still compiled :)

## Other Notes
This PR intentionally does not deprecate the existing `GoogleProductChangeInfo`, `PRORATION_MODE`, or `PurchaseParams.googleProductChangeInfo` APIs, since we haven't yet introduced functions in the hybrid SDKs that consume the newly-introduced APIs. We will deprecate `GoogleProductChangeInfo`, `PRORATION_MODE`, and `PurchaseParams.googleProductChangeInfo` in a future PR once the new product change APIs are fully available.


fire-at-will:
@claude review

typescript/src/purchaseParams.ts
```diff
@@ -20,6 +21,11 @@ export interface PurchaseParams {
    * wish to upgrade from containing the oldProductIdentifier and the optional prorationMode.
    */
   googleProductChangeInfo?: GoogleProductChangeInfo | null;
+  /**
+   * Google Play and Galaxy Store only. Optional {@link StoreProductChangeInfo} that allows you
+   * to make a product change with the oldProductIdentifier and the optional replacementMode.
```
tonidero, line 29: Could be worth documenting that if both this and `googleProductChangeInfo` are given, the other one will be ignored? And add the same in googleProductChangeInfo. At least until we deprecate the other one.

tonidero: Also, note that this is part of the public API in capacitor and RN, so having both without deprecating one could be confusing... I would suggest delaying releases in those hybrids until we have the chance to deprecate the older one.

tonidero (commented):
I'm just thinking we probably want to deprecate the older APIs in typescript, since that's part of the public API in some hybrids, and right now it's confusing... Not sure if we plan to do that in another PR, but we should do that before updating this in RN/Capacitor.

typescript/src/purchaseParams.ts
tonidero: After talking over Slack, this interface is indeed unused right now, so doesn't matter either way for now :)

tonidero (approved):

typescript/src/purchaseParams.ts
fire-at-will: Added the comment here: 3586454695bbc42e82122b79fd0de32511e05e40

PR #1631, opened 2026-05-04
Revert inline dynamic imports for esm bundle

tonidero:
We originally added a setting to inline dynamic imports since it wasn't an issue to do so and we needed it to make it work with some formats like umd used in purchases-flutter, but we want to avoid loading stripe until it's actually needed. purchases-js changes already do so, but in hybrids, this does not happen because of this setting.

For now, we're reverting this option for the esm bundle export, which is what's used by react-native-purchases, but keeping the umd export as is to avoid issues in purchases-flutter


tonidero:
This is not useful anymore since we went with stripe pure in purchases-js

PR #1658, opened 2026-05-28
Fix crash when app is killed while paywall is shown with MY_APP purchase logic

tonidero:
## Summary

- When Android kills and restores the process while a paywall with `purchaseLogic` is displayed, `PaywallFragment` is recreated from its saved Bundle but `PaywallFragmentNonSerializableArgsStore` (in-memory map) is wiped.
- The fragment relaunches `PaywallActivity` without a `purchaseLogic`, while the Purchases SDK is still configured with `PurchasesAreCompletedBy.MY_APP`.
- The first purchase attempt then crashes with `IllegalStateException: myAppPurchaseLogic must not be null`.

**Fix:** Add a `HAS_NON_SERIALIZABLE_ARGS` boolean to the fragment's Bundle arguments (survives process death). In `onCreate`, if the fragment is being recreated from saved state, that flag is set, and the store is empty, dismiss the paywall immediately rather than entering a broken state.

Normal paywall flows (no `purchaseLogic`/`paywallListener`) are completely unaffected — the flag is only set to `true` when those objects are provided.


ajpallares (approved):
Wow, great find! I think this makes sense!

PR #1662, opened 2026-06-01
Fix paid introductory price not mapped for Web Billing

ajpallares:
### Motivation

Web Billing products with a **paid** introductory price (e.g. $10 the first year, then $20/year) were reported by hybrid SDKs with `introPrice: null`. The API returns the intro price correctly, but the `purchases-js` hybrid offerings mapper dropped it.

### Summary

- `mapIntroPrice` only considered the free `trial` phase and hardcoded `price: 0`; it now falls back from `trial` to `introPrice` and uses the real amount.
- The subscription option's `introPhase` was hardcoded to `null` and the paid intro phase was missing from `pricingPhases`; both are now populated from `option.introPrice`.
- Precedence (`trial` before `introPrice`) matches the native Android mapper, so the two stay consistent.
- Added unit tests for a paid intro with no trial, and for the trial+intro case.

### Notes

Free-trial mapping is unchanged.

### DIFF
--- a/purchases-js-hybrid-mappings/src/mappers/offerings_mapper.ts
+++ b/purchases-js-hybrid-mappings/src/mappers/offerings_mapper.ts
@@ -87,8 +87,9 @@ function mapProduct(product: Product): Record<string, unknown> {
 
 function mapIntroPrice(product: Product): Record<string, unknown> | null {
   const option = product.defaultSubscriptionOption;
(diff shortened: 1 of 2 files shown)


tonidero (approved):
I think this looks good! Might be good to test it on either flutter or RN (which support web), but I think it should work!

PR #1665, opened 2026-06-01
Add presentedOfferingContext parameter support to custom paywall impression events

rickvdl:
Builds on native PRs https://github.com/RevenueCat/purchases-ios/pull/6707 and https://github.com/RevenueCat/purchases-android/pull/3424

Adds the `presentedOfferingContext` (optional) parameter support to the custom paywall impression event API. When present the value will be read from the passed map as `[string: any]` and parses it using existing helpers like in other places.

### DIFF
--- a/android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -1303,10 +1303,18 @@ fun trackAdFailedToLoad(adData: Map<String, Any?>) {
 
 // region Custom Paywall Tracking
 
+@OptIn(InternalRevenueCatAPI::class)
 fun trackCustomPaywallImpression(data: Map<String, Any?>) {
     val paywallId = data["paywallId"] as? String
     val offeringId = data["offeringId"] as? String
-    val params = CustomPaywallImpressionParams(paywallId = paywallId, offeringId = offeringId)
+    val presentedOfferingContext = castWildcardMapToStringToOptionalAnyMap(
+        data["presentedOfferingContext"] as? Map<*, *>,
+    )?.toPresentedOfferingContext()
+    val params = CustomPaywallImpressionParams(
+        paywallId = paywallId,
+        offeringId = offeringId,
+        presentedOfferingContext = presentedOfferingContext,
+    )
     Purchases.sharedInstance.trackCustomPaywallImpression(params)
 }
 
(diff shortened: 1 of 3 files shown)


android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -1303,10 +1303,18 @@ fun trackAdFailedToLoad(adData: Map<String, Any?>) {
...
 fun trackCustomPaywallImpression(data: Map<String, Any?>) {
     val paywallId = data["paywallId"] as? String
     val offeringId = data["offeringId"] as? String
-    val params = CustomPaywallImpressionParams(paywallId = paywallId, offeringId = offeringId)
+    val presentedOfferingContext = castWildcardMapToStringToOptionalAnyMap(
+        data["presentedOfferingContext"] as? Map<*, *>,
+    )?.toPresentedOfferingContext()
+    val params = CustomPaywallImpressionParams(
+        paywallId = paywallId,
+        offeringId = offeringId,
```
tonidero, line 1315: TBH, the presented offering context already has the offering id if any, so in theory we could just pass that in PHC considering it's mostly an internal API... But this would indeed require changes in all hybrids and also the natives to read from there, so let's keep it as this for now :)

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -863,7 +863,14 @@ import StoreKit
...
         let offeringId = data["offeringId"] as? String
-        let params = CustomPaywallImpressionParams(paywallId: paywallId, offeringId: offeringId)
+        let presentedOfferingContext = Self.toPresentedOfferingContext(
+            presentedOfferingContext: data["presentedOfferingContext"] as? [String: Any]
+        )
+        let params = CustomPaywallImpressionParams(
+            paywallId: paywallId,
+            offeringId: offeringId,
+            presentedOfferingContext: presentedOfferingContext
+        )
```
tonidero, line 873: [Unrelated to this line]
Do we need to also support this on web? Maybe it can be added later and we can just mention web is not supported in the docs in Flutter/RN for now, but just wanted to check.

tonidero (approved):
🚢

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
rickvdl: Good question, we currently don’t have the `trackCustomPaywallImpression` API there just yet. Will mention this in the docs or implement support depending on if it’s feasible :)

tonidero (approved):
🚢

PR #1668, opened 2026-06-04
Expose originalJson and signature on StoreTransaction

tonidero:
Some developers have asked for access to the purchase `originalJson` and `signature` properties so they can verify purchases on their own backend without using `PurchasesAreCompletedBy.MY_APP`.

This exposes both on the hybrid `StoreTransaction` model. They are optional: only populated on Android (`signature` is Google Play only), and `null` on iOS.


typescript/src/customerInfo.ts
```diff
@@ -236,6 +236,14 @@ export interface PurchasesStoreTransaction {
      * Purchase token of the transaction. Android only.
      */
     purchaseToken: string | null;
+    /**
+     * The original purchase JSON as a string. Android only, null on iOS and some stores.
+     */
+    originalJson: string | null;
```
tonidero, line 242: Note that I'm exposing this as a string... I doubted whether to expose this as a dictionary, or some sort of json representation... But in the end decided to go simple and avoid the extra parsing without need. Most devs won't need to access this, and for those that do, they can reparse this json if they need to. Lmk what you think!

rickvdl: This sounds like the safer bet, not knowing how the JSON could change overtime and the possibility of discrepancies in serializing and deserializing this data, although I don't have any experience in using/parsing this data so please take it with a grain of salt

rickvdl (approved):

PR #1673, opened 2026-06-09
Add Galaxy Store wrapper dependency for hybrid SDKs

rickvdl:
## Summary
This PR adds the `purchases-hybrid-common-store-galaxy` add-on dependency that allows simplified bringing Galaxy Store support to the hybrid SDKs.

## Motivation
The addition of the wrapper dependency simplifies dependency version management by allowing the hybrid SDKs to just depend on another PHC library (next to `purchases-hybrid-common` and optionally `purchases-hybrid-common-ui`) using the same PHC version number. 

The newly added store support wrapper library (`purchases-hybrid-common-store-galaxy`) will in turn depend on the right version (the same version as the core dependency) of the native store support dependency (`purchases-store-galaxy`). This ensures simple version management and avoids the hybrid SDKs having to depend on a specific version of the native store support dependency as well as the matching version of PHC.

## Dependency visualization
Screenshot 2026-06-09 at 15 20 36

## Testing

- [x] Verify version bump works as expected for release
- [x] Extensive manual testing


android/hybridcommon-store-galaxy/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+    detektPlugins(libs.detekt.formatting)
+}
+
+detekt {
+    buildUponDefaultConfig = true
+    autoCorrect = true
+}
+
+android {
+    namespace = "com.revenuecat.purchases.hybridcommon.galaxy"
```
tonidero, line 18: Supernitpicky... I wonder if we should make the namespace `com.revenuecat.purchases.galaxy.hybridcommon`... but not a strong opinion. I think it's probably fine as is.

android/hybridcommon-store-galaxy/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+        languageVersion = libs.versions.kotlinLanguage.get()
+        apiVersion = libs.versions.kotlinLanguage.get()
+    }
+}
+
+dependencies {
+    api(project(":hybridcommon"))
```
tonidero: Considering there is no code, we might not need this... (this means we will bring the main SDK along when adding the galaxy SDK). But having said that, I don't think it's a big deal to do so either so can leave as is.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -996,6 +998,7 @@ fun configure(
     diagnosticsEnabled: Boolean? = null,
     automaticDeviceIdentifierCollectionEnabled: Boolean? = null,
     preferredLocale: String? = null,
+    galaxyBillingMode: String? = null,
```
tonidero, line 1001: Hmm I'm thinking whether it could be worth to add something more extensible like `galaxyConfigurationOptions` which is a map of string to string. So in case we add more galaxy-only options, we don't need to add it to the configure method which is getting unwieldly 😅. TBH, if we wanted to make it really easy in the hybrids (at the cost of any compilation safety), would be to make a single map to any parameter containing all parameters... But that's a separate discussion 😅.

So happy to leave as this for now and we can think about this refactor separately.

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -1017,9 +1020,27 @@ fun configure(
...
+) {
+    if (store != Store.GALAXY) {
+        return
+    }
+
+    when (galaxyBillingMode) {
+        null, "PRODUCTION" -> galaxyBillingMode(GalaxyBillingMode.PRODUCTION)
+        "TEST" -> galaxyBillingMode(GalaxyBillingMode.TEST)
+        "ALWAYS_FAIL" -> galaxyBillingMode(GalaxyBillingMode.ALWAYS_FAIL)
+        else -> warnLog("Attempted to configure with unknown Galaxy billing mode: $galaxyBillingMode.")
```
tonidero: Should we also default to production in this case? Not sure if it's needed to set it though...

fastlane/Fastfile
```diff
@@ -482,6 +482,15 @@ platform :android do
       properties: gradleProperties.merge({"POM_ARTIFACT_ID" => "purchases-hybrid-common-ui-bc7"}),
       project_dir: 'android'
     )
+
+    # Publish hybridcommon-store-galaxy with -bc7 suffix
```
tonidero: TBH, this is likely not needed... We actually should remove it soon (Google Play won't accept updates with BC7 soon). So feel free to leave it out.

android/hybridcommon-store-galaxy/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+    defaultConfig {
+        minSdk = 23
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        consumerProguardFiles("consumer-rules.pro")
+    }
+
+    buildTypes {
+        release {
+            isMinifyEnabled = false
```
tonidero: Hmm I would say probably not needed? or it could be set to true I think? I see we do have it set to false in the "main" hybridcommon library... But I think it's not a good choice... Probably each hybrid will need to reference the necessary methods, and it's there where we might need to set minification to false (not sure TBH)

android/hybridcommon-store-galaxy/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
+@Suppress("DSL_SCOPE_VIOLATION") // TODO: Remove once KTIJ-19369 is fixed
```
tonidero: I believe this might have been fixed in gradle 8.1? If so, we might not need it.

tonidero (approved):
I think this makes sense!

android/hybridcommon-store-galaxy/build.gradle.kts
rickvdl: Good point, removed it for now

android/hybridcommon-store-galaxy/build.gradle.kts
rickvdl: Good question, but naming is important 😄 any thoughts on this @RevenueCat/coresdk?

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
rickvdl: Good safety net, added!

fastlane/Fastfile
rickvdl: I was in doubt about that one indeed, removed it, thanks for flagging

android/hybridcommon-store-galaxy/build.gradle.kts
rickvdl: Alright, indeed I copied the setup from the main PHC. But let's remove it, I also wouldn't expect any issues for this shell dependency

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
rickvdl: Good point, indeed that could improve extensibility. Will keep it like this for now but should be easy to update once we add more options :)

android/hybridcommon-store-galaxy/build.gradle.kts
tonidero: Let's keep it as is for now :)

PR #1680, opened 2026-06-11
[android] Add Android deploy dry-run CI check

rickvdl:
Deploying to Maven [recently](https://github.com/RevenueCat/purchases-hybrid-common/pull/1679) failed, which was only discovered during release, breaking PHC deploys. This PR adds a `dry-run` option and corresponding CI job that runs on every PR and should catch these issues early.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -264,6 +264,10 @@ jobs:
       name: android/android-machine
       resource-class: large
       tag: default
+    parameters:
+      dry_run:
+        type: boolean
+        default: false
     steps:
       - checkout
       - install-sdkman
@@ -275,7 +279,7 @@ jobs:
       - android/restore-build-cache
       - run:
           name: Deployment
-          command: bundle exec fastlane android deploy
+          command: bundle exec fastlane android deploy dry_run:<< parameters.dry_run >>
 
   deploy-typescript:
     docker:
@@ -472,6 +476,9 @@ workflows:
       - check-pods
       - integration-test-ios
       - test-android
+      - deploy-android:
+          name: android-deploy-dry-run
+          dry_run: true
       - detekt-android
       - typescript-lint
       - typescript-apitests
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -444,7 +444,9 @@ platform :android do
   desc "Upload and close a release"
   lane :deploy do |options|
     version = current_version_number
-    gradleProperties = {
+    dry_run = options[:dry_run] || false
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -275,7 +279,12 @@ jobs:
       - android/restore-build-cache
       - run:
           name: Deployment
-          command: bundle exec fastlane android deploy
+          command: |
+            options=""
+            if [ "<< parameters.dry_run >>" = "true" ]; then
+              options="dry_run:true"
+            fi
+            bundle exec fastlane android deploy ${options}
```
ajpallares: Why not
```suggestion
          command: bundle exec fastlane android deploy dry_run:<< parameters.dry_run >>
```

That seems to work in https://github.com/RevenueCat/purchases-ios/blob/b4c4dba8e8f17ced9fb43339cbac85b2f6f42ece/.circleci/default_config.yml#L1858

Then, in the fastfile, [it's evaluated](https://github.com/RevenueCat/purchases-ios/blob/b4c4dba8e8f17ced9fb43339cbac85b2f6f42ece/fastlane/Fastfile#L1991) with
```
dry_run = options[:dry_run] || false
```

ajpallares (approved):
Makes sense! Just a suggestion to simplify the commands a bit

fastlane/Fastfile
```diff
@@ -444,19 +444,21 @@ platform :android do
   desc "Upload and close a release"
   lane :deploy do |options|
     version = current_version_number
-    gradleProperties = {
+    dry_run = options[:dry_run].to_s == "true"
+    dry_run_flag = dry_run ? " --dry-run" : ""
```
tonidero, line 448: Hmm I think with dry run it won't execute most tasks... Though it's likely it will catch this issue since it happens when evaluating gradle... So it's an improvement :) I do wonder if we would like to "simulate" a dry run using `publishToMavenLocal` instead of `publish`... but that will probably take longer, and I guess this catches what we need for now.

tonidero (approved):
Looking good!

.circleci/config.yml
rickvdl: Good one, updated!

fastlane/Fastfile
rickvdl: Right, I see your point. I think this is probably a lightweight check that is an improvement, but we could maybe also do the publish to maven local on the release PR just before triggering the release? Since it would probably take bit longer. I'll note this as a follow up

PR #1682, opened 2026-06-11
Allow custom paywall impressions without event data

rickvdl:
Previously the PHC APIs did not support `nil` values for the params, which required the hybrids to send an empty object instead, which in turn caused the native side to treat it as valid data and not use the fallback (aka using the current offering).

---

> [!NOTE]
> **Low Risk**
> Narrow API signature change on analytics bridging with added tests; behavior fix for hybrid callers that omit event data.
> 
> **Overview**
> **`trackCustomPaywallImpression`** on Purchases Hybrid Common (Android Kotlin and iOS `CommonFunctionality`) now accepts **optional / nil event data** instead of requiring a map.
> 
> When hybrids pass **no data** (`null` / `nil`), the bridge normalizes to an empty payload and forwards **`CustomPaywallImpressionParams` with null `paywallId`, `offeringId`, and `presentedOfferingContext`**, so the native SDK can apply its **current-offering fallback** instead of treating a forced `{}` as explicit (empty) event fields.
> 
> API compile tests (Java/Kotlin/ObjC) and Android unit tests cover null, partial (`paywallId` only), and full maps including `presentedOfferingContext`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71ef12fb3509269576833267703a235e88e24769. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
+++ b/android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
@@ -1328,11 +1328,12 @@ fun trackAdFailedToLoad(adData: Map<String, Any?>) {
 // region Custom Paywall Tracking
 
 @OptIn(InternalRevenueCatAPI::class)
-fun trackCustomPaywallImpression(data: Map<String, Any?>) {
-    val paywallId = data["paywallId"] as? String
-    val offeringId = data["offeringId"] as? String
+fun trackCustomPaywallImpression(data: Map<String, Any?>?) {
+    val eventData = data.orEmpty()
+    val paywallId = eventData["paywallId"] as? String
+    val offeringId = eventData["offeringId"] as? String
     val presentedOfferingContext = castWildcardMapToStringToOptionalAnyMap(
-        data["presentedOfferingContext"] as? Map<*, *>,
+        eventData["presentedOfferingContext"] as? Map<*, *>,
     )?.toPresentedOfferingContext()
     val params = CustomPaywallImpressionParams(
         paywallId = paywallId,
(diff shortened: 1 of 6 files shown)


android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
```diff
@@ -1328,11 +1328,12 @@ fun trackAdFailedToLoad(adData: Map<String, Any?>) {
 // region Custom Paywall Tracking
 
 @OptIn(InternalRevenueCatAPI::class)
-fun trackCustomPaywallImpression(data: Map<String, Any?>) {
-    val paywallId = data["paywallId"] as? String
-    val offeringId = data["offeringId"] as? String
+fun trackCustomPaywallImpression(data: Map<String, Any?>?) {
+    val eventData = data.orEmpty()
```
tonidero, line 1332: Hmm so I'm unsure this matches what was discussed in the PR description. This indeed allows to pass in a null value, but since we default to empty, this means that the value passed to the natives remain the same?

tonidero (commented):
Just a question 🙏

android/hybridcommon/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt
rickvdl: You're totally right and thanks for challenging this. I was going in circles while fixing this, and indeed it can be solved on the hybrid side it seems.. Closing this PR for now.

PR #1698, opened 2026-06-22
Expose `autoResumeDate`, `displayName`, `managementURL`, `productPlanIdentifier` on `SubscriptionInfo`

AlvaroBrey:
Exposes several new fields from `SubscriptionInfo`

Closes SDK-4382

---

> [!NOTE]
> **Low Risk**
> Additive, read-only CustomerInfo fields with cross-platform mapping and test coverage; no purchase or auth logic changes.
> 
> **Overview**
> Adds four **read-only** fields on per-product **`PurchasesSubscriptionInfo`** in `customerInfo.subscriptionsByProductIdentifier`: `autoResumeDate`, `displayName`, `managementURL`, and `productPlanIdentifier`, aligned with native RevenueCat `SubscriptionInfo`.
> 
> **Android** and **iOS** hybrid mappers now include these keys in the bridge dictionary (dates as ISO8601, management URL as string). **TypeScript** types and API report are updated with JSDoc describing Play pause resume, dashboard display name, per-subscription management link, and base/plan identifiers.
> 
> **Web** (`purchases-js-hybrid-mappings`) forwards `displayName`, `managementURL`, and `productPlanIdentifier` from purchases-js and always emits `autoResumeDate: null` because paused Play subscriptions are not modeled on web.
> 
> **iOS** `Package.resolved` bumps `purchases-ios-spm` **5.57.1 → 5.80.0**. Tests and StoreKit integration snapshots were updated for the new JSON shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f082e447425998477d863179d7198e8c588ddf8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1729, opened 2026-07-08
Regenerate Package.resolved via swift package resolve on iOS bumps

AlvaroBrey:
### Motivation

The automatic iOS bump updated `Package.resolved` by text-replacing the previous version and commit hash. If the file ever got out of sync (a manual bump in #1461 missed it), the pattern never matched again and every later bump silently skipped the file. It stayed pinned at 5.57.1 for 5 months until fixed by hand in #1698, where this automation fix was suggested: https://github.com/RevenueCat/purchases-hybrid-common/pull/1698#discussion_r3491490922

### Description

After bumping `Package.swift`, run `swift package resolve` to regenerate `Package.resolved` instead of pattern-matching its current content. Resolution rewrites the file from the `exact:` pin, so any drift self-heals on the next bump, and a resolution failure fails the lane loudly. 

Also removes the now-unused `SPM_RESOLVED_VERSION_PATTERNS` constant and `get_commit_hash_for_tag` helper.

Verified locally

---

> [!NOTE]
> **Low Risk**
> Changes only fastlane release automation; no runtime app code, with clearer failure behavior if SPM resolution breaks.
> 
> **Overview**
> **iOS dependency bump automation** no longer text-replaces `Package.resolved` (version strings or commit hashes). After updating pins in `Package.swift` and related files, the lane runs **`swift package resolve`** at the repo root so the lockfile is rewritten from the `exact:` pin—out-of-sync locks self-heal on the next bump, and resolution errors fail the lane instead of silently skipping updates.
> 
> Removes unused **`SPM_RESOLVED_VERSION_PATTERNS`**, drops `Package.resolved` from the iOS bump file list, and deletes **`get_commit_hash_for_tag`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea6a8c15df6ee23fb337273cfb44955e60b4e92c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -26,14 +26,12 @@ files_with_version_number_without_prerelease_modifiers = {
 }
 POD_VERSION_PATTERNS = ['\'RevenueCat\', \'{x}\'', '\'RevenueCatUI\', \'{x}\'']
 SPM_VERSION_PATTERNS = ['.package(url: "https://github.com/RevenueCat/purchases-ios-spm", exact: "{x}"),']
-SPM_RESOLVED_VERSION_PATTERNS = ['"version" : "{x}"']
 files_to_update_ios_dependency = {
     'PurchasesHybridCommon.podspec' => POD_VERSION_PATTERNS,
     'PurchasesHybridCommonUI.podspec' => POD_VERSION_PATTERNS,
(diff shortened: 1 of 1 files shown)


tonidero (approved):
Makes sense!

PR #1730, opened 2026-07-08
Map autoResumeDate from purchases-js SubscriptionInfo in JS hybrid mappings

AlvaroBrey:
### Motivation

Follow-up to https://github.com/RevenueCat/purchases-hybrid-common/pull/1698#discussion_r3491495706: the JS mappings hardcoded `autoResumeDate: null` because purchases-js did not expose it yet.

### Description

purchases-js exposes `autoResumeDate` on `SubscriptionInfo` since 1.47.0 (https://github.com/RevenueCat/purchases-js/pull/957), which is already our pinned version. Map the real value instead of the hardcoded null.

---

> [!NOTE]
> **Low Risk**
> Small mapper and test change with no auth or payment logic; behavior only improves fidelity when the JS SDK provides a resume date.
> 
> **Overview**
> **Subscription mapping** now forwards **`autoResumeDate`** from purchases-js `SubscriptionInfo` into the hybrid `CustomerInfo` payload, using the same ISO-string-or-`null` pattern as other date fields, instead of always emitting **`null`**.
> 
> The mapper test was updated to set **`autoResumeDate`** on mock subscription data and assert the mapped value is the ISO string.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea90cbadaf5682cf736b15e0b4944d472c780b1d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
+++ b/purchases-js-hybrid-mappings/src/mappers/customer_info_mapper.ts
@@ -175,8 +175,9 @@ function mapSubscriptionInfos(
         storeTransactionId: subscriptionInfo.storeTransactionId,
         isActive: subscriptionInfo.isActive,
         willRenew: subscriptionInfo.willRenew,
-        // Not modeled by purchases-js; web has no Play-paused subscriptions.
-        autoResumeDate: null,
+        autoResumeDate: subscriptionInfo.autoResumeDate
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Thank you for following through! 🙇

PR #1770, opened 2026-07-24
Expose workflows dangerous settings to Objective-C

vegaro:
Flutter and Unity need to create `DangerousSettings(autoSyncPurchases:useWorkflows:)` from Objective-C. That initializer is Swift SPI, so this exposes a small factory through `RCCommonFunctionality` instead of requiring each hybrid SDK to add its own Swift bridge. Both flags are forwarded so Unity keeps its existing auto-sync behavior.

---

> [!NOTE]
> **Low Risk**
> Additive hybrid-common API with tests; behavior only changes when callers pass the new settings into existing configure paths.
> 
> **Overview**
> Adds an Objective-C–callable factory on **`RCCommonFunctionality`** so Flutter and Unity can build **`DangerousSettings`** without each hybrid shipping its own Swift bridge, since the RevenueCat initializer is Swift-only/SPI.
> 
> **`createDangerousSettingsWithAutoSyncPurchases:useWorkflows:`** forwards both **`autoSyncPurchases`** and **`useWorkflows`** into **`DangerousSettings`**, so existing auto-sync behavior can stay the same while workflows can be toggled at configure time.
> 
> Coverage includes the ObjC API compile test and unit tests that assert both boolean combinations on the returned settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56797dfc00b6981e462f2d3d7c0e341d5c14baae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -39,8 +39,9 @@ import StoreKit
     @objc public static var isAnonymous: Bool { Self.sharedInstance.isAnonymous }
     @objc public static var hybridCommonVersion: String { Constants.hybridCommonVersion }
 
-    @objc public static func createDangerousSettings(useWorkflows: Bool) -> DangerousSettings {
-        DangerousSettings(useWorkflows: useWorkflows)
+    @objc public static func createDangerousSettings(autoSyncPurchases: Bool,
+                                                     useWorkflows: Bool) -> DangerousSettings {
+        DangerousSettings(autoSyncPurchases: autoSyncPurchases, useWorkflows: useWorkflows)
     }
 
     @objc public static var proxyURLString: String? {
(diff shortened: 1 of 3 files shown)


ajpallares (approved):

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -39,6 +39,11 @@ import StoreKit
     @objc public static var isAnonymous: Bool { Self.sharedInstance.isAnonymous }
     @objc public static var hybridCommonVersion: String { Constants.hybridCommonVersion }
 
+    @objc public static func createDangerousSettings(autoSyncPurchases: Bool,
```
tonidero, line 42: Hmm I would argue this might be better in a different place... but not a blocker :)

vegaro: that's a good point actually. Let me think about it and open another PR

PR #1825, opened 2026-08-18
mark @amazon-devices/keplerscript-appstore-iap-lib as external

fire-at-will:
We are currently working on adding support to the Amazon Store on the Vega OS to `purchases-js`. To avoid including the `@amazon-devices/keplerscript-appstore-iap-lib` for web consumers, the JS SDK pulls `@amazon-devices/keplerscript-appstore-iap-lib` in dynamically at runtime, like this:

```typescript
const amazonSdkModule = "@amazon-devices/keplerscript-appstore-iap-lib";
await import(amazonSdkModule);
```

This prevents rollup from including `@amazon-devices/keplerscript-appstore-iap-lib` in the PHC bundles produced by purchases-js-hybrid-mappings because Rollup cannot statically resolve and inline the import. However, this behavior is inherited, not explicitly stated. The `external` entry is serves as an explicit safeguard: if the import later becomes statically analyzable (for example import("@amazon-devices/...")), Rollup will leave it external instead of bundling it.

To make this behavior explicit, this PR marks `@amazon-devices/keplerscript-appstore-iap-lib` as [external](https://rollupjs.org/configuration-options/#external) in the rollup configuration to ensure that it is not bundled in the modules we create.

The external entry prevents Rollup from resolving and bundling the Amazon SDK, but Terser can subsequently inline the runtime module-name variable and emit a statically resolvable import("@amazon-devices/..."). The PR also disables Terser’s `reduce_vars` optimization so the import remains runtime-only in the `purchases-js-hybrid-mappings` bundle, preventing web consumers from needing the optional Amazon dependency.

---

> [!NOTE]
> **Low Risk**
> Build-only Rollup/Terser config in hybrid-mappings; no runtime auth or payment logic changes.
> 
> **Overview**
> Updates **`purchases-js-hybrid-mappings`** Rollup build so the optional Amazon Vega IAP package stays out of published bundles and web apps are not forced to resolve it.
> 
> **`@amazon-devices/keplerscript-appstore-iap-lib`** is listed in **`external`**, so Rollup will not bundle it even if the source later uses a statically analyzable dynamic import.
> 
> Terser **`compress.reduce_vars`** is turned off so minification does not collapse the runtime module-name variable into a static `import("@amazon-devices/...")`, which would make downstream web bundlers treat the SDK as a required dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba595f4209a4bb3c87abf70cb3c5354b7ce0f75e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fire-at-will:
@claude please review

tonidero (approved):
Thank you thank you!! 🙌 🚢

james-od (approved):

PR #1836, opened 2026-08-23
introduce an isPurchasesError function in the code

### LINKED ISSUE #291: Vend a typeguard function for checking if an Error is a PurchasesError
Hey. When using TypeScript, the type of the caught error is `unknown` (`any` in non-strict mode of older versions of TypeScript). It would be very useful to have a typeguard function vended that could narrow the type. Would you be willing to accept a PR that implements this?

```typescript
export const isPurchasesError = (err: unknown): err is PurchasesError => {
  // Implementation here
}
```

The usage would look like: 

```typescript
try {
  await someCodeThatTriggersAPurchase()
} catch (err) {
  if (isPurchasesError(err)) {
    // Handle PurchasesError here. 
    // The type will be narrowed by the compiler and IDE completion will work nicely.
  } else {
    // Handle other errors differently
  }
}
```

This is something we can of course just write on our own end, but it seems like it'd be useful to have as part of the core library here.

TheRogue76:

Implements an isPurchasesError function based on the current non-depracated fields in the interface and checks that it's codes match the ones expected.

I had to introduce es2017's object as a lib in the tsconfig, let me know if that is undesired, since the other solutions for creating the set from the enum values were kind of ugly, but i can bring them back

Fixes https://github.com/RevenueCat/react-native-purchases/issues/291

No, but i also noticed purchases-common.test.ts is complaining about missing fields on `mockMonthlyProduct`, will write another PR to fix that

---

> [!NOTE]
> **Low Risk**
> Small public API addition for error identification; no auth, data, or purchase-flow changes. TypeScript lib target is slightly expanded for Object.values.
> 
> **Overview**
> Adds a public `isPurchasesError` type guard so consumers can tell whether an unknown value is a SDK `PurchasesError`.
> 
> The helper checks non-deprecated fields (`code`, `message`, `userInfo.readableErrorCode`, `underlyingErrorMessage`) and that `code` is a known `PURCHASES_ERROR_CODE`. It is exported from the existing errors module.
> 
> `tsconfig` now includes `es2017.object` so `Object.values` can build the known-code set. No tests are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89b76247544c21f01eaa61b859fc5895cbe66234. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/typescript/api-report/purchases-typescript-internal.api.md
+++ b/typescript/api-report/purchases-typescript-internal.api.md
@@ -87,6 +87,9 @@ export interface IntroEligibility {
 }
 
 // @public
+export function isPurchasesError(error: unknown): error is PurchasesError;
+
+// @public
 export enum LOG_LEVEL {
     // (undocumented)
     DEBUG = "DEBUG",
--- a/typescript/src/errors.ts
+++ b/typescript/src/errors.ts
@@ -55,3 +55,43 @@ export class UnsupportedPlatformError extends Error {
         Object.setPrototypeOf(this, UnsupportedPlatformError.prototype);
     }
 }
+
+/**
+ * Returns whether the given value is a {@link PurchasesError} thrown by the SDK.
+ *
+ * A value qualifies when it's schema matches {@link PurchasesError} and a `code` that is one of the
+ * {@link PURCHASES_ERROR_CODE} values.
+ *
+ * @public
+ */
+export function isPurchasesError(error: unknown): error is PurchasesError {
+    if (typeof error !== "object" || error === null) {
+        return false;
+    }
+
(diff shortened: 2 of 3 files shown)


AlvaroBrey:
Hey @TheRogue76, didn't mean to leave you hanging, but we're currently doing a rework of how errors are propagated to the hybrids, so I'll revisit this once that's done and let you know if it still makes sense

TheRogue76:
Hi @AlvaroBrey
Thanks for the heads up. Sure thing! will take up some other task then for the moment and try to see how else i can contribute

PR #1838, opened 2026-08-24
feat: Normalize hybrid bridge errors to the PurchasesError interface

AlvaroBrey:
- Adds `normalizePurchasesError`, so errors reaching JS consumers match the `PurchasesError` interface this package declares. Before this change they did not meet the interface, on either React Native or Capacitor:
  - React Native: the bridge forwards only `NSError.userInfo` to JS, so on iOS the `ErrorContainer.info` payload never arrived at all, and on Android it arrives nested under `userInfo`. `readableErrorCode`, `underlyingErrorMessage` and `userCancelled` are declared at the top level but were never there on either platform.
  - Capacitor: Android puts the payload under `data`; iOS passed nothing, so the same three fields were `undefined` and only `code` and `message` existed. Reported as RevenueCat/purchases-capacitor#302, open since v7.
- iOS: `ErrorContainer.error` is now the original `NSError` rather than a re-wrapped copy with `readableErrorCode` injected into its `userInfo`. Only `react-native-purchases` ever read that key, and it now does that merge at its own reject site in RevenueCat/react-native-purchases#1919. `flutter`, `cordova` and `unity` read `info` and are unaffected.
- Builds on RevenueCat/purchases-hybrid-common#1635, which makes the web mapper emit a string `code`. **Must merge together with RevenueCat/purchases-capacitor#579 and RevenueCat/react-native-purchases#1919**
- `userCancelled` is derived from `code` for every normalized error (SDK-4448). React Native's eight `.catch` blocks that re-derived it are removed in #1919, so it is present on every SDK error: `true` for a cancelled purchase on every platform (Capacitor and RN web never reported it), `false` elsewhere, where it was absent on Capacitor and on RN's non-purchase methods.

Part of react-native-purchases#1756 and purchases-capacitor#302. Tracked in SDK-4450 and SDK-4448.

### Checklist

  - [x] If applicable, unit tests.

<details><summary>Agent description</summary>

### Motivation

This package builds a consistent error payload in `ErrorContainer.info`, but it does not own how that payload reaches JS. Each host framework nests it under a different key: React Native under `userInfo`, Capacitor under `data`. Both iOS paths dropped it entirely. Meanwhile the `PurchasesError` interface declares every field flat at the top level, a shape no bridge produces.

The interface has been wrong since it was copied out of `react-native-purchases` in #455. It was already half wrong then: react-native-purchases#268 had found in 2021 that `readableErrorCode` actually lives in `userInfo`, added `userInfo: ErrorInfo` and deprecated the flat field, but left `underlyingErrorMessage` and `userCancelled` with the identical defect.

Reported downstream as RevenueCat/purchases-capacitor#302, open since v7.

### Description

- `normalizePurchasesError` reads the payload from `userInfo` or `data` and lifts the declared fields to the top level. It mutates the error in place and returns it: copying into a new object would discard the prototype and stack, breaking `instanceof Error` for consumers and `instanceof CapacitorException` on Capacitor.
- It only acts on errors whose `code` is numeric. Plugin level rejections use names such as `UNIMPLEMENTED` or `PAYWALL_ERROR` and are left untouched.
- `userInfo` is merged rather than replaced, so Capacitor gains the full payload it was missing while React Native keeps the richer one its bridge already provides. Both platforms now expose the same `userInfo`.
- Only the normalizer lives here. Each SDK wraps its own native module and calls it on every rejection; the interception is bridge specific (React Native must never copy a TurboModule promise, Capacitor must keep `addListener().remove`), so it belongs with the bridge.
- iOS: `ErrorContainer` no longer rebuilds the `NSError`. The long standing `readableErrorCode` special case goes with it.

**Not visible in the diff:** the JS normalizer is exercised against hand written fixtures of each bridge envelope, taken from reading the bridges. Checked once by hand on 2026-09-10 against an iPhone 17 simulator, a Pixel emulator (React Native) and a Pixel emulator (Capacitor) with a bogus API key: every declared field was present with the declared type on all three, and the captured React Native iOS `userInfo` was a superset of the fixture. Nothing repeats that check automatically, which is the same gap that let the original mismatch persist. `react-native-purchases` has no iOS test target, so the merge it now performs is compile checked there; `ErrorContainerPayloadTests` here pins that merge's input across the three ways this package builds an `NSError`.

**Regression gate:** `errorNormalizer.test.ts` "satisfies PurchasesError" across the per bridge fixtures, which is the contract the consumers' own wrappers rely on. The Hermes constraint that shaped the React Native wrapper (enumerating a TurboModule promise's own keys throws `Cannot read property 'length' of null`) is documented and tested in react-native-purchases#1919, where that wrapper now lives.

**Limitations:**

- Web errors still get an empty `readableErrorCode`, because `purchases-js-hybrid-mappings`' `mapPurchasesError` never emits one. That mapper is the right place to fix it.
- Web `userInfo` carries `statusCode` and `backendErrorCode` lifted from `info`, but not `userCancelled`, which the mapper puts at the top level. Native `info` includes it for purchase errors.
- `userCancelled` is `boolean | null` in the interface but never `null` at runtime any more. Narrowing the type waits for a major.
- Only `react-native-purchases` and `purchases-capacitor` consume this package. `flutter`, `cordova` and `unity` forward `info` faithfully and need no changes.

**Rejected:**

- Matching `code` against `PURCHASES_ERROR_CODE` membership. That enum omits codes the native SDKs emit (36 to 41), and Android and iOS disagree on 28 and 36, so it would reject genuine errors, including the web purchase redemption ones.
- Shipping the module wrapper from here as `withNormalizedErrors`. It was one place to test, but it encoded both bridges' quirks in shared code and made this package a runtime hook rather than a contract; an earlier version of this PR did that.
- Keeping the iOS merge here. It exists only for React Native's bridge, and leaving it gave the other four hybrids a `userInfo` they never asked for while making Capacitor send the payload twice.
- Flattening `info` onto the error in native code, matching `cordova` and `unity`. No bridge lets native add top level keys.
- Guarding the code coercion so web kept its numeric `code` until the next major. It left react-native-purchases#1756 unfixed, and #1635 already stringifies at the source, so the guard would have been dead code.
- Making the phantom interface fields optional. Honest, but it breaks compilation for anyone assigning them to a non-optional `string`. Growing the runtime gets the same honesty additively.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes cross-platform error shape and iOS ErrorContainer forwarding; consumers must call the normalizer on rejections (coordinated PRs), but behavior is covered by new unit tests.
> 
> **Overview**
> Adds **`normalizePurchasesError`** to the TypeScript package so hybrid SDKs can lift bridge-nested payloads (`userInfo`, `data`, `info`) onto a flat **`PurchasesError`** shape in place—preserving prototypes/stacks and only touching numeric SDK codes. Jest fixtures cover React Native, Capacitor, and web envelopes; CI gains a **`typescript-tests`** job and lint runs via **`yarn lint`**.
> 
> On iOS, **`ErrorContainer`** no longer re-wraps the **`NSError`** to inject **`readableErrorCode`** into **`userInfo`**; it forwards the original error and still fills **`info`** (including a **`rc_code_name`** fallback when **`readable_error_code`** is missing). New Swift tests lock the hybrid **`info`** contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8259d11e74614accad802895cc3a60bb5deaa26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
@cursor review

AlvaroBrey:
Reviewer notes: the 2k lines are mostly `yarn.lock` (jest)

tonidero (approved):
Thank you for dealing with this mess! 🙇

AlvaroBrey:
Holding until after the next release train, to have time to merge this and https://github.com/RevenueCat/react-native-purchases/pull/1919 in the same release

PR #1843, opened 2026-08-25
feat: add onRestoreInitiated paywall listener bridge

ninjz:
## Summary

Native paywalls already gate restore before it starts:

- iOS `PaywallViewControllerDelegate.paywallViewController(_:didInitiateRestoreWith:)` (`paywallViewController:didInitiateRestoreWithResume:`, purchases-ios ≥ 5.67, see RevenueCat/purchases-ios#6392)
- Android `PaywallListener.onRestoreInitiated(resume: Resumable)` (purchases-android ≥ 9.28)

`PaywallListenerWrapper` / `PaywallProxy` never forwarded that hook. Hybrids that must `logIn` / identify the user **before** StoreKit restore (e.g. logout left RC on the previous App User ID) have no way to pause restore. Purchase already has this via `onPurchasePackageInitiated` + `resume(requestId, shouldProceed)`.

This PR mirrors that shape for restore **on `PaywallProxy` itself** (not a subclass). `PaywallProxy` is a Swift `public` class, so it is `objc_subclassing_restricted` — RN / ObjC consumers cannot subclass it to add this delegate. Putting the stash + `resumeRestoreInitiated` on the proxy is the only shape that compiles.

**iOS**
- `PaywallProxy` implements `paywallViewController(_:didInitiateRestoreWith:)`, stashes `resume` by UUID
- Wrapper: `paywallViewController:didInitiateRestoreWithRequestId:`
- `PaywallProxy.resumeRestoreInitiated(requestId:shouldProceed:)`
- If the wrapper method is unimplemented, auto-`resume(true)` (same as purchase)

**Android**
- `PaywallListenerWrapper` overrides `onRestoreInitiated(Resumable)`, converts to `onRestoreInitiated(requestId: String)` using the existing `pendingResumeCallbacks` map
- `PaywallListenerWrapper.resumeRestoreInitiated`
- Default open method auto-resumes if the hybrid SDK does not override

Follow-up RN PR: [react-native-purchases#1923](https://github.com/RevenueCat/react-native-purchases/pull/1923) (depends on a published hybrid-common version that includes this).

### DIFF
--- a/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallListenerWrapper.kt
+++ b/android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallListenerWrapper.kt
@@ -22,6 +22,11 @@ abstract class PaywallListenerWrapper : PaywallListener {
         fun resumePurchasePackageInitiated(requestId: String, shouldProceed: Boolean) {
             pendingResumeCallbacks.remove(requestId)?.invoke(shouldProceed)
         }
+
+        @JvmStatic
+        fun resumeRestoreInitiated(requestId: String, shouldProceed: Boolean) {
+            pendingResumeCallbacks.remove(requestId)?.invoke(shouldProceed)
+        }
     }
 
     override fun onPurchasePackageInitiated(rcPackage: Package, resume: Resumable) {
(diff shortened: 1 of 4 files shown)


android/hybridcommon-ui/src/main/java/com/revenuecat/purchases/hybridcommon/ui/PaywallListenerWrapper.kt
```diff
@@ -22,6 +22,11 @@ abstract class PaywallListenerWrapper : PaywallListener {
         fun resumePurchasePackageInitiated(requestId: String, shouldProceed: Boolean) {
             pendingResumeCallbacks.remove(requestId)?.invoke(shouldProceed)
         }
+
+        @JvmStatic
+        fun resumeRestoreInitiated(requestId: String, shouldProceed: Boolean) {
```
tonidero, line 27: This system of having a `shouldProceed` parameter is used for purchasePackageInitiated as a way for developers to "intercept" the purchase attempt, and possibly block it or react to it in some way. 

I think for initiating restore we don't need that customization? Or do you have some use case for it? 
- If you do need it, I think we should first add such functionality to the purchases-ios and purchases-android SDKs.
- If you don't need it, maybe we can change this, to instead calling the callback directly, as we do for the `onPurchaseStarted` callback and the abstract func below?

This comment applies to the Swift implementation as well

tonidero (commented):
Thanks so much for the contribution! It's indeed a missing gap in our API. Just a comment about a possible simplification to the system, so we can follow the existing on `onPurchaseStarted` callback or if we do need the extra "interception" functionality.

PR #1844, opened 2026-08-26
build: update Kotlin to 2.2.20

AlvaroBrey:
- Bumps Kotlin from 2.0.21 to 2.2.20. One line in `android/gradle/libs.versions.toml`.
- **Required by the purchases-android v11 major** (RevenueCat/purchases-android#3964). v11 publishes `kotlin-stdlib` with metadata 2.2.0, and a Kotlin compiler reads metadata at most one minor ahead, so our 2.0.21 compiler cannot consume it.
- **Raises the Kotlin floor for anything that compiles Kotlin against PHC to 2.1**, because we publish `kotlin-stdlib` at compile scope too. This is not extra breakage: `api(libs.purchases)` re-exposes purchases at compile scope, so the v11 bump forces the same floor regardless of what PHC itself compiles with.
- `kotlinLanguage` deliberately stays at `1.8`. Kotlin 2.2.20 still accepts it, so this is a compiler bump and not a K1 to K2 migration.
- Deliberately does **not** bump `purchases` in the same PR. PHC builds fine on 2.2.20 while still on 10.18.1, so this can land ahead of the v11 release and de-risk it.
- **Release ordering matters**: see below before cutting a PHC release with this in it.

Companion PRs: https://github.com/RevenueCat/purchases-flutter/pull/1876 https://github.com/RevenueCat/react-native-purchases/pull/1925

### Checklist
- [x] A description about what and why you are contributing
- [x] The issue number(s) or PR number(s) in the description
- [ ] If applicable, unit tests

<details><summary>Agent description</summary>

### Motivation

purchases-android v11 moves to Kotlin 2.2.21 as part of its AGP 9 upgrade. The Kotlin Gradle Plugin
publishes `kotlin-stdlib` at `compile` scope, so Gradle resolves consumers onto it, and a compiler
can only read metadata from at most one minor version ahead of itself. v11's stdlib carries metadata
2.2.0, which a 2.0.21 compiler rejects with:

```
Module was compiled with an incompatible version of Kotlin.
The binary version of its metadata is 2.2.0, expected version is 2.0.0.
```

So PHC has to move to at least 2.1 before it can take v11. 2.2.20 is chosen to match what
purchases-capacitor already pins, so the hybrids converge on one version.

### Why this is separable from the purchases bump

PHC compiles cleanly on Kotlin 2.2.20 with `purchases` still at 10.18.1, verified locally. That makes
this landable before the v11 release rather than gated behind it, which is the whole reason it is its
own PR.

### Downstream impact

Consumers that compile Kotlin against PHC need a compiler of 2.1 or newer after this:

| Consumer | Compiler | Status |
|---|---|---|
| purchases-flutter `purchases_ui_flutter` | KGP 1.9.20 | needs RevenueCat/purchases-flutter#1875 |
| react-native apps on RN 0.78 / 0.79 | KGP 2.0.21 | needs RevenueCat/react-native-purchases#1925 |
| react-native apps on RN 0.81+ | KGP 2.1.20 | fine, reads 2.2.0 |
| purchases-capacitor | 2.2.20 | fine |
| cordova, unity | Java only, no Kotlin compilation | fine |

**Merging this is safe. Releasing it is what needs sequencing.** Once a PHC version ships with this,
the automated PHC bump PRs open against purchases-flutter and react-native-purchases and will fail
their CI until their Kotlin prep lands. Land #1875 and #1925 first, or hold the PHC release for the
coordinated major.

### Testing

`cd android && ./gradlew assembleRelease` passes on 2.2.20 with `purchases` unchanged at 10.18.1.
Only pre-existing deprecation warnings, plus the expected `Language version 1.8 is deprecated` note.
The generated POM confirms the new contract: `kotlin-stdlib:2.2.20` at `compile` scope.

### Related

- RevenueCat/purchases-android#3964 introduces the floor this responds to
- RevenueCat/purchases-flutter#1875 and RevenueCat/react-native-purchases#1925 are the downstream
  Kotlin bumps that must land before a PHC release carrying this

</details>


android/gradle/libs.versions.toml
```diff
@@ -4,11 +4,11 @@ assertj-core = "3.22.0"
 coroutines = "1.6.4"
 fragmentKtx = "1.6.2"
 junit-jupiter = "5.8.2"
-kotlin = "2.0.21"
+kotlin = "2.2.21"
 kotlinLanguage = "1.8"
 junit = "4.13.2"
 mockk = "1.12.8"
-purchases = "10.22.1"
+purchases = "11.0.0-SNAPSHOT"
```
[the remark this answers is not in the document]
AlvaroBrey: intentional but deserves a comment I guess

android/settings.gradle.kts
```diff
@@ -10,6 +10,9 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
+        maven(url = uri("https://central.sonatype.com/repository/maven-snapshots/")) {
```
tonidero, line 14: Could be worth leaving some TODO/WIP comment to remove these later... But not a blocker!

tonidero (approved):
Nice!

android/settings.gradle.kts
AlvaroBrey: I put it on the [version](https://github.com/RevenueCat/purchases-hybrid-common/pull/1844/changes#diff-baf4a872b28c79c64ddfd3741761d13df7f2942ae5bbf0a0f4a2ff2ab4775d4eL11) but yeah I can put it here too!

AlvaroBrey: Done!

PR #1851, opened 2026-08-28
fix: wait for pods to reach the CocoaPods CDN before triggering hybrid bumps

vegaro:
## What and why

A pod pushed to trunk isn't immediately resolvable by `pod install` — `18.33.1` was on trunk at 14:35 UTC on 2026-08-27 and still wasn't on the CDN index two hours later. The hybrid bump PRs open right after the release, so their iOS jobs fail with `CocoaPods could not find compatible versions for pod "PurchasesHybridCommon"` until someone reruns them by hand.

This waits once in `trigger-dependent-updates`, before kicking off any of the bumps, instead of each hybrid waiting on its own. Per the thread with @vegaro / @tonidero / @facundoolano: waiting here means one already-provisioned `cimg/ruby` docker job blocks, rather than mac executors idling across five repos. The release PR stays open while it waits, so the tagged commit's jobs stay visible.

## Timeout behaviour

Three limits interact here, so spelling them out:

- **The wait is capped at 3 hours.** CircleCI *cancels* a job that exceeds its max runtime, which would skip the bump steps entirely, so the wait has to stay clear of that cap. We're on the Scale plan (5h), confirmed by a `purchases-ios` job that ran 5.09h, so 3 hours leaves a couple of hours of headroom. Set explicitly rather than inheriting the action's default so a change there can't silently push us over.
- **On timeout it still triggers the bumps** (`soft_fail`). Failing would leave every hybrid without a PR at all, which is worse than today; opening them means each can be rerun individually.
- **`no_output_timeout: 190m`**, above the wait. The lane logs on every poll so the 10 minute default wouldn't fire in normal operation, but a stalled CDN request would otherwise kill the step at 10 minutes. If the step does fail, the bump steps still run — they're all `when: always`.

Net: a normal propagation is absorbed silently, and every failure mode still ends with five bump PRs open.

## Contributing checklist

- [x] The issue number(s) or PR number(s) in the description if you are contributing in response to those. Uses `wait_for_pods_in_cocoapods_cdn` from RevenueCat/fastlane-plugin-revenuecat_internal#154; replaces the per-repo approach in RevenueCat/react-native-purchases#1930.
- [ ] If applicable, unit tests. The action's logic is unit tested in the plugin; this is the wiring.

## Verification

- [x] `circleci config validate .circleci/config.yml`
- [x] `ruby -c fastlane/Fastfile`
- [x] `bundle exec fastlane wait_for_pods_on_cdn` against the live CDN, resolving the plugin from git (not a local checkout): 18.33.1 resolved both pods and exited 0
- [x] Timeout path with an unpublished version: warns per pod, exits 0 so the bumps still run, and reports `180m remaining` confirming the cap reaches the action
- [ ] Exercised on a real release

## Follow-up, not blocking

The CDN request has no explicit HTTP timeout, so a hung connection stalls until `no_output_timeout` rather than failing fast. Harmless here given `when: always`, but worth hardening in the plugin.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches the stable-release path that fans out bumps to five hybrid repos; mis-tuned timeouts could delay releases or reopen red iOS jobs, though `soft_fail` and `when: always` limit worst-case impact.
> 
> **Overview**
> Adds a **single centralized wait** in the `trigger-dependent-updates` CircleCI job so hybrid bump PRs are not opened until `PurchasesHybridCommon` and `PurchasesHybridCommonUI` for the release version resolve on the CocoaPods CDN.
> 
> The new **`wait_for_pods_on_cdn`** Fastlane lane polls via `wait_for_pods_in_cocoapods_cdn` (updated **`fastlane-plugin-revenuecat_internal`** in `Gemfile.lock`) with a **3-hour cap** and **`soft_fail: true`**, so timeouts still allow downstream bumps. CircleCI runs that lane **before** the existing `bump_hybrid_dependencies` steps (which remain **`when: always`**) and sets **`no_output_timeout: 190m`** so a hung CDN request does not fail the step at the default 10 minutes. Fastlane README documents the new lane.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c94d4e9b0af27bd8eb5da3ee49379d83eae567a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -378,9 +378,14 @@ jobs:
       - revenuecat/install-gem-unix-dependencies:
           cache-version: v1
       # Blocks the bumps rather than each hybrid waiting on its own mac executors.
+      # The lane logs on every poll, so the default 10 minute no-output timeout
+      # wouldn't fire, but set it explicitly to sit just above the lane's own 45
+      # minute cap. If the step does fail, the bump steps below still run: they
+      # are all `when: always`.
       - run:
           name: Wait for the new pods to reach the CocoaPods CDN
           command: bundle exec fastlane wait_for_pods_on_cdn
+          no_output_timeout: 50m
       - run:
           name: Kick off Flutter automatic dependency update
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -379,13 +379,14 @@ jobs:
           cache-version: v1
       # Blocks the bumps rather than each hybrid waiting on its own mac executors.
       # The lane logs on every poll, so the default 10 minute no-output timeout
-      # wouldn't fire, but set it explicitly to sit just above the lane's own 45
-      # minute cap. If the step does fail, the bump steps below still run: they
-      # are all `when: always`.
+      # wouldn't fire, but set it explicitly to sit just above the lane's own 3
+      # hour cap, otherwise a stalled CDN request kills the step at 10 minutes.
+      # If the step does fail, the bump steps below still run: they are all
+      # `when: always`.
       - run:
           name: Wait for the new pods to reach the CocoaPods CDN
           command: bundle exec fastlane wait_for_pods_on_cdn
-          no_output_timeout: 50m
+          no_output_timeout: 190m
       - run:
           name: Kick off Flutter automatic dependency update
           command: |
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Looks good!

PR #1868, opened 2026-09-03
chore(ads): add ios availability check

peterporfy:
Ad rewards availability annotation will change to ios15+ - it prepares PHC so it won't break.

---

> [!NOTE]
> **Low Risk**
> Small defensive change for unsupported OS versions; callers of token generation must handle a nil return, which is a minor API surface change.
> 
> **Overview**
> Prepares **Purchases Hybrid Common** for RevenueCat ad-reward APIs moving to **iOS 15+** by adding runtime availability checks on the reward verification bridge, matching how other ad-related entry points are gated.
> 
> **`generateRewardVerificationToken`** now returns an optional dictionary: on older OS versions it logs and returns **`nil`** instead of calling into the SDK. **`pollRewardVerification`** (including the ObjC overload) short-circuits on unsupported platforms and invokes the completion with **`ErrorCode.unsupportedError`**, consistent with patterns like promotional offers and win-back offers elsewhere in `CommonFunctionality`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5317c73c301f520c42fdb340040166d6b39516f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -1266,11 +1266,11 @@ private extension CommonFunctionality {
 @objc public extension CommonFunctionality {
 
     @objc(generateRewardVerificationTokenWithImpressionId:)
-    static func generateRewardVerificationToken(impressionId: String) -> [String: Any] {
+    static func generateRewardVerificationToken(impressionId: String) -> [String: Any]? {
         guard #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) else {
             NSLog("[PurchasesHybridCommon] generateRewardVerificationToken: Reward verification requires " +
                   "iOS 15.0 or newer")
-            return [:]
+            return nil
         }
         return Purchases.shared.generateRewardVerificationToken(impressionId: impressionId).rc_dictionary
     }
--- a/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
+++ b/ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
@@ -1267,6 +1267,11 @@ private extension CommonFunctionality {
 
     @objc(generateRewardVerificationTokenWithImpressionId:)
     static func generateRewardVerificationToken(impressionId: String) -> [String: Any] {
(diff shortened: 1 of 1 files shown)


peterporfy:
* **#1868** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-hybrid-common/1868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-hybrid-common/1868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
```diff
@@ -1267,6 +1267,11 @@ private extension CommonFunctionality {
 
     @objc(generateRewardVerificationTokenWithImpressionId:)
     static func generateRewardVerificationToken(impressionId: String) -> [String: Any] {
+        guard #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) else {
+            NSLog("[PurchasesHybridCommon] generateRewardVerificationToken: Reward verification requires " +
+                  "iOS 15.0 or newer")
+            return [:]
```
ajpallares: I'd avoid returning `[:]` here. I _think_ the contract with the hybrids is `customData`/`clientTransactionId`/`appUserID` as required strings, so an empty map means Flutter throws `type 'Null' is not a subtype of type 'String'`, and RN/Unity just resolve it silently — the dev ends up passing `undefined` to AdMob's SSV options and rewards never verify.

Could we make it return `nil` instead? Still no hybrid changes needed, but at least it fails explicitly 🙂

ajpallares (commented):
Thanks for adding this! I have one concern though

ios/PurchasesHybridCommon/PurchasesHybridCommon/CommonFunctionality.swift
peterporfy: Sure, good call, thanks! Changed

ajpallares (approved):

PR #1895, opened 2026-09-17
Fix periodType type on customer info

okcoker:
Update type `PurchasesEntitlementInfo.periodType` to be `PeriodType` instead of `string` for better type hinting, and prevent case sensitive equality bugs. I'm assuming this was an oversight since `PurchasesSubscriptionInfo.periodType` within the same file uses `PeriodType` already.

I noticed iOS uses `UNKNOWN` but didn't want to yak shave on the TS type. Happy to add to the enum (or do a `PeriodType | 'UNKNOWN'`) if you think it makes sense

---

> [!NOTE]
> **Low Risk**
> Public TypeScript surface change only; may surface compile errors where callers compared `periodType` to arbitrary strings, which is the intended tightening.
> 
> **Overview**
> **Tightens typing** on `PurchasesEntitlementInfo.periodType` from `string` to the existing `PeriodType` union (`NORMAL` | `INTRO` | `TRIAL` | `PREPAID`), matching `PurchasesSubscriptionInfo.periodType` in the same module.
> 
> The public API report is regenerated to reflect the change. JSDoc on `expirationDate` / `expirationDateMillis` now references **`TRIAL`** instead of lowercase `trial`, consistent with the literal values in `PeriodType`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad9242a47f02d774897550008d2ae2f538c24716. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
Thanks for the contribution @okcoker! I think this makes sense! However, it could be considered a small breaking change... We're however planning to make a new major soon, so we will consider getting this in for that. We will hold this until then. Thanks again!

AlvaroBrey:
Hello @okcoker and thanks again! I'll be merging this into an intermediate branch to run CI in, while keeping your attribution. As said above we'll be merging into the `20.0-dev` branch for getting it out on the next major

AlvaroBrey (approved):

